PR #161, opened 2022-10-11
Public PrefixedDenom inner type and add as_str func for BaseDenom

### LINKED ISSUE #176: Provide public access to `PrefixedDenom`s trace path and denom

DaviRain-Su:
Closes: https://github.com/cosmos/ibc-rs/issues/176

Add some need help function.
______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/applications/transfer/amount.rs
+++ b/crates/ibc/src/applications/transfer/amount.rs
@@ -19,6 +19,10 @@ impl Amount {
     pub fn checked_sub(self, rhs: Self) -> Option<Self> {
         self.0.checked_sub(rhs.0).map(Self)
     }
+    
+    pub fn as_u256(&self) -> &U256 {
+        &self.0
+    }
 }
 
 impl FromStr for Amount {
--- a/crates/ibc/src/applications/transfer/amount.rs
+++ b/crates/ibc/src/applications/transfer/amount.rs
@@ -9,7 +9,7 @@ use crate::bigint::U256;
 #[derive(
     Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize, Display, From, Into,
 )]
-pub struct Amount(pub U256);
+pub struct Amount(U256);
 
 impl Amount {
     pub fn checked_add(self, rhs: Self) -> Option<Self> {
--- a/crates/ibc/src/applications/transfer/amount.rs
+++ b/crates/ibc/src/applications/transfer/amount.rs
@@ -9,7 +9,7 @@ use crate::bigint::U256;
 #[derive(
     Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize, Display, From, Into,
 )]
-pub struct Amount(U256);
+pub struct Amount(pub U256);
 
 impl Amount {
     pub fn checked_add(self, rhs: Self) -> Option<Self> {
@@ -18,10 +18,6 @@ impl Amount {
 
     pub fn checked_sub(self, rhs: Self) -> Option<Self> {
         self.0.checked_sub(rhs.0).map(Self)
-    }
-    
-    pub fn as_u256(&self) -> &U256 {
-        &self.0
     }
 }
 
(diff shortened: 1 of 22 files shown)


crates/ibc/src/applications/transfer/amount.rs
```diff
@@ -19,6 +19,10 @@ impl Amount {
     pub fn checked_sub(self, rhs: Self) -> Option<Self> {
         self.0.checked_sub(rhs.0).map(Self)
     }
+    
+    pub fn as_u256(&self) -> &U256 {
```
plafer: We should either make the internal type public, or expose just the methods we want from a `U256`. What methods do you need from the `U256` that are currently not exposed?

crates/ibc/src/core/ics04_channel/error.rs
```diff
@@ -26,6 +27,10 @@ define_error! {
             [ port_error::Error ]
             | _ | { "ics05 port error" },
 
+        Ics20Transfer
```
plafer: The [AppModule error](https://github.com/cosmos/ibc-rs/blob/da6653e0716350311d15e6acca08cad625f2d9ff/crates/ibc/src/core/ics04_channel/error.rs#L341-L347) should be used instead. By design, we don't want core IBC to know about specific applications

crates/ibc/src/applications/transfer/amount.rs
DaviRain-Su: For me, I want to get the internal type of u256.

crates/ibc/src/core/ics04_channel/error.rs
DaviRain-Su: removed

crates/ibc/src/applications/transfer/amount.rs
```diff
@@ -9,7 +9,7 @@ use crate::bigint::U256;
 #[derive(
     Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize, Display, From, Into,
 )]
-pub struct Amount(U256);
+pub struct Amount(pub U256);
```
hu55a1n1: `Amount` already implements `Into<U256>` (actually derives it using `derive_more::Into`), so you should be able to convert to `U256` directly using that. I don't think we should make the inner type public, because we want to restrict the arithmetic operations that we allow on `Amount` to `checking_{add,sub}`.

crates/ibc/src/applications/transfer/amount.rs
```diff
@@ -9,7 +9,7 @@ use crate::bigint::U256;
 #[derive(
```
hu55a1n1: Please feel free to add a changelog entry. See here for details on how to do so -> https://github.com/cosmos/ibc-rs/blob/main/CONTRIBUTING.md#changelog.

hu55a1n1 (changes requested):
Thanks for contributing @DaviRain-Su!

crates/ibc/src/applications/transfer/amount.rs
DaviRain-Su: Thx your relay, this is help me.

crates/ibc/src/applications/transfer/denom.rs
```diff
@@ -153,6 +159,14 @@ impl PrefixedDenom {
     pub fn add_trace_prefix(&mut self, prefix: TracePrefix) {
         self.trace_path.add_prefix(prefix)
     }
+
+    pub fn trace_path(&self) -> &TracePath {
```
plafer: It seems to me like we should simply make `trace_path` and `base_denom` public and remove all methods at this point. WDYT @hu55a1n1 ?

hu55a1n1: I agree. :+1: Both fields in `PrefixDenom` are correct-by-construction and `PrefixedDenom` performs no additional validations so we can make the fields public.

hu55a1n1: @DaviRain-Su would you like to make this change and update the changelog?

DaviRain-Su: ok nice

hu55a1n1 (approved):
:ok_hand:

PR #172, opened 2022-10-12
Add missing Tendermint client state checks

### LINKED ISSUE #22: Incomplete ICS 007 `ClientState` verification methods implementation
## Crate

`ibc`

## Summary

__Incomplete implementation of `check_header_and_update_state`__
- [ ] We should add a check that incoming header's `trusted_validators` matches the `next_validators_hash` that was persisted in the trusted consensus state
    - Equivalent check in ibc-go is here: https://github.com/cosmos/ibc-go/blob/147e0f1954ede691bb977ff185b9d0500ec4bd0a/modules/light-clients/07-tendermint/types/update.go#L152
    - The check is significant b/c the header's `trusted_validator_set` is used in the trusted block state in `verify`: https://github.com/informalsystems/ibc-rs/blob/b44ade9f88c7faf4d0b6f9e672dbef08d2ab3817/modules/src/clients/ics07_tendermint/client_def.rs#L98

We should fix (or investigate) the following remarks regarding verification of ICS007 `ClientState`.

- [ ] Unrelated to this issue, but I see that ibc-go restricts 07-tendermint `ChainId`s to `tendermint::MaxChainID` -> https://github.com/cosmos/ibc-go/issues/177

_Originally posted by @hu55a1n1 in https://github.com/informalsystems/ibc-rs/pull/1873#discussion_r809993503_

- [ ] We're also not checking if the `latest_height`'s `revision_number` matches the `ChainId`'s `version`.

_Originally posted by @hu55a1n1 in https://github.com/informalsystems/ibc-rs/pull/1873#discussion_r810002577_

- [ ] ibc-go doesn't allow a zero `max_clock_drift`. Not sure why this would be required. 🤔

_Originally posted by @hu55a1n1 in https://github.com/informalsystems/ibc-rs/pull/1873#discussion_r810004141_

- [ ] Apparently, each key in `upgrade_path` must be non-empty and this must also be checked.

_Originally posted by @hu55a1n1 in https://github.com/informalsystems/ibc-rs/pull/1873#discussion_r810018385_

## Acceptance Criteria
- [ ] Implement the verification checks mentioned above.
- [ ] Add tests similar these -> https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/client_state_test.go#L95
____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

hu55a1n1:
Closes: #22 

## Acceptance Criteria (copied from #22)

__Incomplete implementation of `check_header_and_update_state`__
- [ ] ~~We should add a check that incoming header's `trusted_validators` matches the `next_validators_hash` that was persisted in the trusted consensus state~~ (to be addressed with 02-client-refactor https://github.com/cosmos/ibc-rs/issues/173)
- [x] ibc-go restricts 07-tendermint `ChainId`s to `tendermint::MaxChainID` -> https://github.com/cosmos/ibc-go/issues/177
- [x] Check if the `latest_height`'s `revision_number` matches the `ChainId`'s `version`.
- [x] ibc-go doesn't allow a zero `max_clock_drift`. 
- [x] Apparently, each key in `upgrade_path` must be non-empty and this must also be checked.
- [x] Add tests similar these -> https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/client_state_test.go#L95
______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/clients/ics07_tendermint/client_state.rs
+++ b/crates/ibc/src/clients/ics07_tendermint/client_state.rs
@@ -787,10 +787,41 @@ impl TryFrom<RawTmClientState> for ClientState {
     type Error = Error;
 
     fn try_from(raw: RawTmClientState) -> Result<Self, Self::Error> {
-        let trust_level = raw
-            .trust_level
-            .clone()
-            .ok_or_else(Error::missing_trusting_period)?;
+        let chain_id = ChainId::from_string(raw.chain_id.as_str());
+
+        let trust_level = {
+            let trust_level = raw
+                .trust_level
+                .clone()
+                .ok_or_else(Error::missing_trusting_period)?;
+            trust_level
+                .try_into()
+                .map_err(|e| Error::invalid_trust_threshold(format!("{}", e)))?
+        };
+
+        let trusting_period = raw
+            .trusting_period
+            .ok_or_else(Error::missing_trusting_period)?
+            .try_into()
+            .map_err(|_| Error::negative_trusting_period())?;
+
+        let unbonding_period = raw
+            .unbonding_period
+            .ok_or_else(Error::missing_unbonding_period)?
+            .try_into()
+            .map_err(|_| Error::negative_unbonding_period())?;
+
+        let max_clock_drift = raw
+            .max_clock_drift
+            .ok_or_else(Error::missing_max_clock_drift)?
+            .try_into()
+            .map_err(|_| Error::negative_max_clock_drift())?;
+
+        let latest_height = raw
+            .latest_height
+            .ok_or_else(Error::missing_latest_height)?
+            .try_into()
+            .map_err(|_| Error::missing_latest_height())?;
 
         // In `RawClientState`, a `frozen_height` of `0` means "not frozen".
         // See:
@@ -799,41 +830,29 @@ impl TryFrom<RawTmClientState> for ClientState {
             .frozen_height
             .and_then(|raw_height| raw_height.try_into().ok());
 
-        #[allow(deprecated)]
-        Ok(Self {
-            chain_id: ChainId::from_string(raw.chain_id.as_str()),
-            trust_level: trust_level
-                .try_into()
-                .map_err(|e| Error::invalid_trust_threshold(format!("{}", e)))?,
-            trusting_period: raw
-                .trusting_period
-                .ok_or_else(Error::missing_trusting_period)?
-                .try_into()
-                .map_err(|_| Error::negative_trusting_period())?,
-            unbonding_period: raw
-                .unbonding_period
-                .ok_or_else(Error::missing_unbonding_period)?
-                .try_into()
-                .map_err(|_| Error::negative_unbonding_period())?,
-            max_clock_drift: raw
-                .max_clock_drift
-                .ok_or_else(Error::missing_max_clock_drift)?
-                .try_into()
-                .map_err(|_| Error::negative_max_clock_drift())?,
-            latest_height: raw
-                .latest_height
-                .ok_or_else(Error::missing_latest_height)?
-                .try_into()
-                .map_err(|_| Error::missing_latest_height())?,
-            frozen_height,
-            upgrade_path: raw.upgrade_path,
-            allow_update: AllowUpdate {
-                after_expiry: raw.allow_update_after_expiry,
-                after_misbehaviour: raw.allow_update_after_misbehaviour,
-            },
-            proof_specs: raw.proof_specs.into(),
-            verifier: ProdVerifier::default(),
-        })
+        let allow_update = AllowUpdate {
+            after_expiry: raw.allow_update_after_expiry,
+            after_misbehaviour: raw.allow_update_after_misbehaviour,
+        };
+
+        let client_state = {
+            let mut cs = ClientState::new(
+                chain_id,
+                trust_level,
+                trusting_period,
+                unbonding_period,
+                max_clock_drift,
+                latest_height,
+                raw.proof_specs.into(),
+                raw.upgrade_path,
+                allow_update,
+            )?;
+
+            cs.frozen_height = frozen_height;
+            cs
+        };
+
+        Ok(client_state)
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -47,18 +49,18 @@ pub const TENDERMINT_CLIENT_STATE_TYPE_URL: &str = "/ibc.lightclients.tendermint
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ClientState {
-    pub chain_id: ChainId,
```
plafer, line 50: Can we add this as a breaking change in the changelog?

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -80,6 +82,25 @@ impl ClientState {
...
+
+        // `TrustThreshold` is guaranteed to be in the range `[0, 1)`, but a `TrustThreshold::ZERO`
+        // value is invalid in this context
+        if trust_level == TrustThreshold::ZERO {
+            return Err(Error::invalid_trust_threshold(
+                "ClientState trust-level cannot be zero".to_string(),
+            ));
+        }
+
+        let _ = TendermintTrustThreshold::new(trust_level.numerator(), trust_level.denominator())
```
plafer, line 102: Note there is currently a [tiny bug](https://github.com/informalsystems/tendermint-rs/issues/1208) in tendermint-rs.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -757,41 +823,29 @@ impl TryFrom<RawTmClientState> for ClientState {
...
+                trusting_period,
+                unbonding_period,
+                max_clock_drift,
+                latest_height,
+                raw.proof_specs.into(),
+                raw.upgrade_path,
+                allow_update,
+            )?;
+
+            cs.frozen_height = frozen_height;
```
plafer: why don't we modify the constructor to take in a `frozen_height`?

plafer (changes requested):

crates/ibc/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: I thought of `ClientState::new()` as a function that creates a new client and it wouldn't make sense to create a frozen client. This is also in accordance with the spec -> https://github.com/cosmos/ibc/blob/main/spec/client/ics-007-tendermint-client/README.md#client-initialisation
But I don't think we're checking for this in the create client hander (neither does ibc-go) -> https://github.com/cosmos/ibc-rs/issues/178

crates/ibc/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: Interestingly, the spec also doesn't allow a `TrustThreshold == 1` -> https://github.com/cosmos/ibc/blob/main/spec/client/ics-007-tendermint-client/README.md#client-initialisation

crates/ibc/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: https://github.com/cosmos/ibc-rs/pull/172/commits/26f0c2a5bd7814367fc28d191aebd3de707df3e1

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: > I thought of `ClientState::new()` as a function that creates a new client and it wouldn't make sense to create a frozen client.

Makes sense. Therefore, I don't think we should use `ClientState::new()` here then, since we're not "creating a new client" here, we're only converting a client state from the wire (which might not be "new").

> But I don't think we're checking for this in the create client handler

In our current architecture, aren't we delegating that to `ctx.decode_client_state()` [here](https://github.com/cosmos/ibc-rs/blob/26f0c2a5bd7814367fc28d191aebd3de707df3e1/crates/ibc/src/core/ics02_client/handler/create_client.rs#L43)? Or are you suggesting we add an extra check for `client_state.is_frozen()` right after the context decodes the client_state (assuming the spec changes)?

hu55a1n1: > Therefore, I don't think we should use ClientState::new() here then

You're right :sweat_smile:. I used it here because we're doing all validation as part of `ClientState::new()` now. Would you prefer having a separate function `validate()` that is called by `ClientState::new()` and the `TryFrom` impl?

 > aren't we delegating that to `ctx.decode_client_state()` here?
 
 Yes, but I think this should be checked by the handler and we shouldn't rely on the host to give us an active client.

plafer: > Would you prefer having a separate function validate() that is called by ClientState::new() and the TryFrom impl?

Well, do we even need `new()`? It is never used in the modules. If for some reason we prefer to keep it though, then sure a `validate()` would make sense.

> Yes, but I think this should be checked by the handler and we shouldn't rely on the host to give us an active client.

Agreed.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: Linking the spec issue -> https://github.com/cosmos/ibc/issues/857

crates/ibc/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: I decided to keep `new()` and add a `frozen_height` parameter to it as initially suggested. I like to keep the `new()` method as it provides a way to create a correct `ClientState` (without requiring a raw struct), and it's also being used in tests (otherwise we'd need to make the fields `pub(crate)`).

hu55a1n1: https://github.com/cosmos/ibc-rs/pull/172/commits/27f3cbf5da92920c3e3ff25431de765219a05288 and https://github.com/cosmos/ibc-rs/pull/172/commits/355a4a152189ba5dd304dad714009abec155ff89

plafer (approved):
Great stuff 💯

PR #207, opened 2022-10-31
Add `validate_self_client()`

### LINKED ISSUE #162: `ValidateSelfClient` missing in in `ConnOpen{Try,Ack}`
We're missing the validation that the `ClientState` included in the msg is as we expect it to be.

See [ibc-go](https://github.com/cosmos/ibc-go/blob/567178ce833c8f917645ca4f8b494bb9c177317f/modules/core/03-connection/keeper/handshake.go#L86)
____

## For Admin Use

- [ ] Not duplicate issue

plafer:
Closes: #162
 
______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/host_helpers.rs
+++ b/crates/ibc/src/host_helpers.rs
@@ -7,6 +7,8 @@ use crate::core::ics03_connection::error::Error;
 use crate::core::ics23_commitment::specs::ProofSpecs;
 use crate::core::ics24_host::identifier::ChainId;
 
+use tendermint::trust_threshold::TrustThresholdFraction as TendermintTrustThresholdFraction;
+
 pub trait TmValidateSelfClientContext {
     fn validate_self_client(
         &self,
@@ -38,6 +40,16 @@ pub trait TmValidateSelfClientContext {
         if self.proof_specs() != counterparty_client_state.proof_specs() {
             return Err(Error::invalid_client_state());
         }
+
+        let _ = {
+            let trust_level = counterparty_client_state.trust_level();
+
+            TendermintTrustThresholdFraction::new(
+                trust_level.numerator(),
+                trust_level.denominator(),
+            )
+            .map_err(|_| Error::invalid_client_state())?
+        };
 
         Ok(())
     }
--- a/crates/ibc/src/host_helpers.rs
+++ b/crates/ibc/src/host_helpers.rs
@@ -13,7 +13,9 @@ use crate::core::ics24_host::identifier::ChainId;
 
 use tendermint::trust_threshold::TrustThresholdFraction as TendermintTrustThresholdFraction;
 
-pub trait TmValidateSelfClientContext {
+/// Provides an implementation of `ConnectionReader::validate_self_client` for
+/// Tendermint-based hosts. 
+pub trait TendermintValidateSelfClientContext {
     fn validate_self_client(
         &self,
         ctx: &dyn ConnectionReader,
--- a/crates/ibc/src/host_helpers.rs
+++ b/crates/ibc/src/host_helpers.rs
@@ -1,6 +1,7 @@
 use core::time::Duration;
 
-use alloc::string::String;
+use alloc::format;
+use alloc::string::{String, ToString};
 use ibc_proto::google::protobuf::Any;
 
 use crate::clients::ics07_tendermint::client_state::ClientState as TmClientState;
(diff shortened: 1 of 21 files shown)


crates/ibc/src/host_helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::clients::ics07_tendermint::client_state::ClientState as TmClientState;
+use crate::core::ics02_client::client_state::ClientState;
+use crate::core::ics03_connection::context::ConnectionReader;
+use crate::core::ics03_connection::error::Error;
+use crate::core::ics23_commitment::specs::ProofSpecs;
+use crate::core::ics24_host::identifier::ChainId;
+
+use tendermint::trust_threshold::TrustThresholdFraction as TendermintTrustThresholdFraction;
+
+pub trait TmValidateSelfClientContext {
```
hu55a1n1: IIUC, this trait is intended to be implemented by the host, right? If so, we can remove the `ctx` parameter as `self` would be the host itself.

crates/ibc/src/host_helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+        let self_revision_number = self_chain_id.version();
+        if self_revision_number != counterparty_client_state.latest_height().revision_number() {
+            return Err(Error::invalid_client_state(format!(
+                "client is not in the same revision as the chain. expected: {}, got: {}",
+                self_revision_number,
+                counterparty_client_state.latest_height().revision_number()
+            )));
+        }
+
+        if counterparty_client_state.latest_height() >= ctx.host_current_height() {
```
hu55a1n1: I would recommend adding a `host_current_height()` trait method to facilitate the removal of the `ConnectionReader` arg. Also, would suggest moving this into the Tendermint light client dir (to avoid circular dependencies in the future if and when we decide to extract the light client into it's own crate).

crates/ibc/src/host_helpers.rs
hu55a1n1: Since this is public API, it would be nice to add some doc comments here.

crates/ibc/src/mock/context.rs
```diff
@@ -1133,6 +1133,10 @@ impl ConnectionReader for MockContext {
     fn connection_counter(&self) -> Result<u64, Ics03Error> {
         Ok(self.ibc_store.lock().unwrap().connection_ids_counter)
     }
+
+    fn validate_self_client(&self, _counterparty_client_state: Any) -> Result<(), Ics03Error> {
+        Ok(())
```
hu55a1n1, line 1138: Would be helpful to implement this treating the Mock chain as a Tendermint chain (maybe in a separate PR) to ensure the trait can be used as intended. The impl will also serve as a reference. WDYT?

hu55a1n1 (commented):
Nicely done @plafer! I added some suggestions.

crates/ibc/src/host_helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        if counterparty_client_state.unbonding_period < counterparty_client_state.trusting_period {
+            return Err(Error::invalid_client_state(format!(
+                "unbonding period must be greater than trusting period. unbonding period ({:?}) < trusting period ({:?})",
+                counterparty_client_state.unbonding_period,
+                counterparty_client_state.trusting_period
+            )));
+        }
+
+        if self.upgrade_path() != counterparty_client_state.upgrade_path {
```
hu55a1n1: Just noticed that ibc-go allows an empty `counterparty_client_state.upgrade_path` regardless of whether or not the host upgrade path is empty -> https://github.com/cosmos/ibc-go/blob/v5.0.1/modules/core/02-client/keeper/keeper.go#L322-L329

crates/ibc/src/host_helpers.rs
plafer: > If so, we can remove the ctx parameter as self would be the host itself.

That could indeed, if we added `ConnectionReader` as a super trait. The current approach gives the flexibility for hosts to implement `TmValidateSelfClientContext` on a different struct; not sure if this would be needed though?

plafer: > Since this is public API, it would be nice to add some doc comments here.

[354f9fc](https://github.com/cosmos/ibc-rs/pull/207/commits/354f9fc855abdd1d6be6aedc229bd0c244b98a17)

plafer: > That could indeed, if we added ConnectionReader as a super trait. The current approach gives the flexibility for hosts to implement TmValidateSelfClientContext on a different struct; not sure if this would be needed though?

Ah, will [apply this suggestion](https://github.com/cosmos/ibc-rs/pull/207#discussion_r1012623420) instead.

crates/ibc/src/host_helpers.rs
plafer: Good idea:
[70c76dc](https://github.com/cosmos/ibc-rs/pull/207/commits/70c76dc015f5af365a692372285f55cb59720445), [1dd6856](https://github.com/cosmos/ibc-rs/pull/207/commits/1dd6856418ac164192497f43e309f1a46270b933) and [2799101](https://github.com/cosmos/ibc-rs/pull/207/commits/2799101fa7e9d17b21aa1eee263d9f195f093310)

crates/ibc/src/mock/context.rs
plafer: Good idea; except won't it cause trouble when we move the tendermint client to another crate?

crates/ibc/src/host_helpers.rs
plafer: Great catch! [af79e61](https://github.com/cosmos/ibc-rs/pull/207/commits/af79e611b108ee4644b184a82007dc5dd24e0e34)

crates/ibc/src/mock/context.rs
hu55a1n1: Good point :sweat_smile:, it would be problematic unless we extract the Mock impl to another crate as well! Let's leave it for now.

hu55a1n1 (approved):

PR #208, opened 2022-11-01
Update to tendermint 0.26

mzabaluev:
## Description

Updates for tendermint-rs 0.26

Requres changes from https://github.com/cosmos/ibc-proto-rs/pull/31

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/ics02_client/events.rs
```diff
@@ -45,12 +45,9 @@ struct ClientTypeAttribute {
...
-impl From<ClientTypeAttribute> for Tag {
+impl From<ClientTypeAttribute> for abci::EventAttribute {
     fn from(attr: ClientTypeAttribute) -> Self {
-        Tag {
-            key: CLIENT_TYPE_ATTRIBUTE_KEY.parse().unwrap(),
-            value: attr.client_type.to_string().parse().unwrap(),
-        }
+        (CLIENT_TYPE_ATTRIBUTE_KEY, attr.client_type.as_str()).into()
     }
 }
```
mzabaluev, line 48: Note that the `EventAttribute` domain type also has an `index` field, which is set to false by the `From` conversion used here.

Cargo.toml
```diff
@@ -9,3 +9,6 @@ members = [
 exclude = [
     "ci/no-std-check",
 ]
+
+[patch.crates-io]
```
hu55a1n1: Just FYI, ibc-proto-rs v0.22.0 has now been released (with these changes) so we can now depend on it directly. https://github.com/cosmos/ibc-proto-rs/releases/tag/v0.22.0

crates/ibc/src/core/ics02_client/events.rs
mzabaluev: Hmm, it may have been too early to switch to `tendermint::abci::Event` before https://github.com/informalsystems/tendermint-rs/pull/1204 is released. There is still `tendermint_rpc::abci::Event` confusingly, but it's what hermes is using.

mzabaluev: Upon further examination, the ABCI event infrastructure in Hermes now duplicates this code and is not dependent on `ibc`. So we can go on with this change.

crates/ibc/src/core/ics03_connection/events.rs
```diff
@@ -30,30 +29,25 @@ struct Attributes {
 /// is infallible, even if it is not represented in the error type.
```
hu55a1n1: We can finally remove this doc comment now.

crates/ibc/src/core/ics04_channel/msgs/acknowledgement.rs
```diff
@@ -20,6 +20,11 @@ impl Acknowledgement {
     pub fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
+
+    // Returns the data as a slice of bytes.
+    pub fn as_bytes(&self) -> &[u8] {
```
hu55a1n1, line 25: Nit: we already have a `AsRef<[u8]>` impl below so can we use that instead of defining this new public method?

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -194,6 +194,12 @@ impl Default for ClientId {
     }
 }
 
+impl From<ClientId> for String {
```
hu55a1n1: Nit: Can we use `derive_more::Into` instead?

crates/ibc/Cargo.toml
```diff
@@ -48,19 +48,19 @@ primitive-types = { version = "0.12.0", default-features = false, features = ["s
 dyn-clone = "1.0.8"
 
 [dependencies.tendermint]
-version = "=0.25.0"
+version = "0.26.0"
```
hu55a1n1, line 51: Does tendermint-rs always bump the versions of all these component crates whenever there is a breaking change in any one of them? Just thinking if not pinning these dependencies could break builds for our customers in the future.

hu55a1n1 (commented):
Love the changes in this PR! :ok_hand: Added some minor suggestions and a question about tendermint-rs dep versions.

crates/ibc/Cargo.toml
hdevalence: Bystander, but: the key change in tendermint-rs `0.26` (and the reason we're interested in it) was to go back to "normal" use of semver, where the patch versions don't have breaking changes. So this should be fine.

hu55a1n1: Thank you for the clarification @hdevalence. :pray:

crates/ibc/src/core/ics04_channel/msgs/acknowledgement.rs
mzabaluev: I find `.as_ref()` not descriptive enough, and following the std practice also provided an intrinsic method with a clearer name.
In general, `AsRef` makes the type more usable as a parameter in generic functions that use that bound, while the intrinsic `.as_slice()`/`.as_bytes()` method is more readable in code using the concrete type.

crates/ibc/src/core/ics03_connection/events.rs
mzabaluev: Ah, missed this one.

crates/ibc/src/core/ics24_host/identifier.rs
mzabaluev: https://github.com/cosmos/ibc-rs/pull/208/commits/43b8589fa6211ed50f4025579bcd5912f2becab1

crates/ibc/src/core/ics03_connection/events.rs
mzabaluev: https://github.com/cosmos/ibc-rs/pull/208/commits/5eaa6b337e5bd47c83974268d94391ddffb66680

hu55a1n1 (approved):

Issue #173, opened 2022-10-12
Implement (parts of) ibc-go/ADR006


hu55a1n1:
## Summary

This is a tracking issue for the Q1 objective of _"Implementing of relevant parts of [ADR006 (AKA 02-client-refactor)](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-006-02-client-refactor.md)"_ 
(_See https://github.com/cosmos/ibc-go/pull/1871 for more context_)

### Acceptance Criteria
- [x] ~~merging header/misbehaviour interface and renaming to ClientMessage~~
- [ ] https://github.com/cosmos/ibc-rs/issues/404
- [ ] https://github.com/cosmos/ibc-rs/issues/530
- [ ] https://github.com/cosmos/ibc-rs/issues/535
- [ ] setting client/consensus states within the light client instead of at the 02-client level
- [ ] emission of upgraded consensus state
- [ ] We should add a check that incoming header's `trusted_validators` matches the `next_validators_hash` that was persisted in the trusted consensus state (copied from https://github.com/cosmos/ibc-rs/issues/22 as this wasn't addressed in #22 and depends on 02-client-refactor changes)
    - Equivalent check in ibc-go is here: https://github.com/cosmos/ibc-go/blob/147e0f1954ede691bb977ff185b9d0500ec4bd0a/modules/light-clients/07-tendermint/types/update.go#L152
    - The check is significant b/c the header's `trusted_validator_set` is used in the trusted block state in `verify`: https://github.com/informalsystems/ibc-rs/blob/b44ade9f88c7faf4d0b6f9e672dbef08d2ab3817/modules/src/clients/ics07_tendermint/client_def.rs#L98
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

hu55a1n1:
We must add fix the misbehaviour execute handler to not perform full verification. (see https://github.com/cosmos/ibc-rs/pull/290#pullrequestreview-1214171067)

Farhad-Shabani:
Looks like we've resolved items of this issue. @plafer, could you also check?

plafer:
Yup!

plafer:
Yup!

PR #241, opened 2022-11-14
Implement `ValidationContext` and `ExecutionContext` for clients (ICS-2)

### LINKED ISSUE #240: ADR 5: Implement `ValidationContext` and `ExecutionContext` for clients (ICS-2)
Define both traits and implement `ValidationContext::validate()` & `ExecutionContext::execute()`.

plafer:
Closes: #240

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


.changelog/unreleased/breaking-changes/240-validation-execution-contexts.md
```diff
@@ -0,0 +1,2 @@
+- Add  and , and implement for clients (ICS-2)
```
hu55a1n1: Looks like the terminal treated the validation traits as commands. :sweat_smile:

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -19,14 +21,117 @@ use crate::timestamp::Timestamp;
...
+
+    // Read client type from the host chain store. The client should already exist.
+    // Read client state from the host chain store.
+    let client_state = ctx.client_state(&client_id)?;
+
+    let UpdatedState {
+        client_state,
+        consensus_state,
+    } = client_state
+        .check_header_and_update_state(ctx, client_id.clone(), header.clone())
```
hu55a1n1, line 101: I think we shouldn't be running the complete header verification here and should only decode the header to obtain new client and consensus states. https://github.com/anoma/namada/blob/main/shared/src/ledger/ibc/handler.rs#L1187

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -19,14 +21,117 @@ use crate::timestamp::Timestamp;
...
+        ));
+    }
+
+    // Use client_state to validate the new header against the latest consensus_state.
+    // This function will return the new client_state (its latest_height changed) and a
+    // consensus_state obtained from header. These will be later persisted by the keeper.
+    let _ = client_state
+        .check_header_and_update_state(ctx, client_id.clone(), header)
+        .map_err(|e| Error::header_verification_failure(e.to_string()))?;
+
```
hu55a1n1, line 79: IIUC, the `validate()` function is Anoma's VP equivalent right? If so, the logic is incomplete, for e.g. w're missing verifying the previously emitted event (emitted during execution). https://github.com/anoma/namada/blob/main/shared/src/ledger/ibc/vp/client.rs#L311

crates/ibc/src/core/context.rs
```diff
@@ -0,0 +1,396 @@
...
+    fn max_expected_time_per_block(&self) -> Duration;
+
+    /// Calculates the block delay period using the connection's delay period and the maximum
+    /// expected time per block.
+    fn block_delay(&self, delay_period_time: Duration) -> u64 {
+        calculate_block_delay(delay_period_time, self.max_expected_time_per_block())
+    }
+}
+
+pub trait ExecutionContext: ValidationContext {
```
hu55a1n1, line 243: I think using `ValidationContext` as a supertrait here is convenient but might be confusing to implementors if it has methods such as `check_emitted_event()`, etc. which only make sense for VPs.

crates/ibc/src/core/ics02_client/handler/create_client.rs
```diff
@@ -28,6 +34,84 @@ pub struct Result {
...
+    let client_state = ctx.decode_client_state(client_state)?;
+
+    let client_type = client_state.client_type();
+
+    let client_id = ClientId::new(client_type.clone(), id_counter).map_err(|e| {
+        Error::client_identifier_constructor(client_state.client_type(), id_counter, e)
+    })?;
+    let consensus_state = client_state.initialise(consensus_state)?;
+
+    ctx.store_client_type(ClientTypePath(client_id.clone()), client_type.clone())?;
```
hu55a1n1, line 83: Would be nice if we could extract blocks like these (e.g. `write_create_client()`, `emit_create_client_event()`, etc.) and reused them in the future to avoid code duplication.

hu55a1n1 (commented):
I like the general structure and approach of the changes in this PR. Will approve once the logic mismatch in the VPs and `validate()` is addressed/clarified. We can improve on the other suggestions iteratively in future PRs.

.changelog/unreleased/breaking-changes/240-validation-execution-contexts.md
plafer: [977a870](https://github.com/cosmos/ibc-rs/pull/241/commits/977a8704aaee9501211ff40c02e9b967e4432740)

crates/ibc/src/core/ics02_client/handler/update_client.rs
plafer: I see Anoma's VP as a superset of `validate()`. For example, "verifying that the event was previously emitted" is specific to Anoma's architecture, and doesn't belong in ibc-rs. So Anoma's VP would first call `validate()` (IBC-specific checks), and then do their own custom checks (suck as making sure the event was properly emitted).

crates/ibc/src/core/context.rs
plafer: As discussed [here](https://github.com/cosmos/ibc-rs/pull/241#discussion_r1026565196), I don't think we'll have Anoma VP-specific methods in `ValidationContext`. Still worth making sure that all methods in `ValidationContext` make sense in `ExecutionContext` though.

crates/ibc/src/core/ics02_client/handler/update_client.rs
hu55a1n1: Thanks for clarifying this. :+1:

hu55a1n1 (approved):

crates/ibc/src/core/ics02_client/handler/update_client.rs
plafer: Right. Should we do the `check_header_and_update_state()` split from #173 first then? Or perhaps merge this but open an issue, which we address before the next release?

hu55a1n1: I am okay with both, but I would prefer the second option (i.e. merge this and open an issue).

PR #314, opened 2022-12-22
Trim redundant code

### LINKED ISSUE #154: Remove relayer specific code and dependencies from the modules crate

Farhad-Shabani:
Closes: #154

## Purpose

Clean up unnecessary/redundant elements

## Description
In reviewing the codebase to address #154, I also found some other redundant code that has been fixed in this PR as below:

1. Streamlined  `lookup_module()` of `ChannelMsg` and `PacketMsg`
2. Trimmed the `route()` method and exposed it as the default implementation of the `Msg` trait.
(More viable pruning around `Msg` covered in #315)
3. Modified the output of `MsgCreateClient::new()` to return `Self` instead of `Result<Self, ClientError>`, as the error case is unreachable.
4. Removed redundant `client_type()` method of `ConsensusState` and `Header` traits since they have no clear usage (This was confirmed through a search for use cases in dependent projects such as Nomic, ComposableFi, and Penumbra)
5. Removed unnecessary `PortReader` implementation on `DummyTransferModule` struct
6. Cleared unused `moderator` dev dependency
7. Cleared redundant precludes

Additional to consider:
1. Clear `PrettyDuration` and `PrettyOption` structs [here](https://github.com/cosmos/ibc-rs/blob/6b6548315fc7bee0b39f4aefa558db21a60a0d8f/crates/ibc/src/utils/pretty.rs#L8-L25)

______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:
- [ ] Look for any out-of-scope side effects of clearing out these parts


crates/ibc/Cargo.toml
```diff
@@ -95,7 +95,6 @@ default-features = false
 env_logger = "0.10.0"
 tracing-subscriber = { version = "0.3.14", features = ["fmt", "env-filter", "json"]}
 test-log = { version = "0.2.10", features = ["trace"] }
-modelator = "0.4.2"
 sha2 = { version = "0.10.6" }
```
plafer: I don't think we need this one either

crates/ibc/src/relayer/ics18_relayer/utils.rs
```diff
@@ -111,14 +111,6 @@ mod tests {
             // Update client on chain B to latest height of A.
             // - create the client update message with the latest header from A
             let a_latest_header = ctx_a.query_latest_header().unwrap();
-            assert_eq!(
```
plafer: I think we should get rid of the entire `crates/ibc/relayer/` directory. `RelayerContext` is only implemented on `MockContext`; we should move the used methods there directly. Otherwise the test in `utils.rs` should be moved somewhere in the `ics26_routing` directory.

plafer (changes requested):
Great work, looks good! I identified a few more things to change:

- Can we remove `bigint.rs`? The re-exported `primitive_types::U256` should be simply used directly in other modules.
- `crate::utils::PrettyDuration` is never used and should be deleted (as you identified)
- Can we move the `crate::macros` module under `crate::utils`?

Farhad-Shabani:
> * `crate::utils::PrettyDuration` is never used and should be deleted (as you identified)

Can we also remove `crate::utils::PrettyOption`? This also isn't used anywhere

crates/ibc/src/relayer/ics18_relayer/utils.rs
Farhad-Shabani: I removed the `relayer` and moved `ics18_relayer` under the mock module. https://github.com/cosmos/ibc-rs/pull/314/commits/65af6b50807ee701344777210816ffb1fe4432e3
Do you think it makes sense?

plafer:
> Can we also remove crate::utils::PrettyOption? This also isn't used anywhere

Yes!

crates/ibc/src/relayer/ics18_relayer/utils.rs
plafer: Good idea yes. While we're at it, I think we should:

- Remove `utils.rs` and move the contents in `context.rs` (doesn't make sense to me why this big test is a `utils` module
- move `build_client_update_datagram()` in the `tests` submodule, since it's really a helper function for the test

WDYT?

Farhad-Shabani: Agree, Will do so

plafer:
Just realized we should also update the module docstring in `lib.rs` (remove the references to the relayer module

plafer (approved):

PR #318, opened 2022-12-28
Make the `mocks` feature work with `no_std`

### LINKED ISSUE #311: Make the code under mocks features work in no-std
## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue

DaviRain-Su:
Closes: #311 

## Description

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


plafer:
Thank you for identifying this issue! However I don't think this is the right approach; it seems like we can make the entire `mocks` feature `no_std`. Namely, we are [using a `Mutex`](https://github.com/cosmos/ibc-rs/blob/060d80fc66c1de1a82c64ac193f7329ed3f2d326/crates/ibc/src/mock/context.rs#L13), which we really don't need. I'll look into it this week, and if this is indeed the case, we'll close this PR in favor of the other one I'll open.

DaviRain-Su:
> Thank you for identifying this issue! However I don't think this is the right approach; it seems like we can make the entire `mocks` feature `no_std`. Namely, we are [using a `Mutex`](https://github.com/cosmos/ibc-rs/blob/060d80fc66c1de1a82c64ac193f7329ed3f2d326/crates/ibc/src/mock/context.rs#L13), which we really don't need. I'll look into it this week, and if this is indeed the case, we'll close this PR in favor of the other one I'll open.

`Mutex` does not fit in `no_std`, I found a crate `parking_lot` that provides `Mutex` in `no_std`, but the most important issue is `Timestamp`, which cannot be used in `no_std`, because the `now` method of `Timestamp` can only be used in `std`.

plafer:
> but the most important issue is Timestamp, which cannot be used in no_std, because the now method of Timestamp can only be used in std.

For this, we can do solve the problem in a similar way that you did, except conditioning on the `std` feature instead of `mock-primitives`.

So I think the end solution should look like:
1. Remove the unneeded use of `Mutex`
2. Use `Timestamp::none` if the `std` feature is disabled

crates/ibc/Cargo.toml
```diff
@@ -73,6 +73,8 @@ parity-scale-codec = { version = "3.0.0", default-features = false, features = [
 scale-info = { version = "2.1.2", default-features = false, features = ["derive"], optional = true }
 ## for borsh encode or decode
 borsh = {version = "0.9.3", default-features = false, optional = true }
+parking_lot = { version = "0.12.1", default-features = false }
+cfg-if = "1.0.0"
```
plafer: Could you move those under `[dev-dependencies]`?

plafer (changes requested):
Thank you for this!

Could you also add a changelog entry under "improvements"?

plafer:
While we're at it, we should also modify the CI. In `ci/no-std-check/Cargo.toml`, could you please change line 8 from

```toml
ibc = { path = "../../crates/ibc", default-features = false }
```
to
```toml
ibc = { path = "../../crates/ibc", default-features = false, features = ["mocks"] }
```

DaviRain-Su:
The `mock` feature also relies on `mocks = ["tendermint-testgen", "clock"]`, since `tendermint-testgen` and `clock` seem to be available only in `std`. I think it is difficult to modify it to support `no_std`, here I think it is better to expose only `Clientstate, ConsensusState, Header` in `no_std` through the feature, the whole mocks support `no_std` in the modification is still found to be very problematic.

crates/ibc/Cargo.toml
DaviRain-Su: I don't think it should be transferred over, because opening the mocks feature brings in some needed code

DaviRain-Su:
I configured `tendermint-testgen` to `no-std-check` and found that `tendemrint-testgen` does not pass the `check-substrate` test.

crates/ibc/Cargo.toml
plafer: Ah yes you're right

plafer:
Agreed. Let's reintroduce the feature; can we call it `mocks-no-std` instead?

I opened #329 too, which we can take care of later.

crates/ibc/Cargo.toml
plafer: Then can you make them optional, and a dependency of `mocks`?

plafer:
~~`cargo test --no-default-features --features mocks-no-std` currently fails. Could you add this to the CI (in `.github/workflows/rust.yml`) as well?~~

This doesn't work on `main` either, let's not fix this here

crates/ibc/src/mock/header.rs
```diff
@@ -71,9 +71,18 @@ impl MockHeader {
...
-        Self {
-            height,
-            timestamp: Timestamp::now(),
+        cfg_if::cfg_if! {
+            if #[cfg(any(test, feature = "mocks"))]  {
+                Self {
+                    height,
+                    timestamp: Timestamp::now(),
+                }
+            } else if #[cfg(any(test,feature = "mocks-no-std"))] {
```
plafer:
```suggestion
            } else {
```
Improves readability, and ensures that we always compile something in.

crates/ibc/src/mock/header.rs
```diff
@@ -71,9 +71,18 @@ impl MockHeader {
     }
 
     pub fn new(height: Height) -> Self {
-        Self {
-            height,
-            timestamp: Timestamp::now(),
+        cfg_if::cfg_if! {
+            if #[cfg(any(test, feature = "mocks"))]  {
```
plafer:
```suggestion
            if #[cfg(any(test, feature = "std"))]  {
```
This file already only gets compiled only if `any(test, feature = "mocks", feature = "mocks-no-std")`. This better represents when we want to use `Timestamp::now()` vs `Timestamp::none()`

plafer (changes requested):
nits

DaviRain-Su:
> ~`cargo test --no-default-features --features mocks-no-std` currently fails. Could you add this to the CI (in `.github/workflows/rust.yml`) as well?~
> 
> This doesn't work on `main` either, let's not fix this here

```rust
pub fn now() -> Timestamp {
        cfg_if::cfg_if! {
            if #[cfg(feature = "clock")]  {
                Time::now().into()
            } else {
                Self::none()
            }
        }
    }
```
Try to make `Timestamp's now` method like this, but some tests will not run through.

plafer (approved):
Thank you!

I'll remove the `clock` feature in another PR

PR #327, opened 2023-01-06
Exclude `ChannelEnd` from `MsgChannelOpenInit` and `MsgChannelOpenTry`

### LINKED ISSUE #20: Refactor: `MsgChannelOpenInit` and `MsgChannelOpenTry` should not contain a `ChannelEnd`
## Summary of bug
Both `MsgChannelOpenInit` and `MsgChannelOpenTry` contain a `ChannelEnd` field. Although convenient because `ChannelEnd` contains most of the fields we need, it's semantically incorrect and confusing to developers not used to the codebase. For example, in the case of `MsgChannelOpenInit`, how am I supposed to provide one end of the channel I am trying to create?

Note that `MsgChannelOpenAck` and `MsgChannelOpenConfirm` don't contain a `ChannelEnd` field.

## Version

v0.13.0

## Acceptance Criteria
`MsgChannelOpenInit` and `MsgChannelOpenTry` don't contain a `ChannelEnd` field.

____

## For Admin Use

- [ ] Not duplicate issue

Farhad-Shabani:
Closes: #20

### Side effect
Bringing `ChannelEnd` fields along with including proofs inside the struct of messages, adds `#allow(clippy::too_many_arguments)]` for `MsgChannelOpenTry`

______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/ics04_channel/msgs/chan_open_init.rs
```diff
@@ -17,15 +20,31 @@ pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgChannelOpenInit";
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MsgChannelOpenInit {
     pub port_id_on_a: PortId,
-    pub chan_end_on_a: ChannelEnd,
+    pub ordering_on_a: Order,
```
plafer: Since the ordering is forced to be the same on A and B, I think we should rename this to `ordering`

crates/ibc/src/core/ics04_channel/msgs/chan_open_init.rs
```diff
@@ -17,15 +20,31 @@ pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgChannelOpenInit";
...
+    pub version_on_a: Version,
```
plafer: Since the [removal of crossing hellos](https://github.com/cosmos/ibc/issues/628), this field is an "optional version proposition by the relayer". So I think a better name would be `version_proposal`. I would also add a docstring for that field explaining just that.

crates/ibc/src/core/ics04_channel/msgs/chan_open_init.rs
```diff
@@ -17,15 +20,31 @@ pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgChannelOpenInit";
...
+    pub connection_hops_on_a: Vec<ConnectionId>,
+    pub port_id_on_b: PortId,
+    pub chan_id_on_b: Option<ChannelId>,
```
plafer: This field is guaranteed to be `None` in the `OpenInit` message (the counterparty channel id has not been created yet!), so we should remove it.

This is also part of the reason why using a `ChannelEnd` wasn't elegant :)

crates/ibc/src/core/ics04_channel/msgs/chan_open_try.rs
```diff
@@ -20,7 +22,11 @@ pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgChannelOpenTry";
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MsgChannelOpenTry {
     pub port_id_on_b: PortId,
-    pub chan_end_on_b: ChannelEnd,
+    pub ordering_on_b: Order,
+    pub version_on_b: Version,
+    pub connection_hops_on_b: Vec<ConnectionId>,
+    pub port_id_on_a: PortId,
+    pub chan_id_on_a: Option<ChannelId>,
```
plafer: The type should be `ChannelId` here, since it is a requirement to provide the counterparty channel id in `OpenTry`

crates/ibc/src/core/ics04_channel/msgs/chan_open_try.rs
```diff
@@ -20,7 +22,11 @@ pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgChannelOpenTry";
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MsgChannelOpenTry {
     pub port_id_on_b: PortId,
-    pub chan_end_on_b: ChannelEnd,
+    pub ordering_on_b: Order,
```
plafer: Similarly, this should be just `ordering`

crates/ibc/src/core/ics04_channel/msgs/chan_open_try.rs
```diff
@@ -20,7 +22,11 @@ pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgChannelOpenTry";
...
+    pub version_on_b: Version,
```
plafer: This field is unused, and should theoretically be removed. However, we should keep it around like `previous_channel_id` just for proper conversion purposes. Could you move it under `previous_channel_id`, add the same comment + deprecation, and rename to `version_proposal`?

I also noticed that a few tests incorrectly use it; these should be changed to use `version_on_a` instead.

crates/ibc/src/core/ics04_channel/msgs/chan_open_try.rs
```diff
@@ -20,7 +22,11 @@ pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgChannelOpenTry";
...
+    pub connection_hops_on_b: Vec<ConnectionId>,
+    pub port_id_on_a: PortId,
+    pub chan_id_on_a: Option<ChannelId>,
     pub version_on_a: Version,
```
plafer: I now think a better name for this would be `version_supported_on_a`, as it better represents what the version field means at this point in the handshake.

plafer (changes requested):
🙌

crates/ibc/src/core/ics04_channel/msgs/chan_open_init.rs
Farhad-Shabani: I doubted that it would be so. Alright

crates/ibc/src/core/ics04_channel/msgs/chan_open_init.rs
Farhad-Shabani: Makes sense 👍

crates/ibc/src/core/ics04_channel/msgs/chan_open_try.rs
Farhad-Shabani: IIUC, for the test you mentioned, I think the correct one should be `Version::empty()`

plafer (commented):
Approved, but let's wait for #328 to be merged first as there'll probably merge conflicts/tiny things to fix

plafer (approved):
🙌

PR #341, opened 2023-01-10
Remove `clock` feature

### LINKED ISSUE #340: Remove the `clock` feature
Ultimately, the `clock` feature adds a dependency on `std` and one method. We can safely assume that anyone that uses `std` will want `clock` as well.

plafer:
Closes: #340 

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -85,7 +85,7 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --no-default-features --features "clock" --all-targets # tests need the clock feature
+          args: --no-default-features --all-targets # tests need the clock feature
 
   test-stable:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -85,7 +85,7 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --no-default-features --all-targets # tests need the clock feature
+          args: --no-default-features
 
   test-stable:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -85,7 +85,7 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


crates/ibc/src/lib.rs
```diff
@@ -11,6 +11,7 @@
     rust_2018_idioms
 )]
 #![forbid(unsafe_code)]
+#![allow(clippy::result_large_err)]
```
plafer, line 15: Note that I created #342 to track the removal of this.

.github/workflows/rust.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --no-default-features --features "clock" --all-targets # tests need the clock feature
+          args: --no-default-features --lib
```
plafer, line 88: We used to run clippy in `no_std` for the tests, which is not a real requirement. And we weren't even doing that right, since `clock` needs `time/std`, which pulls in `std` anyways.

crates/ibc/src/lib.rs
Farhad-Shabani: Nice 👌 
There's just a few warnings that clippy catches in `--all-features` mode because of `#!deny(clippy::uninlined_format_args)]`. Do you mind if you inline below formats, so we only leave with  `result_large_err`?
- https://github.com/cosmos/ibc-rs/blob/aa97ced3f936dd0cb4bf586d340f36af4e5c986a/crates/ibc/src/core/ics02_client/handler/create_client.rs#L125
- https://github.com/cosmos/ibc-rs/blob/aa97ced3f936dd0cb4bf586d340f36af4e5c986a/crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs#L72

plafer: > There's just a few warnings that clippy catches in --all-features mode

I can't reproduce, how exactly did you get these warnings?

Farhad-Shabani: Ran `cargo clippy --all-features --all-targets` without `#![allow(clippy::result_large_err)]`

plafer: I still don't get any warnings; do you want to implement the changes, as they seem pretty minor anyway?

Farhad-Shabani (approved):
🙏

Issue #325, opened 2023-01-05
Refactor `ValidationContext` and `ExecutionContext`


plafer:
Comes out of a comment by @Farhad-Shabani and ensuing sync discussion. We still need to figure out the details, but a main point of confusion was that `validate()` was in the `Context` trait; it's not clear from a first look that all the methods in the `Context` are used in the `validate()` function. We can also think of different ways to separate the methods into a few different traits if that helps with readability, and doesn't duplicate methods.

Blocked on #221.

Below is the original comment:

_Originally posted by @Farhad-Shabani in https://github.com/cosmos/ibc-rs/issues/221#issuecomment-1371403478_

Is it right that we have a tailored validation and execution function for each of the messages like `MsgCreateClient`, `MsgUpdateClient`, etc.? 
Couldn’t we use the following pattern (as a general concept) and implement them on each message type? 
```rust
impl validator<Ctx: RouterContext> {
	pub fn validate(&self, ctx: &mut Ctx) -> Result<(), Error>;
}

impl executor<Ctx: RouterContext> {
	pub fn execute(&self, ctx: &mut Ctx) -> Result<(), Error>;
}

// consider it for MsgCreateClient 
impl validator<MockContext> for MsgCreateClient {...}
impl executor<MockContext> for MsgCreateClient {...}
```
It looks more modular, maintainable, and extensible if it works! (please direct me to the places if I am missing something. tnx)

plafer:
Suggested grouping of methods: https://github.com/cosmos/ibc-rs/pull/386#issuecomment-1429959273

plafer:
We should also keep in mind/revisit the host-based API that was discussed in [ADR 5](https://github.com/cosmos/ibc-rs/blob/main/docs/architecture/adr-005-handlers-redesign.md#host-based-api).

adizere:
We could consider not including this in v1, to avoid ballooning the scope. Let's keep v1 more lean.

Issue #322, opened 2023-01-04
Make `ValidationContext` and `ExecutionContext` implement the ICS-26 router


plafer:
When we wrote ADR 5, we forgot to consider how the ICS-26 router would work with the new `ValidationContext` and `ExecutionContext`. Also, since then, the removal of crossing hellos from connection and channel handshakes changed how certain handlers interact with the router (e.g. we can no longer emit events directly in `ChanOpenTry` and `ChanOpenAck` handlers). This issue serves to amend ADR 5 in order to address all of the above.

The bottom line is: both `validate()` and `execute()` need to perform the relevant parts of the ICS-26 router. As an example, here is what `validate()` and `execute()` would do to handle a `ChanOpenTry` message:

+ `validate()`:
    1. Get the associated `Module` based on the `msg.port_id`
    2. Call the core `chan_open_try::validate()` handler
    3. Query the host for the `channel_id` (through the channel counter)
    4. Call the `Module` callback `on_chan_open_try_validate()` (which only performs validation)

+ `execute()`:
    1. Get the associated `Module` based on the `msg.port_id`
    2. Query the host for the `channel_id` (through the channel counter)
    3. Call the `Module` callback `on_chan_open_try_execute(channel_id, ...)` (which only performs state changes), and save the returned `version`
    4. Call the core `chan_open_try::execute(version, ...)` handler, feeding it the `version` obtained from the `execute()` callback

A few things to note:
+ we need to extend the [`Module` trait](https://github.com/cosmos/ibc-rs/blob/968e553605acacdf799f51745373e08e02365a3b/crates/ibc/src/core/ics26_routing/context.rs#L113) with validation methods (1 per callback).
+ The `channel_id` is no longer generated by the core handler, as specified in the specs (although see [this spec issue](https://github.com/cosmos/ibc/issues/861) which shows that the spec needs to be updated). Instead we generate it in what is effectively the ICS-26 router.

plafer:
Blocked on #361.

PR #347, opened 2023-01-11
Implement ICS-26 routing in `{Validation, Execution}Context` for `ChanOpenTry`

plafer:
Work towards: #322

This implements ICS-26 routing in `{Validation, Execution}Context` for the `ChanOpenTry` message only. After this PR is merged, I'll implement the other messages in a similar manner.

Note that we prioritize here a *working* implementation. We will improve it once a working version is shipped (see #325).

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/context.rs
```diff
@@ -126,7 +163,26 @@ mod val_exec_ctx {
...
                 }
                 .map_err(RouterError::ContextError),
-                MsgEnvelope::Channel(_message) => todo!(),
+                MsgEnvelope::Channel(message) => {
+                    let module_id = self.lookup_module(&message).map_err(ContextError::from)?;
+                    if !self.has_route(&module_id) {
+                        return Err(ChannelError::RouteNotFound)
+                            .map_err(ContextError::ChannelError)
+                            .map_err(RouterError::ContextError);
+                    }
```
Farhad-Shabani, line 173: By having this check here, We'll drop following lines in the future. Right? 
- https://github.com/cosmos/ibc-rs/blob/78297c39ac9a892ee6367dc2b96cdd881013a08a/crates/ibc/src/core/ics04_channel/handler.rs#L67-L77
- https://github.com/cosmos/ibc-rs/blob/aa97ced3f936dd0cb4bf586d340f36af4e5c986a/crates/ibc/src/core/ics04_channel/msgs.rs#L45-L58

crates/ibc/src/core/context.rs
```diff
@@ -99,11 +105,42 @@ mod val_exec_ctx {
     };
     use crate::events::IbcEvent;
     use crate::timestamp::Timestamp;
-    use crate::{prelude::*, Height};
+    use crate::Height;
 
     use super::ContextError;
 
-    pub trait ValidationContext {
+    pub trait Router {
```
Farhad-Shabani, line 113: So, we'll finaly merge the `Router` trait under `ics26_routing` module with this new one? 
https://github.com/cosmos/ibc-rs/blob/380056be894a6edf8b0d9fc23175d92641407a16/crates/ibc/src/core/ics26_routing/context.rs#L248

crates/ibc/src/core/ics04_channel/error.rs
```diff
@@ -19,6 +19,8 @@ pub enum ChannelError {
     Connection(connection_error::ConnectionError),
     /// port error: `{0}`
     Port(port_error::PortError),
+    /// port `{port_id}` is unknown
+    PortNoSource { port_id: PortId },
```
Farhad-Shabani: Couldn't we use the above `Port` variant instead? This looks the same as `UnknownPort`

crates/ibc/src/core/ics04_channel/handler/chan_open_try.rs
```diff
@@ -11,6 +11,112 @@ use crate::core::ics24_host::identifier::ChannelId;
...
+                    &msg.proof_chan_end_on_a,
+                    consensus_state_of_a_on_b.root(),
+                    port_id_on_a,
+                    &chan_id_on_a,
+                    &expected_chan_end_on_a,
+                )
+                .map_err(ChannelError::VerifyChannelFailed)?;
+        }
+
+        let chan_id_on_b = ChannelId::new(ctx_b.channel_counter()?);
```
Farhad-Shabani: IMO, returning `chan_id_on_b` here gives the false impression that it is a result or at least by-product of the validation process. The declaration can be made after the passed "validate" process, in "chan_open_try_validate". Therefore, `validate` would return Ok(())

crates/ibc/src/core/ics04_channel/handler/chan_open_try.rs
```diff
@@ -11,6 +11,112 @@ use crate::core::ics24_host::identifier::ChannelId;
...
+
+        Ok(chan_id_on_b)
+    }
+
+    pub fn execute<Ctx>(ctx_b: &mut Ctx) -> Result<ChannelId, ContextError>
+    where
+        Ctx: ExecutionContext,
+    {
+        let chan_id_on_b = ChannelId::new(ctx_b.channel_counter()?);
```
Farhad-Shabani: The same comment for `validate` applies here, so makes the `execute` no longer necessary in this case...

crates/ibc/src/core/context.rs
```diff
@@ -126,7 +163,26 @@ mod val_exec_ctx {
...
+                    if !self.has_route(&module_id) {
+                        return Err(ChannelError::RouteNotFound)
+                            .map_err(ContextError::ChannelError)
+                            .map_err(RouterError::ContextError);
+                    }
+
+                    match message {
+                        ChannelMsg::OpenInit(_) => todo!(),
+                        ChannelMsg::OpenTry(message) => {
+                            chan_open_try_validate(self, module_id, message)
```
Farhad-Shabani, line 178: I thought how about calling `chan_open_try::validate` directly instead of adding another layer of calls by `chan_open_try_validate`? (or just keeping `chan_open_try_validate`) 
Can't we merge these two somehow? The reasoning behind is not clear, making the code a bit hard to read and understand.

Farhad-Shabani (commented):
Thank you @plafer. 
I'll probably revisit this PR after getting your response so I can absorb the design. Hope it doesn't cause you any delays.

crates/ibc/src/core/context.rs
plafer: yep!

crates/ibc/src/core/context.rs
plafer: We will be removing the one in 'ics26_routing' completely

crates/ibc/src/core/ics04_channel/error.rs
plafer: yes good point; I wanted to avoid using `PortError` because I think we should get rid of it, but this is better for now

[c7f10ad](https://github.com/cosmos/ibc-rs/pull/347/commits/c7f10adfa0024037669515b85f5936c4997fb425)

crates/ibc/src/core/ics04_channel/handler/chan_open_try.rs
plafer: Good point! The original intention was that making sure that the `channel_counter()` call succeeds is part of validation; however it's minor and I much prefer this too.

[6047f02](https://github.com/cosmos/ibc-rs/pull/347/commits/6047f02d2cce8b9661f0bcb85fd0dbc68928126e)

crates/ibc/src/core/ics04_channel/handler/chan_open_try.rs
plafer: [28d04f0](https://github.com/cosmos/ibc-rs/pull/347/commits/28d04f0fe09cd12dc31897a2b4c340c80c0d90f9)

crates/ibc/src/core/context.rs
plafer: The current design is based on the standard: `chan_open_try_validate()` is the validation part of the [ICS-26 router function](https://github.com/cosmos/ibc/blob/main/spec/core/ics-026-routing-module/README.md#channel-lifecycle-management) (which is currently incorrect in the spec), and `chan_open_try::validate()` is the validation part of the [core handler](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/README.md#opening-handshake).

In other words:
- `chan_open_try_validate()` calls both the core handler and the application/module callback (getting the channel id from the core handler,  and the channel version from the callback)
- `chan_open_try::validate()` is the core handler validation logic

Does this make sense? Do you still think it would be better to move away from this design? At the very least, using better naming to encode what I just explained would help.

Farhad-Shabani: On this basis, `chan_open_try_validate` operates more as a handler than a validator, shouldn't contain any validation step. Also, the latter link you sent seems to correspond more with module interfaces iiuc (e.g. `on_chan_open_try_validate`).
Btw, I don't think naming help with this issue. I propose all checkings be done in one place either in `chan_open_try::validate()` or `on_chan_open_try_validate`, like what we have now for `on_chan_open_try_execute` or like before structure for `validate()` `execute()`

Farhad-Shabani: Maybe not bad, keeping this change, and see what the implications will be when we refactor others. Beside this, it LGTM and your explanations are great, thanks 🙏

plafer: Indeed I think we should leave it as is for now because correctness doesn't seem to be a concern; we will focus on improving readability/maintainability once `Validation/ExecutionContext` have shipped. WDYT?

Farhad-Shabani: Agree, let's keep it for now 👍

Farhad-Shabani (approved):
👌

PR #349, opened 2023-01-13
Implement `verify_upgrade_and_update_state` for Tendermint Client

### LINKED ISSUE #19: `TendermintClient::verify_upgrade_and_update_state` is not implemented
Related to: https://github.com/informalsystems/ibc-rs/issues/722

## Summary

The `TendermintClient::verify_upgrade_and_update_state` is currently not implemented, even though it is called in the UpgradeClient handler.

https://github.com/informalsystems/ibc-rs/blob/efbf7d18512fe6efcb593de68735b7f8783fd652/modules/src/clients/ics07_tendermint/client_def.rs#L395-L403

https://github.com/informalsystems/ibc-rs/blob/efbf7d18512fe6efcb593de68735b7f8783fd652/modules/src/core/ics02_client/handler/upgrade_client.rs#L52-L57
## Problem Definition

The UpgradeClient is therefore non-functional at the moment, as it will panic once the method is called.

## Proposal

Implement the `verify_upgrade_and_update_state` method for `TendermintClient`.

## Acceptance Criteria

The `verify_upgrade_and_update_state` is implemented in `TendermintClient`, and its behavior is captured in a test for the UpgradeClient handler.

____

#### For Admin Use

- [ ] Not duplicate issue

Farhad-Shabani:
Closes: #19

## Summary

This PR handles unimplemented `verify_upgrade_and_update_state` method for Tendermint clients.  

## Remarks

- `verify_upgrade_and_update_state` has been kept the same as IBC-Go's so far, but the processes it perform do not match the ongoing ADR05 change and requires both **updating** and **verifying** in one operation. Here, it's broken down into `verify_upgrade_client` and `update_state_with_upgrade_client`.
Similarly, there are some other methods within `ClientState` trait like `new_check_header_and_update_state` that can be refactored to have clearer distinction. In particular, when it comes to implement `pub(crate) fn validate()` and `pub(crate) fn execute()` functions.

- `upgrade_client` message tests are done with `MockContext`, which does not cover every aspect of this change. The main purpose of this handler is to serve Tendermint clients. A separate PR or this could be covered here ?!

_____

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -876,13 +879,145 @@ impl Ics2ClientState for ClientState {
...
+
+        // The new consensus state is merely used as a trusted kernel against which headers on the new
+        // chain can be verified. The root is just a stand-in sentinel value as it cannot be known in advance, thus no proof verification will pass.
+        // The timestamp and the NextValidatorsHash of the consensus state is the blocktime and NextValidatorsHash
+        // of the last block committed by the old chain. This will allow the first block of the new chain to be verified against
+        // the last validators of the old chain so long as it is submitted within the TrustingPeriod of this client.
+        // NOTE: We do not set processed time for this consensus state since this consensus state should not be used for packet verification
+        // as the root is empty. The next consensus state submitted using update will be usable for packet-verification.
+        let new_consensus_state = TmConsensusState::new(
+            vec![].into(),
```
plafer: I don't have a deep understanding of how the SDK does upgrades. For me, many questions need to be answered here. Why does it make sense to put an empty root here? What is the behavior of the next `update_client` when it sees this root? Is the host expected to do anything else (that would normally be provided by the SDK)? Does the SDK's strategy for chain upgrades make sense for ibc-rs, given that our users typically have radically different architectures from the SDK?

This set of questions is why I [initially suggested](https://github.com/cosmos/ibc-rs/issues/141#issuecomment-1349633869) we simply output an error for now, and figure out the details later. However, maybe you have a better understanding of this than I do at this point? If so, then perhaps we'd need to write an ADR about this to explain why we designed our upgrade API like so, and answer the questions I asked above.

WDYT?

Farhad-Shabani: Actually, I followed very similar logic to what has been implemented on the `ibc-go` with some changes to make it fit our context. It is, of course, largely compatible with SDK chains, but I think most of the validation checkings (even all) and some of the update-related stuff apply also here. 
But, I understand what you mean. there are some parts (as you truly mentioned) need further investigation. 
If the decision would be to postpone the implementation until we have some real cases, that's a different story. Though, I propose that, even in its minimal shape, to have a kind of implementation as I think future cases may differ considerably...
Thus, if we agree on its existence, I will work on the ADR. We can at least come up with some clear explanations

plafer: > If the decision would be to postpone the implementation until we have some real cases, that's a different story. Though, I propose that, even in its minimal shape, to have a kind of implementation as I think future cases may differ considerably...

I'm okay with having an implementation right now, as long as it is self-consistent and well-documented. Which is the perfect use case for an ADR.

> Thus, if we agree on its existence, I will work on the ADR. We can at least come up with some clear explanations

Yep, if you're interested, then please write an ADR! We can discuss the specifics there.

Farhad-Shabani: Thanks for bringing this up. 
Alright, I'll go for it 👍

Farhad-Shabani: Issue #362 - PR #383

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -876,13 +854,161 @@ impl Ics2ClientState for ClientState {
...
+
+        let last_height = self.latest_height().revision_height();
+
+        // Construct the merkle path for the client state
+        let mut client_upgrade_path = upgrade_path.clone();
+        client_upgrade_path.push(ClientUpgradePath::UpgradedClientState(last_height).to_string());
+
+        let client_upgrade_merkle_path = MerklePath {
+            key_path: client_upgrade_path,
+        };
```
plafer: I'm concerned that we're not building the right path, comparing with [ibc-go](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L129-L141). Can you confirm/disconfirm this?

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -876,13 +854,161 @@ impl Ics2ClientState for ClientState {
...
+            )
+            .map_err(ClientError::Ics23Verification)?;
+
+        // Construct the merkle path for the consensus state
+        let mut cons_upgrade_path = upgrade_path;
+        cons_upgrade_path
+            .push(ClientUpgradePath::UpgradedClientConsensusState(last_height).to_string());
+        let cons_upgrade_merkle_path = MerklePath {
+            key_path: cons_upgrade_path,
+        };
```
plafer, line 933: Similarly here, this doesn't seem to be [the same as ibc-go](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L144-L156).

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -876,13 +854,161 @@ impl Ics2ClientState for ClientState {
...
+
+        // Construct the merkle path for the client state
+        let mut client_upgrade_path = upgrade_path.clone();
+        client_upgrade_path.push(ClientUpgradePath::UpgradedClientState(last_height).to_string());
+
+        let client_upgrade_merkle_path = MerklePath {
+            key_path: client_upgrade_path,
+        };
+
+        upgraded_tm_client_state.zero_custom_fields();
```
plafer, line 911: why are we zeroing some fields here? [ibc-go doesn't seem](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L75-L84) to be doing that

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -876,13 +854,161 @@ impl Ics2ClientState for ClientState {
...
+        let new_client_state = TmClientState::new(
+            upgraded_tm_client_state.chain_id,
+            self.trust_level,
+            self.trusting_period,
+            upgraded_tm_client_state.unbonding_period,
+            self.max_clock_drift,
+            upgraded_tm_client_state.latest_height,
+            upgraded_tm_client_state.proof_specs,
+            upgraded_tm_client_state.upgrade_path,
+            upgraded_tm_client_state.allow_update,
```
plafer: should be `self.allow_update` according to [ibc-go](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L104)

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -876,13 +854,161 @@ impl Ics2ClientState for ClientState {
...
+            upgraded_tm_client_state.frozen_height,
```
plafer: We should put `None` directly here, [as ibc-go](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/client_state.go#L38). 

But make it a variable for better readability:
```rust
let new_frozen_height = None;
let new_client_state = ...;
```

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -63,12 +63,7 @@ pub trait ClientState:
     /// Helper function to verify the upgrade client procedure.
     /// Resets all fields except the blockchain-specific ones,
     /// and updates the given fields.
-    fn upgrade(
-        &mut self,
-        upgrade_height: Height,
-        upgrade_options: &dyn UpgradeOptions,
-        chain_id: ChainId,
-    );
+    fn zero_custom_fields(&mut self);
```
plafer, line 66: Assuming we need this (see previous comments), I don't think this should be in the ICS-2 interface. Can't it just be a private function in the tendermint client module?

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
```diff
@@ -35,22 +35,50 @@ where
...
-    // Read client state from the host chain store.
+    // Read the current latest client state from the host chain store.
     let old_client_state = ctx.client_state(&client_id)?;
 
+    // Check if the client is frozen.
     if old_client_state.is_frozen() {
-        return Err(ClientError::ClientFrozen { client_id }.into());
+        return Err(ContextError::ClientError(ClientError::ClientFrozen {
+            client_id,
+        }));
```
plafer, line 52: ibc-go doesn't check if the client is frozen (unless I missed it). There probably is a good reason for it, such as allowing a chain to "restart" clients on other chains? Not sure.

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
```diff
@@ -35,22 +35,50 @@ where
...
+
+    // Check if the latest consensus state is within the trust period.
+    if old_client_state.expired(duration) {
+        return Err(ContextError::ClientError(
+            ClientError::HeaderNotWithinTrustPeriod {
+                latest_time: old_consensus_state.timestamp(),
+                update_time: now,
+            },
+        ));
+    };
```
plafer, line 79: ibc-go also doesn't seem to have that check?

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: When a chain creates an `UpgradedClientState` through the upgrade proposal, it should zero out all the cleint-customizable fields ([Ref](https://github.com/cosmos/ibc-go/blob/main/docs/ibc/upgrades/quick-guide.md#step-by-step-upgrade-process-for-sdk-chains)). The upgrade client proof is for that zeroed-out client state, and to validate against, we also should zero out respective fields of `UpgradedClientState`. 
Check [here](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/core/keeper/msg_server_test.go#L678) and [here](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/light-clients/07-tendermint/upgrade.go#L76) please

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: Good catch, mended! [5bc1a04](https://github.com/cosmos/ibc-rs/pull/349/commits/5bc1a049097a2f0ce0dd6160f6cbc0d26b04d77a)

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: Fixed [5bc1a04](https://github.com/cosmos/ibc-rs/pull/349/commits/5bc1a049097a2f0ce0dd6160f6cbc0d26b04d77a)

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: Considering that in any chain and client state struct there will be some relayer customizable fields that are needed to be zeroed out for the upgrade process. I think it makes sense to keep it here

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: Quite understandable. Do you think we can keep this checked with issue #385?

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: First, I coded this the same as ibc-go, but then found out on our side we access something simpler. So, one difference here is that I tried to use the internal structs to `ibc-rs`, and created the path with `ClientUpgradePath::UpgradedClientConsensusState(last_height)`. 
I didn't find such an instance on `ibc-go`
But truly it should be verified with tests

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
Farhad-Shabani: Good point, Could be the reason! 
So, upgrade process would also allow restarting a frozen client, but I am not sure if it is a safe permission! 🤔

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
Farhad-Shabani: This one also is the same as "Client frozen" matter. 
Dropping this enables a client that is "not within the trusting period" to restart.

Farhad-Shabani: I think we should answer this question:
Is there any case that by "having a frozen client" or "exceeding the allowed trusting period" we also must not preserve its connections? (in any cost)

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: > So, one difference here is that I tried to use the internal structs to ibc-rs, and created the path with ClientUpgradePath::UpgradedClientConsensusState(last_height)

This part of ibc-rs hasn't been touched in a while and it is likely outdated and/or incorrect (like a lot of things were just a few months ago). If an internal struct is incorrect or outdated, we should just change it.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: If were that unsure about the current implementation, we should then put it behind a feature flag (similar to `val_exec_ctx`), which will allow us to test it with basecoin-rs, and at the same time, signal to users that this feature shouldn't be used yet.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: Interesting, your references point to the `main` branch, [which changed since v5.0.1](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L76), the version we are targeting compatibility with (notice how they don't zero the fields there). I'm not sure which version we should target anymore for ibc-rs chains to be compatible with ibc-go chains (as it depends on the version of ibc-go that the chain is using).

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
Farhad-Shabani: #391

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
Farhad-Shabani: #391

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: #391

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: #391

crates/ibc/Cargo.toml
```diff
@@ -17,7 +17,7 @@ description  = """
 all-features = true
 
 [features]
-default = ["std"]
+default = ["std", "disable_upgrade_client"]
```
plafer: we should instead have the feature be `upgraded_client`, and _if enabled_, we provide the unfinished implementation. This is a subtle but necessary change: `no_std` users will build with `default-features=false`, which would "enable" the upgrade client code.

Shouldn't be a big change; feature checks just need to be flipped
```rust
cfg!(not(feature = "upgrade_client"))
```

crates/ibc/Cargo.toml
```diff
@@ -45,6 +45,9 @@ serde = ["dep:serde", "dep:serde_derive", "serde_json", "erased-serde"]
 # This feature guards the unfinished implementation of ADR 5.
 val_exec_ctx = []
 
+# This feature disables the upgrade client handler and return an error when a client upgrade is attempted.
```
plafer:
```suggestion
# This feature guards the unfinished implementation of the `UpgradeClient` handler.
```

Let's make it clear that this is not a feature users should be using.

plafer (changes requested):

crates/ibc/Cargo.toml
Farhad-Shabani: Done [83de030](https://github.com/cosmos/ibc-rs/pull/349/commits/83de030a251baf740659d3c8cb1612ad01cc2b81)

crates/ibc/Cargo.toml
Farhad-Shabani: [83de030](https://github.com/cosmos/ibc-rs/pull/349/commits/83de030a251baf740659d3c8cb1612ad01cc2b81)

plafer (approved):
👌

PR #352, opened 2023-01-13
`{Validation,Execution}Context` complete ICS-26 routing for channel handshakes

plafer:
Work towards: #322

## Description

Completes ICS-26 routing for channel handshakes.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/applications/transfer/context.rs
```diff
@@ -374,6 +332,171 @@ pub fn on_timeout_packet(
     Ok(())
 }
 
+#[cfg(feature = "val_exec_ctx")]
```
Farhad-Shabani, line 335: Thanks for moving and separating this part from the rest :ok_hand:

crates/ibc/src/applications/transfer/context.rs
```diff
@@ -374,6 +332,171 @@ pub fn on_timeout_packet(
...
+        }
+
+        if !version.is_empty() && version != &Version::new(VERSION.to_string()) {
+            return Err(TokenTransferError::InvalidVersion {
+                expect_version: Version::new(VERSION.to_string()),
+                got_version: version.clone(),
+            });
+        }
+
+        Ok(Version::new(VERSION.to_string()))
```
Farhad-Shabani: Can we return `Ok(())` here and let the version be determined from the execution part? I think it makes more sense (to extend it's possible) to keep `*_*_validate()` methods return `Ok(())`. This way, some developers in practice might not use validate's output, others use execute's.

crates/ibc/src/applications/transfer/context.rs
```diff
@@ -374,6 +332,171 @@ pub fn on_timeout_packet(
...
+            });
+        }
+        if counterparty_version != &Version::new(VERSION.to_string()) {
+            return Err(TokenTransferError::InvalidCounterpartyVersion {
+                expect_version: Version::new(VERSION.to_string()),
+                got_version: counterparty_version.clone(),
+            });
+        }
+
+        Ok(Version::new(VERSION.to_string()))
```
Farhad-Shabani: Similarly, return 'Ok(())' applies here as well

crates/ibc/src/core/context.rs
```diff
@@ -559,6 +571,105 @@ mod val_exec_ctx {
...
+    where
+        ValCtx: ValidationContext,
+    {
+        chan_open_init::validate(ctx_a, &msg)?;
+        let chan_id_on_a = ChannelId::new(ctx_a.channel_counter()?);
+
+        let module = ctx_a
+            .get_route(&module_id)
+            .ok_or(ChannelError::RouteNotFound)?;
+        let _ = module.on_chan_open_init_validate(
```
Farhad-Shabani: can we drop `let _ = `?

crates/ibc/src/core/context.rs
```diff
@@ -143,48 +153,50 @@ mod val_exec_ctx {
...
         where
             Self: Sized,
         {
-            match message {
-                MsgEnvelope::Client(message) => match message {
-                    ClientMsg::CreateClient(message) => create_client::validate(self, message),
-                    ClientMsg::UpdateClient(message) => update_client::validate(self, message),
-                    ClientMsg::Misbehaviour(message) => misbehaviour::validate(self, message),
-                    ClientMsg::UpgradeClient(message) => upgrade_client::validate(self, message),
+            match msg {
```
Farhad-Shabani, line 160: Liked turning these to `msg`

crates/ibc/src/core/ics26_routing/context.rs
```diff
@@ -110,6 +110,30 @@ impl OnRecvPacketAck {
 pub type ModuleOutputBuilder = HandlerOutputBuilder<(), ModuleEvent>;
 
 pub trait Module: Send + Sync + AsAnyMut + Debug {
+    #[cfg(feature = "val_exec_ctx")]
+    #[allow(clippy::too_many_arguments)]
+    fn on_chan_open_init_validate(
```
Farhad-Shabani, line 115: As a side note: upon splitting the validation and execution parts, it might not be necessary to have all the fields(args) for both. Perhaps a later investigation would be helpful if we could get rid of some. Otherwise, why we don't pass the own `msg` to the these methods instead of flattening the fields? It seems just the `Signer` fields is the one not used anywhere. Is there any reservation for `Signer`?

crates/ibc/src/core/context.rs
```diff
@@ -612,10 +719,13 @@ mod val_exec_ctx {
         )?;
 
         let conn_id_on_b = msg.connection_hops_on_b[0].clone();
-        let port_channel_id_on_b = (msg.port_id_on_b.clone(), chan_id_on_b.clone());
 
         // emit events and logs
```
Farhad-Shabani, line 747: Would the logged events be reverted if something fails in the next `state change` step? is this a possible case?

crates/ibc/src/core/context.rs
```diff
@@ -635,16 +745,18 @@ mod val_exec_ctx {
             }
         }
 
+        // state changes
```
Farhad-Shabani: Some of the parts in `emit events and logs` and `state changes` sections looks similar and gatherable in second functions (as a later work)

Farhad-Shabani (commented):
I just left a few suggestions and some questions for myself to better understand some angles. thanks :pray:

crates/ibc/src/applications/transfer/context.rs
plafer: Good point, I agree. [d6fc4c9](https://github.com/cosmos/ibc-rs/pull/352/commits/d6fc4c965d2756449d2933968648f79b86be2aaf)

crates/ibc/src/applications/transfer/context.rs
plafer: [d6fc4c9](https://github.com/cosmos/ibc-rs/pull/352/commits/d6fc4c965d2756449d2933968648f79b86be2aaf)

crates/ibc/src/core/context.rs
plafer: [d6fc4c9](https://github.com/cosmos/ibc-rs/pull/352/commits/d6fc4c965d2756449d2933968648f79b86be2aaf)

crates/ibc/src/core/ics26_routing/context.rs
plafer: > upon splitting the validation and execution parts, it might not be necessary to have all the fields(args) for both.

I'd be careful removing some field because we "think" they might not be useful. We have to remember that these fields might be used in ways that we don't foresee by future IBC applications. The current list is [based on the standard](https://github.com/cosmos/ibc/blob/main/spec/core/ics-026-routing-module/README.md#onchanopeninit), and I don't think we should arbitrarily restrict future IBC apps by passing less data. Also, intuitively, I could see an app needing to check every field in validation, and storing every field in execution.

> Otherwise, why we don't pass the own msg to the these methods instead of flattening the fields?

Good question. We don't pass the message directly because not all fields come from the message (e.g. the [channel id](https://github.com/cosmos/ibc-rs/blob/d6fc4c965d2756449d2933968648f79b86be2aaf/crates/ibc/src/core/context.rs#L592)). So although the message and callbacks are very similar, they should be thought of as distinct. We could however create a `struct` that encompasses all these fields as opposed to passing them one by one and needing `#[allow(clippy::too_many_arguments)]`.

WDYT?

crates/ibc/src/core/context.rs
plafer: Good point [d65553c](https://github.com/cosmos/ibc-rs/pull/352/commits/d65553c923e4d30b54796c8e2168751bf75111b4)

crates/ibc/src/core/context.rs
plafer: Indeed, we can include that in the work to be done with #325

Farhad-Shabani: Sure 👍

crates/ibc/src/core/ics26_routing/context.rs
Farhad-Shabani: Thank you for the clarifications. 
I also thought of that `struct` creation, but I was concerned that it may end up in multiple structs that might not offer much more than the current design. 
Let's put it off for now and keep it in mind for a later thought :)

Farhad-Shabani (approved):

Issue #14, opened 2022-06-29
Allow the host implementation to specify the `generateIdentifier` function for connection and channel identifiers


hu55a1n1:
## Summary of Bug
We currently define trait methods like `ConnectionReader::connection_counter()` and use them to generate the connection identifier -> 
https://github.com/informalsystems/ibc-rs/blob/57b8af939eedab3a5ad55d7045a108ed0dfab1f6/modules/src/core/ics03_connection/handler/conn_open_init.rs#L42-L44

This restricts host implementations' ability to generate custom identifiers and is a deviation from the spec. 

## Version
v0.16.0

## Acceptance Criteria
Either provide a way for host impls to specify their own `generateIdentifier` or document this as a divergence (in [divergence from the ICS](https://github.com/informalsystems/ibc-rs/tree/master/modules#divergence-from-the-interchain-standards-ics)).

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

hu55a1n1:
Related to informalsystems/ibc-rs-tmp#39

Farhad-Shabani:
### Closer Look
Having a closer look at the related part of the specs:
Following clauses **ALLOW** users to define custom identifiers:

> [ICS24](https://github.com/cosmos/ibc/blob/17d62c054af4ec2d6a18ea4ae37c4e6fe50b5fdc/spec/core/ics-024-host-requirements/README.md#paths-identifiers-separators): Identifiers are not intended to be valuable resources — to prevent name squatting, minimum length requirements or **pseudorandom generation** MAY be implemented, but particular restrictions are not imposed by this specification.

> [ICS02](https://github.com/cosmos/ibc/tree/17d62c054af4ec2d6a18ea4ae37c4e6fe50b5fdc/spec/core/ics-002-client-semantics#identifier-validation): Clients are stored under a unique `Identifier` prefix. This ICS does not require that client identifiers be generated in a particular manner, only that they be unique. However, it is possible to restrict the space of `Identifier`s if required.

[...]

PR #360, opened 2023-01-19
Additional tests for counterparty conn & chan ids at initialization

### LINKED ISSUE #175: Investigate: `Connection`'s `RawCounterparty.connection_id` can be `""` in the spec, but not our impl
This issues serves as a reminder to investigate the following.

The spec actually allows `ConnectionEnd.counterpartyConnectionIdentifier` to be `""` (even though this is an illegal value for `Identifier`) - see [issue](https://github.com/cosmos/ibc/issues/854). Our implementation [disallows empty strings](https://github.com/cosmos/ibc-rs/blob/da6653e0716350311d15e6acca08cad625f2d9ff/crates/ibc/src/core/ics03_connection/connection.rs#L260) when converting from `RawCounterparty` -> `Counterparty`. This probably fine and due to the interplay between ibc-go's proto files [using `(gogoproto.nullable) = false`](https://github.com/cosmos/ibc-go/blob/c8e2ebcf2eaa6b487ca52499063c7c3db4253398/proto/ibc/core/connection/v1/connection.proto#L27), and how prost's code generation. We also successfully opened connections with basecoin-rs & hermes, which tells me it's right. 

However, I want to convince myself that this code is correct, and document the reasoning somewhere.

Farhad-Shabani:
Closes: #175

## Summary

Ensure that disallows empty strings for counterparty connection and channel ids work correctly at the initialization phases (`conn_open_init` and `chan_open_init`)

## Description

Firstly, It looks like the ICS24_host statement around identifiers implicitly refers to successfully made (open) connections. However, it’s true... it does not consider the transient state of the ***ends** during the opening process.

Btw, I also tested the opening of a connection using `basecoin-rs` and `Hermes` and it worked properly. 
The counterparty connection id was set to None for the sent `MsgConnectionOpenInit`. 

In this PR, a few comments and tests have been added to check the corner scenarios. 
Also, while skimming the codes, I changed the naming of the `client_counter`, `connection_counter`, and `channel_counter`  methods to correspond with the exact work they do. Counting is their side job, but they're basically used for identifier creation all over the repository. This way it would also align with IBC-go. Though… take it as a suggestion.

______

### PR author checklist:
- [X] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/context.rs
```diff
@@ -255,7 +255,7 @@ mod val_exec_ctx {
 
         /// Returns a natural number, counting how many clients have been created thus far.
         /// The value of this counter should increase only via method `ClientKeeper::increase_client_counter`.
-        fn client_counter(&self) -> Result<u64, ContextError>;
+        fn generate_client_identifier(&self) -> Result<u64, ContextError>;
```
plafer: Note that we have an issue (#14) that tracks this. But also if the function were to be named `generate_client_identifier`, I would expect it to return a `ClientId` as opposed to an "id counter".

Can we make that change in another PR that addresses #14, in order to have self-contained PRs?

Farhad-Shabani: Didn't notice that issue. 
Sure. Will do so

plafer:
Note that I simplified the expression that [confused me previously](https://github.com/cosmos/ibc-rs/issues/175#issuecomment-1397657808). Let me know if you also find that cleaner.

[70c6c0d](https://github.com/cosmos/ibc-rs/pull/360/commits/70c6c0dc2b2f680b3450830d632a7d9bfb2397e0)
[30727a0](https://github.com/cosmos/ibc-rs/pull/360/commits/30727a02a9017eecd16984062d25d8c8b06036a0)

crates/ibc/src/core/context.rs
```diff
@@ -705,6 +705,9 @@ mod val_exec_ctx {
         ExecCtx: ExecutionContext,
     {
         let chan_id_on_b = ChannelId::new(ctx_b.channel_counter()?);
+        ctx_b.log_message(format!(
```
plafer: Notice that we already write this log message in the `// emit events and logs` section. The reason is that we want to make sure every state change succeeds before we log the success

crates/ibc/src/core/ics02_client/context.rs
```diff
@@ -68,7 +68,7 @@ pub trait ClientReader {
     /// Returns the pending `ConsensusState` of the host (local) chain.
     fn pending_host_consensus_state(&self) -> Result<Box<dyn ConsensusState>, ClientError>;
 
-    /// Returns a natural number, counting how many clients have been created thus far.
+    /// Returns a natural number, counting how many clients have been created thus far..
```
plafer:
```suggestion
    /// Returns a natural number, counting how many clients have been created thus far.
```

crates/ibc/src/core/ics03_connection/context.rs
```diff
@@ -70,8 +70,7 @@ pub trait ConnectionReader {
     }
 
     /// Returns a counter on how many connections have been created thus far.
-    /// The value of this counter should increase only via method
-    /// `ConnectionKeeper::increase_connection_counter`.
+    /// The value of this counter increases only via method `ConnectionKeeper::increase_connection_counter`.
```
plafer: I think the comment was correct before; we're telling the host that they should never change that value other than with `ConnectionKeeper::increase_connection_counter`

crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs
```diff
@@ -58,6 +58,8 @@ where
         msg.client_id_on_a.clone(),
         Counterparty::new(
             msg.counterparty.client_id().clone(),
+            // Note: the counterparty connection Id passed by the relayer alternated to `None`,
+            // left for the `conn_open_try` process to set
```
plafer: I don't think we need this comment; we're only following the standard here.

Farhad-Shabani:
> Note that I simplified the expression that [confused me previously](https://github.com/cosmos/ibc-rs/issues/175#issuecomment-1397657808). Let me know if you also find that cleaner.
> 
> [70c6c0d](https://github.com/cosmos/ibc-rs/pull/360/commits/70c6c0dc2b2f680b3450830d632a7d9bfb2397e0) [30727a0](https://github.com/cosmos/ibc-rs/pull/360/commits/30727a02a9017eecd16984062d25d8c8b06036a0)

Now it is easier to catch 👍

crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs
```diff
@@ -118,6 +120,8 @@ pub(crate) fn process(
         msg.client_id_on_a.clone(),
         Counterparty::new(
             msg.counterparty.client_id().clone(),
+            // Note: the counterparty connection Id passed by the relayer alternated to `None`,
```
plafer: Same, I don't think we need this comment

crates/ibc/src/core/ics04_channel/context.rs
```diff
@@ -157,8 +157,7 @@ pub trait ChannelReader {
     ) -> Result<Height, ChannelError>;
 
     /// Returns a counter on the number of channel ids have been created thus far.
-    /// The value of this counter should increase only via method
-    /// `ChannelKeeper::increase_channel_counter`.
+    /// The value of this counter increases only via method `ChannelKeeper::increase_channel_counter`.
```
plafer: Similarly, the previous comment was more accurate

crates/ibc/src/core/ics04_channel/handler/chan_open_init.rs
```diff
@@ -74,6 +74,8 @@ pub(crate) fn process<Ctx: ChannelReader>(
     let chan_end_on_a = ChannelEnd::new(
         State::Init,
         msg.ordering,
+        // Note: the counterparty channel Id passed by the relayer alternated to `None`,
+        // left for the `chan_open_try` handler to set
```
plafer: Similarly not needed

plafer (changes requested):
🙏 thank you for all these tests!

crates/ibc/src/core/ics04_channel/handler/chan_open_init.rs
Farhad-Shabani: [e2ea605](https://github.com/cosmos/ibc-rs/pull/360/commits/e2ea6054fee2be83cd9421fce8b0b21bd92e754b)

crates/ibc/src/core/ics04_channel/context.rs
Farhad-Shabani: [e2ea605](https://github.com/cosmos/ibc-rs/pull/360/commits/e2ea6054fee2be83cd9421fce8b0b21bd92e754b)

crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs
Farhad-Shabani: [e2ea605](https://github.com/cosmos/ibc-rs/pull/360/commits/e2ea6054fee2be83cd9421fce8b0b21bd92e754b)

crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs
Farhad-Shabani: [e2ea605](https://github.com/cosmos/ibc-rs/pull/360/commits/e2ea6054fee2be83cd9421fce8b0b21bd92e754b)

crates/ibc/src/core/ics03_connection/context.rs
Farhad-Shabani: [e2ea605](https://github.com/cosmos/ibc-rs/pull/360/commits/e2ea6054fee2be83cd9421fce8b0b21bd92e754b)

crates/ibc/src/core/ics02_client/context.rs
Farhad-Shabani: [e2ea605](https://github.com/cosmos/ibc-rs/pull/360/commits/e2ea6054fee2be83cd9421fce8b0b21bd92e754b)

crates/ibc/src/core/context.rs
Farhad-Shabani: [e2ea605](https://github.com/cosmos/ibc-rs/pull/360/commits/e2ea6054fee2be83cd9421fce8b0b21bd92e754b)

plafer (approved):
🎉

PR #363, opened 2023-01-20
Add tests to verify `AbciEvent` match emitted events

### LINKED ISSUE #163: Add tests that verify the `AbciEvent` keys match the events actually emitted by ibc-go.
Would be nice to have tests that verify the `AbciEvent` keys match the events actually emitted by ibc-go. (We could do this in a later PR.)

_Originally posted by @hu55a1n1 in https://github.com/cosmos/ibc-rs/pull/138#discussion_r992316683_

Farhad-Shabani:
Closes: #163
______

### PR author checklist:
- [X] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/ics02_client/events.rs
+++ b/crates/ibc/src/core/ics02_client/events.rs
@@ -394,7 +394,7 @@ impl From<UpgradeClient> for abci::Event {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::mock::client_state::client_type as mock_client_type;
+    use crate::mock::client_state::{client_type as mock_client_type, MOCK_CLIENT_TYPE};
     use crate::mock::header::MockHeader;
     use ibc_proto::google::protobuf::Any;
     use tendermint::abci::Event as AbciEvent;
@@ -405,6 +405,7 @@ mod tests {
             kind: IbcEventType,
             event: AbciEvent,
             expected_keys: Vec<&'static str>,
+            expected_values: Vec<String>,
         }
 
         let client_type = mock_client_type();
@@ -419,6 +420,17 @@ mod tests {
             CONSENSUS_HEIGHTS_ATTRIBUTE_KEY,
             HEADER_ATTRIBUTE_KEY,
         ];
+        let expected_values = vec![
+            client_id.to_string(),
+            MOCK_CLIENT_TYPE.to_string(),
+            consensus_height.to_string(),
+            consensus_heights
+                .iter()
+                .map(|h| h.to_string())
+                .collect::<Vec<_>>()
+                .join(","),
+            String::from_utf8(hex::encode(header.clone().value)).unwrap(),
+        ];
 
         let tests: Vec<Test> = vec![
             Test {
@@ -426,6 +438,7 @@ mod tests {
                 event: CreateClient::new(client_id.clone(), client_type.clone(), consensus_height)
                     .into(),
                 expected_keys: expected_keys[0..3].to_vec(),
+                expected_values: expected_values[0..3].to_vec(),
             },
             Test {
                 kind: IbcEventType::UpdateClient,
@@ -438,17 +451,20 @@ mod tests {
                 )
                 .into(),
                 expected_keys: expected_keys.clone(),
+                expected_values: expected_values.clone(),
             },
             Test {
                 kind: IbcEventType::UpgradeClient,
                 event: UpgradeClient::new(client_id.clone(), client_type.clone(), consensus_height)
                     .into(),
                 expected_keys: expected_keys[0..3].to_vec(),
+                expected_values: expected_values[0..3].to_vec(),
             },
             Test {
                 kind: IbcEventType::ClientMisbehaviour,
                 event: ClientMisbehaviour::new(client_id, client_type).into(),
                 expected_keys: expected_keys[0..2].to_vec(),
+                expected_values: expected_values[0..2].to_vec(),
             },
         ];
 
@@ -456,7 +472,18 @@ mod tests {
             assert_eq!(t.event.kind, t.kind.as_str());
             assert_eq!(t.expected_keys.len(), t.event.attributes.len());
             for (i, key) in t.expected_keys.iter().enumerate() {
-                assert_eq!(t.event.attributes[i].key, *key);
+                assert_eq!(
+                    t.event.attributes[i].key, *key,
+                    "key mismatch for {:?}",
+                    t.kind
+                );
+            }
+            for (i, value) in t.expected_values.iter().enumerate() {
+                assert_eq!(
+                    t.event.attributes[i].value, *value,
+                    "value mismatch for {:?}",
+                    t.kind
+                );
             }
         }
     }
--- a/crates/ibc/src/core/ics02_client/events.rs
+++ b/crates/ibc/src/core/ics02_client/events.rs
@@ -394,7 +394,6 @@ impl From<UpgradeClient> for abci::Event {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::mock::client_state::client_type as mock_client_type;
     use crate::mock::header::MockHeader;
     use ibc_proto::google::protobuf::Any;
     use tendermint::abci::Event as AbciEvent;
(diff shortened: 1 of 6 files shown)


crates/ibc/src/core/ics02_client/events.rs
```diff
@@ -390,3 +390,101 @@ impl From<UpgradeClient> for abci::Event {
...
+            expected_values: Vec<String>,
+        }
+
+        let client_type = mock_client_type();
+        let client_id = ClientId::new(client_type.clone(), 0).unwrap();
+        let consensus_height = Height::new(0, 10).unwrap();
+        let consensus_heights = vec![Height::new(0, 5).unwrap(), Height::new(0, 7).unwrap()];
+        let header: Any = MockHeader::new(consensus_height).into();
+        let expected_keys = vec![
+            CLIENT_ID_ATTRIBUTE_KEY,
```
plafer: Can we hardcode the value `"client_id"`, and a [URL to a the ibc-go value](https://github.com/cosmos/ibc-go/blob/887cba1658bbb2db46b96d1ae492cbe8095b1871/modules/core/03-connection/types/events.go#L10-L15) above the definition of `expected_keys`? This is to avoid the case that we change the value of `CLIENT_ID_ATTRIBUTE_KEY` for some reason; the test wouldn't catch that.

The general principle in these sorts of tests is you want to reuse as little as possible of the definitions that the implementation uses; otherwise you run the risk of changing the implementation incorrectly, and changing the test in the same way such that it doesn't fail when it should have.

crates/ibc/src/core/ics02_client/events.rs
```diff
@@ -390,3 +390,101 @@ impl From<UpgradeClient> for abci::Event {
...
+            client_id.to_string(),
+            MOCK_CLIENT_TYPE.to_string(),
+            consensus_height.to_string(),
+            consensus_heights
+                .iter()
+                .map(|h| h.to_string())
+                .collect::<Vec<_>>()
+                .join(","),
+            String::from_utf8(hex::encode(header.clone().value)).unwrap(),
+        ];
```
plafer: Similarly here, can we use the correct hardcoded values?

plafer: Same comment applies to the other 2 `events.rs` files

crates/ibc/src/core/ics04_channel/events/channel_attributes.rs
```diff
@@ -8,14 +8,13 @@ use crate::core::{
...
-/// to an `AbciEvent`
-pub const COUNTERPARTY_CHANNEL_ID_ATTRIBUTE_KEY: &str = "counterparty_channel_id";
-const COUNTERPARTY_PORT_ID_ATTRIBUTE_KEY: &str = "counterparty_port_id";
-const VERSION_ATTRIBUTE_KEY: &str = "version";
+pub(super) const CONNECTION_ID_ATTRIBUTE_KEY: &str = "connection_id";
+pub(super) const CHANNEL_ID_ATTRIBUTE_KEY: &str = "channel_id";
+pub(super) const PORT_ID_ATTRIBUTE_KEY: &str = "port_id";
+pub(super) const COUNTERPARTY_CHANNEL_ID_ATTRIBUTE_KEY: &str = "counterparty_channel_id";
+pub(super) const COUNTERPARTY_PORT_ID_ATTRIBUTE_KEY: &str = "counterparty_port_id";
+pub(super) const VERSION_ATTRIBUTE_KEY: &str = "version";
```
plafer: Then we won't need to expose these anymore

plafer (changes requested):

crates/ibc/src/core/ics02_client/events.rs
Farhad-Shabani: Well said 👍🏻 
Fixed https://github.com/cosmos/ibc-rs/pull/363/commits/ffa0bbcf972b93e074829390cb9cd3511eae23f9

crates/ibc/src/core/ics02_client/events.rs
Farhad-Shabani: https://github.com/cosmos/ibc-rs/pull/363/commits/ffa0bbcf972b93e074829390cb9cd3511eae23f9

crates/ibc/src/core/ics04_channel/events/channel_attributes.rs
Farhad-Shabani: https://github.com/cosmos/ibc-rs/pull/363/commits/ffa0bbcf972b93e074829390cb9cd3511eae23f9

crates/ibc/src/core/ics02_client/events.rs
```diff
@@ -390,3 +390,101 @@ impl From<UpgradeClient> for abci::Event {
...
+            client_id.to_string(),
+            "9999-mock".to_string(),
+            consensus_height.to_string(),
+            consensus_heights
+                .iter()
+                .map(|h| h.to_string())
+                .collect::<Vec<_>>()
+                .join(","),
+            String::from_utf8(hex::encode(header.clone().value)).unwrap(),
+        ];
```
plafer: This test doesn't give me the confidence that we're constructing the events in the exact same way as ibc-go. We're still using our own objects for the `expected_values`; we want hardcoded strings here too.

The best way to write this test is to 

1. run [ibc-go's `EmitCreateClientEvent()`](https://github.com/cosmos/ibc-go/blob/887cba1658bbb2db46b96d1ae492cbe8095b1871/modules/core/02-client/keeper/events.go#L18) with a given set of arguments
2. print the output events to the console
3. copy/paste the strings to construct `expected_values`
4. Recreate the same arguments in (1) for `client_type`, `client_id`, etc

Repeat for the connection/channel events tests.

Make sure to use the `main` branch for ibc-go; the `"consensus_heights"` attribute doesn't exist yet in 5.0.1, and I made the mistake of adding it to ibc-rs.

Farhad-Shabani: Should be fine now https://github.com/cosmos/ibc-rs/pull/363/commits/dc20ce72c523360cd693a2af000146f728cbbc5c
Just an issue I came across... Is it ok to have a default implementation for the `ClientId` ([here](https://github.com/cosmos/ibc-rs/blob/ef52941e29aa2759e2861cdae8a7398b114cc5f2/crates/ibc/src/core/ics24_host/identifier.rs#L241))? 
Therefore, the `core` module becomes chain-specific depending on a specific `client`. I think we should discard it.

plafer: Yes let's discard it; and we have more `Default` implementations like that that we should discard at the same time. Can you create a new issue? Should be low priority though

Farhad-Shabani: of course

crates/ibc/src/core/ics02_client/events.rs
```diff
@@ -390,3 +390,90 @@ impl From<UpgradeClient> for abci::Event {
...
+        let consensus_heights = vec![Height::new(0, 5).unwrap(), Height::new(0, 7).unwrap()];
+        let header: Any = MockHeader::new(consensus_height).into();
+        let expected_keys = vec![
+            "client_id",
+            "client_type",
+            "consensus_height",
+            "consensus_heights",
+            "header",
+        ];
+        let expected_values = vec!["07-tendermint-0", "07-tendermint", "0-5", "0-5,0-7"];
```
plafer: Oops, we're still missing the value for `header`

crates/ibc/src/core/ics04_channel/events.rs
```diff
@@ -1058,3 +1058,149 @@ impl TryFrom<TimeoutPacket> for abci::Event {
...
+            kind: IbcEventType,
+            event: AbciEvent,
+            expected_keys: Vec<&'static str>,
+            expected_values: Vec<&'static str>,
+        }
+
+        let port_id = PortId::default();
+        let channel_id = ChannelId::default();
+        let connection_id = ConnectionId::default();
+        let counterparty_port_id = PortId::default();
```
plafer: Let's not use `default()` here too; we can just use `new()`. I don't like the use of `default()` in tests as they make it harder to verify correctness quickly (I need to go check the implementation for `default()` for each of these), for little gain.

crates/ibc/src/core/ics02_client/events.rs
Farhad-Shabani: oh, weird. let me fix it

Farhad-Shabani: As discussed, this header's expected value is checked using `MockHeader`here. In order to verify it later against an emitted ibc-go event, a PR has been opened (https://github.com/cosmos/ibc-rs/issues/374)

plafer (approved):
🙏

Issue #279, opened 2022-12-02
Remove legacy `Reader`s and `Keeper`s


plafer:
In favor of `ValidationContext` and `ExecutionContext`. Each handler's `process()` should be implemented as suggested [here](https://github.com/cosmos/ibc-rs/issues/271#issuecomment-1331347210).

Follow-up to #257.

plafer:
This includes renaming the temporary `new_*` functions ([example](https://github.com/cosmos/ibc-rs/blob/649e7bd3a8cbd03183ed93ea42d35cec2779b8a8/crates/ibc/src/core/ics02_client/client_state.rs#L86)).

Issue #184, opened 2022-10-17
Refactor `HandlerOutput` and `HandlerResult`


plafer:
`HandlerOutput` and `HandlerResult` should be modeled roughly as:

```rust
struct HandlerResult<T> {
    // We want to return logs even if the processing fails
    logs: Vec<String>,
    result: Result<HandlerOutput<T>, Error>
}

struct HandlerOutput<T> {
    // previously `result`, which "collides" with Rust's `Result`
    value: T,
    events: Vec<IbcEvent>,
}
```

`HandlerOutputBuilder` would also be renamed to `HandlerResultBuilder`, and build a `HandlerResult`. `with_result()` could be split into `with_value(T)`, and `without_value()` (to avoid the ugly `with_result(())` when there's no result). Hopefully we could get rid of `merge()` and `merge_output()` as well.

@hu55a1n1 and @romac contributed to these ideas.

plafer:
Related: #23

plafer:
This API is no longer used since #279. It will be completely removed after #442.

PR #377, opened 2023-01-26
Implement ICS-26 router for `RecvPacket` in `{Validation,Execution}Context`

### LINKED ISSUE #322: Make `ValidationContext` and `ExecutionContext` implement the ICS-26 router

plafer:
Work towards: #322

Changed the scope of this PR to include only `recvPacket`, since it turned out to have many subtleties to get right.

## Description

As with the other PRs, this one focuses on correctness. The more we get closer to being done with #322, the more it becomes obvious that we need a refactor. Let's hold off until we're completely done though; it will be easier to make a big refactor after #279.

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/applications/transfer/acknowledgement.rs
+++ b/crates/ibc/src/applications/transfer/acknowledgement.rs
@@ -56,9 +56,10 @@ impl Display for TokenTransferAcknowledgement {
 
 impl From<TokenTransferAcknowledgement> for Vec<u8> {
     fn from(ack: TokenTransferAcknowledgement) -> Self {
+        // WARNING: Make sure all branches always return a non-empty vector.
+        // Otherwise, the conversion to `Acknowledgement` will panic.
         match ack {
             TokenTransferAcknowledgement::Success(_) => r#"{"result":"AQ=="}"#.as_bytes().into(),
-            // TokenTransferAcknowledgement::Error(s) => s.as_bytes(),
             TokenTransferAcknowledgement::Error(s) => alloc::format!(r#"{{"error":"{s}"}}"#).into(),
         }
     }
@@ -66,7 +67,10 @@ impl From<TokenTransferAcknowledgement> for Vec<u8> {
 
 impl From<TokenTransferAcknowledgement> for Acknowledgement {
     fn from(ack: TokenTransferAcknowledgement) -> Self {
-        ack.into()
+        let v: Vec<u8> = ack.into();
+
+        v.try_into()
+            .expect("token transfer internal error: ack is never supposed to be empty")
     }
 }
 
@@ -98,18 +102,23 @@ mod test {
         let ack_success: Vec<u8> = TokenTransferAcknowledgement::success().into();
 
         // Check that it's the same output as ibc-go
+        // Note: this also implicitly checks that the ack bytes are non-empty,
+        // which would make the conversion to `Acknowledgement` panic
         assert_eq!(ack_success, r#"{"result":"AQ=="}"#.as_bytes());
     }
 
     #[test]
     fn test_ack_error_to_vec() {
-        let ack_error = TokenTransferAcknowledgement::Error(
+        let ack_error: Vec<u8> = TokenTransferAcknowledgement::Error(
             "cannot unmarshal ICS-20 transfer packet data".to_string(),
-        );
+        )
+        .into();
 
         // Check that it's the same output as ibc-go
+        // Note: this also implicitly checks that the ack bytes are non-empty,
+        // which would make the conversion to `Acknowledgement` panic
         assert_eq!(
-            Vec::<u8>::from(ack_error),
+            ack_error,
             r#"{"error":"cannot unmarshal ICS-20 transfer packet data"}"#.as_bytes()
         );
     }
--- a/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
+++ b/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
@@ -60,3 +60,73 @@ pub fn process_recv_packet<Ctx: 'static + TokenTransferContext>(
 
     Ok(())
 }
+
+#[cfg(feature = "val_exec_ctx")]
+pub use val_exec_ctx::*;
+#[cfg(feature = "val_exec_ctx")]
+mod val_exec_ctx {
+    pub use super::*;
+    use crate::core::ics04_channel::handler::ModuleExtras;
+
+    ////////////////////////////////////////////////////////////
+    // FIXME BEFORE MERGE: events need to be emitted even when there's an error
+    ////////////////////////////////////////////////////////////
+    pub fn process_recv_packet_execute<Ctx: TokenTransferContext>(
+        ctx: &mut Ctx,
+        packet: &Packet,
+        data: PacketData,
+    ) -> Result<ModuleExtras, TokenTransferError> {
+        if !ctx.is_receive_enabled() {
+            return Err(TokenTransferError::ReceiveDisabled);
+        }
+
+        let receiver_account = data
+            .receiver
+            .clone()
+            .try_into()
+            .map_err(|_| TokenTransferError::ParseAccountFailure)?;
+
+        let extras = if is_receiver_chain_source(
+            packet.port_on_a.clone(),
+            packet.chan_on_a.clone(),
+            &data.token.denom,
+        ) {
+            // sender chain is not the source, unescrow tokens
+            let prefix = TracePrefix::new(packet.port_on_a.clone(), packet.chan_on_a.clone());
+            let coin = {
+                let mut c = data.token;
+                c.denom.remove_trace_prefix(&prefix);
+                c
+            };
+
+            let escrow_address =
+                ctx.get_channel_escrow_address(&packet.port_on_b, &packet.chan_on_b)?;
+
+            ctx.send_coins(&escrow_address, &receiver_account, &coin)?;
+
+            ModuleExtras::empty()
+        } else {
+            // sender chain is the source, mint vouchers
+            let prefix = TracePrefix::new(packet.port_on_b.clone(), packet.chan_on_b.clone());
+            let coin = {
+                let mut c = data.token;
+                c.denom.add_trace_prefix(prefix);
+                c
+            };
+
+            let denom_trace_event = DenomTraceEvent {
+                trace_hash: ctx.denom_hash_string(&coin.denom),
+                denom: coin.denom.clone(),
+            };
+
+            ctx.mint_coins(&receiver_account, &coin)?;
+
+            ModuleExtras {
+                events: vec![denom_trace_event.into()],
+                log: Vec::new(),
+            }
+        };
+
+        Ok(extras)
+    }
+}
--- a/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
+++ b/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
@@ -68,23 +68,21 @@ mod val_exec_ctx {
     pub use super::*;
     use crate::core::ics04_channel::handler::ModuleExtras;
 
-    ////////////////////////////////////////////////////////////
-    // FIXME BEFORE MERGE: events need to be emitted even when there's an error
-    ////////////////////////////////////////////////////////////
     pub fn process_recv_packet_execute<Ctx: TokenTransferContext>(
         ctx: &mut Ctx,
         packet: &Packet,
         data: PacketData,
-    ) -> Result<ModuleExtras, TokenTransferError> {
+    ) -> Result<ModuleExtras, (ModuleExtras, TokenTransferError)> {
         if !ctx.is_receive_enabled() {
-            return Err(TokenTransferError::ReceiveDisabled);
+            return Err((ModuleExtras::empty(), TokenTransferError::ReceiveDisabled));
         }
 
-        let receiver_account = data
-            .receiver
-            .clone()
-            .try_into()
-            .map_err(|_| TokenTransferError::ParseAccountFailure)?;
+        let receiver_account = data.receiver.clone().try_into().map_err(|_| {
+            (
+                ModuleExtras::empty(),
+                TokenTransferError::ParseAccountFailure,
+            )
+        })?;
 
         let extras = if is_receiver_chain_source(
             packet.port_on_a.clone(),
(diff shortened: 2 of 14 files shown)


crates/ibc/src/core/ics04_channel/handler/write_acknowledgement.rs
```diff
@@ -56,10 +56,6 @@ pub fn process<Ctx: ChannelReader>(
         Err(e) => return Err(e),
```
Farhad-Shabani, line 56: There is a slight deviation in the way we handle the `PacketAcknowledgementNotFound` error variant. We perceive it as an error, but eventually, it is accepted as a valid arm. I think `get_packet_acknowledgement(...)` should return `Result<Option<AcknowledgementCommitment>, PacketError>` type, so under normal circumstance, the `None` variant would show a valid absence of ACK, but in case where ever needed as an error, we can use another match or apply `ok_or()` (Though I didn't see any)

crates/ibc/src/mock/client_state.rs
```diff
@@ -355,6 +355,22 @@ impl ClientState for MockClientState {
         Ok(())
     }
 
+    #[cfg(feature = "val_exec_ctx")]
+    fn new_verify_packet_data(
```
Farhad-Shabani, line 359: So, this should be renamed to `verify_packet_data` eventually. Okay

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -175,6 +175,22 @@ pub trait ClientState:
         expected_client_state: Any,
     ) -> Result<(), ClientError>;
 
+    /// Verify a `proof` that a packet has been commited.
```
Farhad-Shabani: Typo from past: "committed"

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -175,6 +175,22 @@ pub trait ClientState:
...
+        connection_end: &ConnectionEnd,
+        proof: &CommitmentProofBytes,
+        root: &CommitmentRoot,
+        port_id: &PortId,
+        channel_id: &ChannelId,
+        sequence: Sequence,
+        commitment: PacketCommitment,
+    ) -> Result<(), ClientError>;
+
     /// Verify a `proof` that a packet has been commited.
```
Farhad-Shabani: Typo from past: "committed"

crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
```diff
@@ -60,3 +60,78 @@ pub fn process_recv_packet<Ctx: 'static + TokenTransferContext>(
...
+#[cfg(feature = "val_exec_ctx")]
+mod val_exec_ctx {
+    pub use super::*;
+    use crate::core::ics04_channel::handler::ModuleExtras;
+
+    pub fn process_recv_packet_execute<Ctx: TokenTransferContext>(
+        ctx: &mut Ctx,
+        packet: &Packet,
+        data: PacketData,
+    ) -> Result<ModuleExtras, (ModuleExtras, TokenTransferError)> {
```
Farhad-Shabani, line 75: Are we trying to move away from the output argument?

One thing that occurred to me is that grouping two orthogonal events and logs under the same struct (ModuleExtras) doesn't seem right.

- A logger should work till the last line of code is worked, but it's not true for events. Hence, the logger handler should be passed to functions somehow similar to the output arg.
- Contrary to logs, events can be returned or produced by functions, and this should be reflected in results.

I think this should be related to #184

crates/ibc/src/applications/transfer/acknowledgement.rs
```diff
@@ -56,17 +56,21 @@ impl Display for TokenTransferAcknowledgement {
 
 impl From<TokenTransferAcknowledgement> for Vec<u8> {
     fn from(ack: TokenTransferAcknowledgement) -> Self {
+        // WARNING: Make sure all branches always return a non-empty vector.
+        // Otherwise, the conversion to `Acknowledgement` will panic.
```
Farhad-Shabani, line 60: Would this panic, as it wasn't before?

Imho, this warning sounds like something couldn't be handled with the Rust error system, so it's up to users to be cautious about it. This also led to using `expect` in `From<TokenTransferAcknowledgement>` conversion.

We moved the `is_empty` check from the `write_acknowledgement::process` level to the higher `on_recv_packet` level. This makes me think we should change the return of `on_recv_packet` to `Result<Acknowledgment, PacketError>`. not sure ofc.

I did like removing that `is_empty()` check, but I have no idea how that could be handled better :)

crates/ibc/src/core/context.rs
```diff
@@ -1114,4 +1169,121 @@ mod val_exec_ctx {
...
+                    // the sequence number has already been incremented, so
+                    // another relayer already relayed the packet
+                    msg.packet.sequence < next_seq_recv
+                }
+            };
+
+            if packet_already_received {
+                return Ok(());
+            }
+        }
```
Farhad-Shabani, line 1219: This sounds more like a validation step.

Farhad-Shabani (commented):
All good :ok_hand: 
I just left some thoughts to get also your take on them

crates/ibc/src/core/ics04_channel/handler/write_acknowledgement.rs
plafer: Note that this is legacy code and is never called from `Validation/ExecutionContext`. We no longer treat `PacketAcknowledgementNotFound` specially in the [new implementation](https://github.com/cosmos/ibc-rs/blob/dc84bf2e8359f4bfcb78e3a8cbb9cad3a1c47002/crates/ibc/src/core/ics04_channel/handler/recv_packet.rs#L153-L161).

> I think get_packet_acknowledgement(...) should return Result<Option<AcknowledgementCommitment>, PacketError> type, so under normal circumstance, the None variant would show a valid absence of ACK

What's a "valid" absence of an ACK (as opposed to an "invalid one")? The (undocumented) contract of `get_packet_acknowledgement()` is "return an ack commitment, or tell me why you were not able to do it". The acknowledgement not being present is just another reason why the host would not be able to return it, and is an error in this context. IMO `Result<Option<AcknowledgementCommitment>, PacketError>` is harder to read, and doesn't provide us any gains that I can see. Note also that this would make the new implementation more complex, as we'd not only need to check whether or not there was an error, but also if the return was `None` in the success case.

crates/ibc/src/core/ics02_client/client_state.rs
plafer: [147ec47](https://github.com/cosmos/ibc-rs/pull/377/commits/147ec47a78503d30dfb26cdc1cba20766066b9c0)

crates/ibc/src/core/ics02_client/client_state.rs
plafer: [147ec47](https://github.com/cosmos/ibc-rs/pull/377/commits/147ec47a78503d30dfb26cdc1cba20766066b9c0)

crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
plafer: > Are we trying to move away from the output argument?

Yes.

> A logger should work till the last line of code is worked, but it's not true for events. Hence, the logger handler should be passed to functions somehow similar to the output arg.

Actually the events also need to be returned even in the case of an error (see [ibc-go](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/apps/transfer/ibc_module.go#L176-L210)).

This interface looks funny because I wanted to still use `?` for errors, but then `ModuleExtras` need to be returned also in the case of an error. `(ModuleExtras, Result<(), Error>)` would not have allowed the use of `?`.

There's probably a cleaner way to do it, but again, I prioritized correctness.

crates/ibc/src/applications/transfer/acknowledgement.rs
plafer: I'm also not a fan of needing that comment. Note also that I added a test to check that this conversion doesn't return a non-empty vector.

The fix would look like:
```rust
impl From<TokenTransferAcknowledgement> for NonEmptyVec<u8> {
    // ...
}

impl From<TokenTransferAcknowledgement> for Acknowledgement {
    fn from(ack: TokenTransferAcknowledgement) -> Self {
        let v: NonEmptyVec<u8> = ack.into();

        // this can be an `into()` because the only possible error case
        // (empty vector) is not possible with `NonEmptyVec<u8>`
        v.into()
    }
}
```

plafer: Created #389

crates/ibc/src/core/context.rs
plafer: It's not; notice that what we're trying to achieve here seeing if we should return early or not (or, in other words, don't do any writes). Does that make sense?

crates/ibc/src/applications/transfer/acknowledgement.rs
Farhad-Shabani: :pray:

crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
Farhad-Shabani: > Actually the events also need to be returned even in the case of an error (see ibc-go). This interface looks funny because I wanted to still use ? for errors, but then ModuleExtras need to be returned also in the case of an error. (ModuleExtras, Result<(), Error>) would not have allowed the use of ?.

Yea, I also tried that option locally, but it does not buy much. I see.

crates/ibc/src/core/context.rs
Farhad-Shabani: :thinking: I still think it is not of an early return type. A duplicate msg is not valid to process. A case can be considered: multiple relayers try to manipulate and attack by sending tons of messages, none of them should be allowed to pass the validation.

crates/ibc/src/core/ics04_channel/handler/write_acknowledgement.rs
Farhad-Shabani: > Note that this is legacy code and is never called from Validation/ExecutionContext. We no longer treat PacketAcknowledgementNotFound specially in the new implementation.

So, if it would not be the case, we shoule be ok then. thank you :pray:

crates/ibc/src/core/context.rs
plafer: > A duplicate msg is not valid to process.

Actually it is. With every packet to relay, relayers are in a race to relay the packet, and they'll all try to send the same `recvPacket` transaction at the same time. But only the first one will be accepted by the chain. Previously, ibc-go didn't have that "already received" check, and so all the relayers' transactions (except the first successful one) would error, which meant that other messages in the transaction would be discarded, and increased fees for the relayer (I think). But there's no point in penalizing relayers in this way, so ibc-go made this case a valid `recvPacket` message. See this ibc-go issue for more context: https://github.com/cosmos/ibc-go/issues/245.

> multiple relayers try to manipulate and attack by sending tons of messages

How is this an attack? What exactly would be manipulated? AFAIK, they're not gaining anything, because they spent money on the transaction, but the transaction ended up being a no-op and not changing any of the blockchain state.

Farhad-Shabani: >  Previously, ibc-go didn't have that "already received" check, and so all the relayers' transactions (except the first successful one) would error, 

I don't see it was considered as a valid msg. Instead of returning errors they return no-ops. 
Let's look at it from another side: which of the execution steps (state change, store, emit, log) are done on a redundant message? All of them are dropped as they are "not valid for execution"

(I am discussing this to clarify the validation and execution boundaries. All in all, It's not a big deal)

> How is this an attack? What exactly would be manipulated? AFAIK, they're not gaining anything, because they spent money on the transaction, but the transaction ended up being a no-op and not changing any of the blockchain state.

Something like DDoS attack, Getting the network down for a while by consuming the chain or relayers processing capacity to incur a hack by manipulating some other parts at that time.

plafer: > Let's look at it from another side: which of the execution steps (state change, store, emit, log) are done on a redundant message? All of them are dropped as they are "not valid for execution"

That's right; none are executed. Stated differently, sending a `recvPacket` msg for a packet that has already been received is a valid transaction which doesn't change the state. The reason for it to be a valid message is simply so that relayers don't have to pay extra fees for doing exactly what they're supposed to do. If we failed the validation, then they would pay extra fees; so instead we succeed the transaction, but don't change the state machine.

> Something like DDoS attack, Getting the network down for a while by consuming the chain or relayers processing capacity to incur a hack by manipulating some other parts at that time.

Can you expand on how that attack would work? How is this any different than someone spamming the chain with any other valid transaction (e.g. sending tokens back and forth between their own accounts)?

Farhad-Shabani: Per our discussion and giving it more thoughts, it makes sense :pray:

Farhad-Shabani (approved):

PR #383, opened 2023-01-27
ADR 006: Upgrade client implementation proposal

### LINKED ISSUE #362: ADR 006: Implement upgrade client functionality
## Summary

There is a requirement for an ADR that explains and justifies the logic behind of upgrade client process. Chains that host `ibc-rs` may bring different criteria from Tendermint’s, resulting in a different implementation from Tendermint's ibc-go

## Proposal
ADR coverage includes:

- Define the boundary between chain-specific operations and general processes
- List all the necessary validation steps
- Specify upgrade client implementation for the Tendermint light client (ics07_tendermint)
- Serve as a reference for those who want to implement this functionality.

## Related Issues
- #19 
- #141 
- #384 
- #385

Farhad-Shabani:
Closes: #362
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/docs/architecture/adr-006-upgrade-client-implementation.md
+++ b/docs/architecture/adr-006-upgrade-client-implementation.md
@@ -14,10 +14,10 @@ capability, chains that implement `IBC-rs` may bring various concerns and
 characteristics than Tendermint chains leading to different ways for upgrading
 their clients. However there are general rules that apply to all and can serve
 as a framework for any `IBC-rs` implementation. On this basis, this record aims
-to justify the logic behind upgrading a light client, determine the boundary
-between client-wide and client-specific upgrade processes, list requisites for
-validation and execution steps, and explain Tendermint's upgrade client
-implementation within the
+to justify the chain-wide logic behind upgrading light clients, list requisites
+for validation and execution steps, determine the boundary between basic and
+upgrade-specific validations by an IBC handler, and explain Tendermint's upgrade
+client implementation within the
 [ics07_tendermint](../../crates/ibc/src/clients/ics07_tendermint).
 
 ## Decision
@@ -46,18 +46,17 @@ implementation of this rationale in `IBC-rs` is explained.
 * It is **UP TO** the chain's architecture how updated client and consensus
   states are committed, either through decentralized approaches, like governance
   or centralized methods, like a multisig account, etc.
-* Upon commitment, chain's store **MUST** contain updated client and consensus
+* Upon commitment, chain's store **MUST** contain upgraded client and consensus
   states, which can be retrieved using respective upgrade paths
 
 #### IBC Handler accepts upgrade?
 
 After ensuring that the chain upgrades is supported by IBC, the general
 validation and execution steps that apply to all chains are as follows. The
-criteria for classifying a validation as client-wide (also can be known as
-basic) or client-specific was whether that IBC handler can perform that action
-just using its own context data, which are available upon calling interfaces
-provided by `ClientState` and `ConsensusState` traits, like `is_frozen()`,
-`latest_height()`, etc.
+criteria for classifying a validation as basic or upgrade-specific was whether
+that IBC handler can perform that check just using its own contextual data,
+which are available upon calling interfaces provided by `ClientState` and
+`ConsensusState` traits, like `is_frozen()`, `latest_height()`, etc.
 
 * Latest client state **MUST NOT** be frozen
 * Received update message including verification proofs **MUST** successfully be
@@ -66,6 +65,8 @@ provided by `ClientState` and `ConsensusState` traits, like `is_frozen()`,
   current revision which is somehow encoded in the proof verification process.
   This prevents premature upgrades, as the counterparty may cancel or modify
   upgrade plans before the last planned height.
+* Latest consensus state **MUST** be within the trusting period of the latest
+  client state, which for clients without a trusting period is not applicable.
 
 Any other requisite beyond the above rules are considered client-specific. Next,
 we go through the upgrade process for Tendermint clients to justify the logic
@@ -121,7 +122,9 @@ supported by `IBC-rs`:
 An IBC-connected Tendermint chain will take the following steps to completely
 upgrade its own chain and counterparty's IBC client. Note that the chain-level
 upgrade instruction (1) is not a part of the IBC protocol. It is provided for
-the sake of the big picture and as an example of how the upgrade process can be.
+the sake of the big picture and as a reference to follow the upgrade process
+from the very beginning when a proposal is initiated to when the upgrade message
+is entirely handled.
 
 1. Upgrade chain through governance
    1. Create a 02-client
@@ -147,17 +150,21 @@ the sake of the big picture and as an example of how the upgrade process can be.
         upgrade/UpgradedIBCState/{upgradeHeight}/upgradedConsState
         ```
 
+      Notice that since the `UpgradedConsensusState` will not be available at
+      the upgrade path prior to this height, relayers cannot submit a valid
+      upgrade message as the proof verification would fail
+
 2. Submit an upgrade client message by relayers to the counterparty chain
    1. Wait for the upgrading chain to reach the upgrade height and halt
    2. Query a full node for the proofs of `UpgradedClient` and
       `UpgradedConsensusState` at the last height of the old chain
    3. Update the counterparty client to the last height of the old chain using
       the `UpdateClient` msg
-   4. Submit an `UpgradeClient` msg to the counterparty chain with the
+   4. Submit a `MsgUpgradeClient` message to the counterparty chain with the
       `UpgradedClient`, `UpgradedConsensusState` and their respective proofs
 
 3. Process the upgrade message on the counterparty chain upon receiving a
-`MsgUpgradeClient` message performs basic validations (BV), client-specific
+`MsgUpgradeClient` message performs basic validations (BV), upgrade-specific
 validations (SV) and lastly execution (E) steps as follows:
    1. (BV) Check that the current client is not frozen
    2. (BV) Check if the latest consensus state is within the trust period
--- a/docs/architecture/adr-006-upgrade-client-implementation.md
+++ b/docs/architecture/adr-006-upgrade-client-implementation.md
@@ -170,6 +170,7 @@ client-specific validations (SV) and lastly execution (E) steps as follows:
       such as ChainID, UnbondingPeriod, and ProofSpecs with the committed client
    8. (SV) Verify that the upgrading chain did indeed commit to the upgraded
       client state at the upgrade height by provided proof
+      Note: client-customizable fields must be zeroed out for this check
    9. (SV) Verify that the upgrading chain did indeed commit to the upgraded
       consensus state at the upgrade height by provided proof
    10. (E) Upgrade client to the new client by retaining old client-customizable
@@ -222,6 +223,12 @@ with that one of the decisions made is to split off the
       upgraded_consensus_state: Any,
    ) -> Result<UpdatedState, ClientError>;
    ```
+
+There is also a need to move away from `upgrade` method of `ClientState` trait
+leaving only zeroed out fields through `zero_custom_fields()`. New state
+commitment should be done in `update_state_with_upgrade_client` and stored via
+`store_client_result` of keeper context, so the upgrade part is no longer
+necessary.
 
 Listed below are the code snippets that correspond to the third step of the
 previous section as mentioned:
@@ -315,6 +322,7 @@ previous section as mentioned:
       key_path: client_upgrade_path,
    };
 
+   upgraded_tm_client_state.zero_custom_fields();
    let client_state_value =
       Protobuf::<RawTmClientState>::encode_vec(&upgraded_tm_client_state)
          .map_err(ClientError::Encode)?;
@@ -384,6 +392,8 @@ previous section as mentioned:
        upgraded_tm_cons_state.next_validators_hash,
     );
     ```
+
+```rust
 
 ## Status
 
--- a/docs/architecture/adr-006-upgrade-client-implementation.md
+++ b/docs/architecture/adr-006-upgrade-client-implementation.md
@@ -170,7 +170,7 @@ validations (SV) and lastly execution (E) steps as follows:
    7. (SV) Match any Tendermint chain specified parameter in upgraded client
       such as ChainID, UnbondingPeriod, and ProofSpecs with the committed client
    8. (SV) Verify that the upgrading chain did indeed commit to the upgraded
-      client state at the upgrade height by provided proof Note:
+      client state at the upgrade height by provided proof. Note that the
       client-customizable fields must be zeroed out for this check
    9. (SV) Verify that the upgrading chain did indeed commit to the upgraded
       consensus state at the upgrade height by provided proof
@@ -178,16 +178,18 @@ validations (SV) and lastly execution (E) steps as follows:
       parameters (sent by relayers) such `TrustingPeriod`, `TrustLevel`,
       `MaxClockDrift` and adopt the new chain-specified fields such as
       `UnbondingPeriod`, `ChainId`, `UpgradePath`, etc.
-   11. (E) Upgrade consensus state with a stand-in sentinel value for root Note:
-      The upgraded consensus state serve purely as a basis of trust for future
-      `UpdateClientMsgs`, and therefore does not require a root for proof
-      verification. Also, we do not set processed time for this consensus state
-      since this consensus state should not be used for packet verification as
-      the root is empty. To ensure the connection can be used for relaying
-      packets, relayers must submit an `UpdateClientMsg` with a header from the
-      new chain.
+   11. (E) Upgrade consensus state with a stand-in sentinel value for root. Note
+      that the upgraded consensus state serves purely as a basis of trust for
+      future `UpdateClientMsgs`, and therefore it does not require a root for
+      proof verification and it is not used for packet verifications as well.
+      That sentinel value serves as a temporary substitute until the root of new
+      chain gets available by `UpdateClientMsg`. It is set by module with a
+      distinct, easily recognizable value to reduce the risk of bugs. Thereby,
+      we do not also set a processed time for this consensus state. To ensure
+      the connection can be used for relaying packets, relayers must submit an
+      `UpdateClientMsg` with a header from the new chain.
 
-1. Submit an `UpdateClient` msg by a relayer to the counterparty chain with a
+4. Submit an `UpdateClient` msg by a relayer to the counterparty chain with a
    header from the newly upgraded chain
 
 #### Decisions
--- a/docs/architecture/adr-006-upgrade-client-implementation.md
+++ b/docs/architecture/adr-006-upgrade-client-implementation.md
@@ -6,16 +6,16 @@
 
 ## Context
 
-The ability to upgrade is crucial for IBC-connected chains. It allows them to
-evolve and improve without any restrictions. The IBC module may not be affected
-by some upgrades, but some may require it to be upgraded as well to maintain
-high-value connections to other chains secure. For having this capability,
-chains that implement `IBC-rs` may bring various concerns and characteristics
-than Tendermint chains leading to different ways for upgrading their clients.
-However there are generic rules that apply to all and can serve as a framework
-for any `IBC-rs` implementation. On this basis, this record aims to justify the
-logic behind upgrading a light client, determine the boundary between
-client-wide and client-specific upgrade processes, list requisites for
+The ability to upgrade is a crucial feature for IBC-connected chains, as it
+enables them to evolve and improve without limitations. The IBC module may
+not be affected by some upgrades, but some may require it to be upgraded as well
+to maintain high-value connections to other chains secure. For having this
+capability, chains that implement `IBC-rs` may bring various concerns and
+characteristics than Tendermint chains leading to different ways for upgrading
+their clients. However there are general rules that apply to all and can serve
+as a framework for any `IBC-rs` implementation. On this basis, this record aims
+to justify the logic behind upgrading a light client, determine the boundary
+between client-wide and client-specific upgrade processes, list requisites for
 validation and execution steps, and explain Tendermint's upgrade client
 implementation within the
 [ics07_tendermint](../../crates/ibc/src/clients/ics07_tendermint).
@@ -38,10 +38,6 @@ implementation of this rationale in `IBC-rs` is explained.
   breaks the counterparty IBC client.
 * There **MUST** be a proof verification process to check upgraded client and
   consensus states against the host chain's state.
-* Clients **MUST** ensure that the planned last height of the current revision
-  is somehow encoded in the proof verification process. This prevents premature
-  upgrades, as the counterparty may cancel or modify upgrade plans before the
-  last planned height.
 * Chain upgrades **MUST NOT** result in changing
   [ClientState](../../crates/ibc/src/core/ics02_client/client_state.rs#ClientState)
   or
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+validation and execution steps that apply to all chains are as follows. The
+criteria for classifying a validation as client-wide (also can be known as
+basic) or client-specific was whether that IBC handler can perform that action
+just using its own context data, which are available upon calling interfaces
+provided by `ClientState` and `ConsensusState` traits, like `is_frozen()`,
+`latest_height()`, etc.
+
+* Latest client state **MUST NOT** be frozen
+* Latest consensus state **MUST** be within the trusting period of the latest
+  client state
```
plafer: Note that currently our ICS-2 has methods about the trusting period, but I think that's a mistake as it's actually a Tendermint concept?

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+* Received update message including verification proofs **MUST** successfully be
+  decoded into the domain types
+* Clients **MUST** ensure that the planned last height of the current revision
+  is somehow encoded in the proof verification process. This prevents premature
+  upgrades, as the counterparty may cancel or modify upgrade plans before the
+  last planned height.
```
plafer: The sdk today only writes to the store when it's *at* the planned height; never before.

> Clients **MUST** ensure that the planned last height of the current revision
  is somehow encoded in the proof verification process.

What's the source of this statement? If I send the planned upgrade for future height `h` in advance to the client, I'll still need to send a header update to the client when the upgrading chain reaches height `h`, and only then the client would be allowed to upgrade. So then why even send the information in advance? And what if the planned upgrade is canceled; what's the message type that tells the client that the upgrade is no longer happening?

Can we expand on this?

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+
+#### Acceptable Client State Upgrades
+
+Below list enumerates upgrades on Tendermint chains that would break
+counterparty IBC Tendermint clients and indicates whether or not the change is
+supported by `IBC-rs`:
+
+* S: Supported, P: Partially Supported, U: Unsupported
+
+1. (S) Changing the `ChainId`
```
plafer, line 88: For readability, could you order them by (S), (P) and (U)?

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+   negotiated on connection handshake.
+9. (U) Changing parameters that are customizable by relayers like `TrustLevel`
+    and `TrustingPeriod`, `max_clock_drift`
+10. (P) Changing the Tendermint LightClient algorithm: Changes to the light
+   client algorithm that do not change the
+   [ClientState](../../crates/ibc/src/clients/ics07_tendermint/client_state.rs#ClientState)
+   or
+   [ConsensusState](../../crates/ibc/src/clients/ics07_tendermint/consensus_state.rs#ConsensusState)
+   struct may be supported, provided that the counterparty is also upgraded to
+   support the new light client algorithm. Changes that require updating the
```
plafer: > provided that the counterparty is also upgraded to support the new light client algorithm

Is this actually included in the definition of what qualifies as a "client upgrade"? I was under the impression that this wouldn't be included since the client upgrade can't be done by simply processing a `MsgClientUpgrade`.

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+        ```md
+        upgrade/UpgradedIBCState/{upgradeHeight}/upgradedClient
+        ```
+
+   4. Commit an initial consensus state by upgrade module on the block right
+      before the upgrade height for the next chain under the following key:
+
+        ```md
+        upgrade/UpgradedIBCState/{upgradeHeight}/upgradedConsState
+        ```
```
plafer, line 151: Most of this section doesn't concern us; a lot of what's described here is how the sdk chose to do it, but is not fundamental to IBC. All we care about from the client's perspective is the path at which the upgraded `ClientState` and `ConsensusState` are placed at, and how to interpret the data structures in this context. 

> Commit an initial consensus state by upgrade module on the block right before the upgrade height for the next chain under the following key

This confirms my earlier comment: it seems to be a requirement for the chain to write the upgraded `ConsensusState` (and `ClientState`?) to the store only when it is ready to upgrade. The mechanism that the chain uses to get there (e.g. governance) is irrelevant from IBC's perspective.

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+   1. Wait for the upgrading chain to reach the upgrade height and halt
+   2. Query a full node for the proofs of `UpgradedClient` and
+      `UpgradedConsensusState` at the last height of the old chain
+   3. Update the counterparty client to the last height of the old chain using
+      the `UpdateClient` msg
+   4. Submit an `UpgradeClient` msg to the counterparty chain with the
+      `UpgradedClient`, `UpgradedConsensusState` and their respective proofs
+
+3. Process the upgrade message on the counterparty chain</br> IBC handler upon
+receiving a `MsgUpgradeClient` message performs basic validations (BV),
```
plafer:
```suggestion
3. Process the upgrade message on the counterparty chain
Upon receiving a `MsgUpgradeClient` message, the IBC handler performs basic validations (BV),
```

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+client-specific validations (SV) and lastly execution (E) steps as follows:
+   1. (BV) Check that the current client is not frozen
+   2. (BV) Check if the latest consensus state is within the trust period
+   3. (BV) Check if the message containing proofs decoded successfully
+   4. (SV) Verify that the upgradedClient be of a Tendermint `ClientState` type
+   5. (SV) Verify that the upgradedConsensusState be of a Tendermint
+      `ConsensusState` type
+   6. (SV) Check the latest height of the current client state does not have the
+      same revision number or has a greater height than the committed client
```
plafer: The requirement [seems to be](https://github.com/cosmos/ibc-go/blob/15c090809046e165eb2c4be5fd27b93bc8cb4bf4/modules/light-clients/07-tendermint/types/upgrade.go#L39-L42) that the upgraded client can't have a height greater than the current client, right?

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+   7. (SV) Match any Tendermint chain specified parameter in upgraded client
+      such as ChainID, UnbondingPeriod, and ProofSpecs with the committed client
```
plafer, line 178: Can you point to me where this is in ibc-go? I don't see it anywhere

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+   8. (SV) Verify that the upgrading chain did indeed commit to the upgraded
+      client state at the upgrade height by provided proof
+      Note: client-customizable fields must be zeroed out for this check
+   9. (SV) Verify that the upgrading chain did indeed commit to the upgraded
+      consensus state at the upgrade height by provided proof
+   10. (E) Upgrade client to the new client by retaining old client-customizable
+      parameters (sent by relayers) such `TrustingPeriod`, `TrustLevel`,
+      `MaxClockDrift` and adopt the new chain-specified fields such as
+      `UnbondingPeriod`, `ChainId`, `UpgradePath`, etc.
+   11. (E) Upgrade consensus state with a stand-in sentinel value for root</b>
```
plafer:
```suggestion
   11. (E) Upgrade consensus state with a stand-in sentinel value for root
```

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+      Note: The upgraded consensus state serve purely as a basis of trust for
+      future `UpdateClientMsgs`, and therefore does not require a root for proof
+      verification. Also, we do not set processed time for this consensus state
```
plafer: Can you expand on this? What's a "stand-in sentinel value"? How does it serve as a basis of trust for future `UpdateClient` msgs? That is, how is the next `UpdateClient` message going to use the value?

docs/architecture/adr-006-upgrade-client-implementation.md
```diff
@@ -0,0 +1,433 @@
...
+
+7. ```rust
+   // Ensure that the new unbonding period is not shorter than the current's client unbonding period
+   if self.unbonding_period > upgraded_tm_client_state.unbonding_period {
+      return Err(ClientError::ClientSpecific {
+            description:
+               "cannot upgrade client with a shorter unbonding period than the current one"
+                  .to_string(),
+      });
+   }
```
plafer, line 313: I don't see this in ibc-go. I'd be interested to know why they don't have that check, which would inform whether or not we should keep it.

Same goes for any other check that ibc-go doesn't have.

plafer (commented):
Great work 🎉👌! I'm now convinced that we can have an upgrade handler that's compatible with ibc-go (modulo a few questions I left in this review).

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Not necessarily. I think it's more of the protocol bound and looking at it from the security perspective, having that method enforces IBC implementors to consider it at their design stage. Isn't it?

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Check this [line](https://github.com/cosmos/ibc-go/blob/e650be91614ced7be687c30eb42714787a3bbc59/modules/core/exported/client.go#L117). We wont't send the message in advance. Actually, The purpose of this statement was to put an emphasize on "there must be a proof verification for the upgrade process" which occurs at the planned last height. Maybe I should reword it.

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: It is concerned that the upgrade would change the abstraction of those structs. It shoudn't be so, but is free to change the `ClientState` or `ConsensusState` instances/fields

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: > Most of this section doesn't concern us; a lot of what's described here is how the sdk chose to do it, but is not fundamental to IBC. All we care about from the client's perspective is the path at which the upgraded ClientState and ConsensusState are placed at, and how to interpret the data structures in this context.

Actually, I put this here for the sake of big picture and as a reference for someone who wants to follow all the steps from the very beginning when an upgrade proposal is initiated to when the upgrade message is entirely handled. But I can discard this.

> This confirms my earlier comment: it seems to be a requirement for the chain to write the upgraded ConsensusState (and ClientState?) to the store only when it is ready to upgrade. The mechanism that the chain uses to get there (e.g. governance) is irrelevant from IBC's perspective.

That's right. Honestly, there are still some edges that I haven't fully absorbed, which won't be possible until I perform some integration tests

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Yes!

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Check [this](https://github.com/cosmos/ibc-go/blob/c86d27fc280cfb342a9e4689b381e5823441b694/modules/light-clients/07-tendermint/upgrade.go#L25) out. It was the place I wanna use that mentioned `validate()` method in [#384](https://github.com/cosmos/ibc-rs/issues/384)

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: It was ordered same as the client state fields 
Done: [1b4dd31](https://github.com/cosmos/ibc-rs/pull/383/commits/1b4dd3115e36682e4486d0f18ca3f3f2bc92e2fd)

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: [1b4dd31](https://github.com/cosmos/ibc-rs/pull/383/commits/1b4dd3115e36682e4486d0f18ca3f3f2bc92e2fd)

Farhad-Shabani:
Thank you for the comments :pray:
I'll retouch this proposal after investigating some more corners plus the points you mentioned.

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: I looked up [ICS-2](https://github.com/cosmos/ibc/blob/main/spec/core/ics-002-client-semantics/README.md#updatestate), and it mentions:

> Clients MAY have time-sensitive validity predicates, such that if no ClientMessage is provided for a period of time (e.g. an unbonding period of three weeks) it will no longer be possible to update the client, i.e., the client is being frozen. In this case, a permissioned entity such as a chain governance system or trusted multi-signature MAY be allowed to intervene to unfreeze a frozen client & provide a new correct ClientMessage.

So the trusting period is optional in general. Therefore I think it would be best to have this as a client-specific check? That is, protocols with a trusting period will have that check, but not all protocols have a trusting period.

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: I feel like the comment might be outdated? The only height checks I see are [here](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L37-L42), and neither is a check about the "planned height". Or maybe I'm misunderstanding what a "planned last height" is?

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: > Actually, I put this here for the sake of big picture and as a reference for someone who wants to follow all the steps from the very beginning when an upgrade proposal is initiated to when the upgrade message is entirely handled.

That's a good idea. But let's make it clear that this is *an example implementation* rather than what's mandated by IBC. Perhaps a markdown "block quote", or whatever makes it visually clear too that we're discussing an example

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: Can we just write that then, as it's equivalent but more readable?

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: What I'm asking is more: is this a "client upgrade"? The upgrade algorithms are very different:

1. process a `UpgradeClient` message, vs
2. perform a chain hard fork

My understanding was that a client upgrade was only about (1); (2) would be something else

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: > Check [this](https://github.com/cosmos/ibc-go/blob/c86d27fc280cfb342a9e4689b381e5823441b694/modules/light-clients/07-tendermint/upgrade.go#L25) out.

Sorry, I meant the code that actually checks that rather than the "written statement". The only one I was able to find was [the `chainID` check](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/client_state.go#L129).

> It was the place I wanna use that mentioned validate() method in #384 

Notice that `validate()` [is called](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L101-L109) right after the constructor, so I don't see the gain in separating the validation from the constructor. And I still believe it is a code smell to allow an object to be constructed in an invalid state.

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Makes sense. Thanks for clearing this out 
[d59ddfe](https://github.com/cosmos/ibc-rs/pull/383/commits/d59ddfe9c718ded5121c479e562338b9ebcc2c36)

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: If the upgrade process is planned to happen at height H, the counterparty client won’t accept a submitted upgrade message before height H - 1 (This is the last planned height at the current(old) version of chain). This is checked via proofs, as if they are not for the correct height (H - 1) the proof verifications would fail.
[d59ddfe](https://github.com/cosmos/ibc-rs/pull/383/commits/d59ddfe9c718ded5121c479e562338b9ebcc2c36)

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Sure [d59ddfe](https://github.com/cosmos/ibc-rs/pull/383/commits/d59ddfe9c718ded5121c479e562338b9ebcc2c36)

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: >Sorry, I meant the code that actually checks that rather than the "written statement". The only one I was able to find was [the chainID check](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/client_state.go#L129).

Right, It is stated but the only check, on `ibc-go`, is done for the `ChainID` plus the validation of the constructed fields of the upgraded client state....
I added a check for `unboundingPeriod`, which does not exist on `ibc-go` (as you also pointed out), but for others I think it could be restrictive if something be placed. Though, I am still investigating.

>Notice that validate() [is called](https://github.com/cosmos/ibc-go/blob/c0acd5bd1778f2b7ecdf593006f56bd3e273bd49/modules/light-clients/07-tendermint/types/upgrade.go#L101-L109) right after the constructor, so I don't see the gain in separating the validation from the constructor. And I still believe it is a code smell to allow an object to be constructed in an invalid state.

Actually, the main concern has been elaborated [here](https://github.com/cosmos/ibc-rs/issues/384#issuecomment-1407034388), but maybe we should leave it as is.

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: [92e0334](https://github.com/cosmos/ibc-rs/pull/383/commits/92e0334803b5e0363a1c5274b9194a3096e36f3a)

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: > This is checked via proofs, as if they are not for the correct height (H - 1) the proof verifications would fail.

I was going to ask you to point to me the line of code that does that. But I think I just understood what they meant by "the planned last height of the current revision is somehow encoded in the proof verification process". That didn't make sense to me, because all proofs in IBC are Merkle proofs, which doesn't check anything other than "this item is/isn't in this Merkle tree". 

So then how do you "encode the last height in the proof verification"? I think how they implement that is to only write to the store *at* the planned height, as part of the protocol. Therefore, if the Merkle proof verification passes, the added semantic is "we are currently at the last planned height".

So this is why I couldn't find any line of code that did that; it is implied by the protocol. So if I'm right here, it means that it is then a requirement for Tendermint *hosts* to only write the upgraded state *at the planned height* in order to be compatible with ibc-go's light client implementation.

@AdityaSripal can you confirm that this is correct?

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: Let's move the `validate()` conversation to #384

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Exactly, My understanding was that.  It's implied by the protocol.

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: [d59ddfe](https://github.com/cosmos/ibc-rs/pull/383/commits/d59ddfe9c718ded5121c479e562338b9ebcc2c36)

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: After thinking about it some more, I get it, and it's actually quite simple 😅. I would call it a "dummy value", as it will never be used for packet verification or anything else. Basically the only fields that matter is the validator hash ones, and the one needed for checking the trusting period. I guess it just makes intuitive sense for me now why putting a dummy value as the root will only affect Merkle proofs, and not the ability to accept a new valid header with a `MsgUpdateClient`. All good!

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Another thing we should note is:
The initial consensus state for the next chain is only committed at height H - 1 ([Ref](https://github.com/cosmos/ibc-go/blob/main/docs/ibc/upgrades/quick-guide.md#step-by-step-upgrade-process-for-sdk-chains)). So, there won't be any `UpgradedConsensusState` available at the upgrade path before that height and if a relayer tries to submit an upgrade message prior to that, it should fail.
I added a short description in this regard: [ec94b51](https://github.com/cosmos/ibc-rs/pull/383/commits/ec94b512018aec6c5b8d5947426ccf781734673f)

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: I revisited this part. The issue was with the way I classified validations. 
From the implementation standpoint, it does make more sense to categorize validations into two **basic** vs **upgrade-specific** groups. I reverted the previous change: https://github.com/cosmos/ibc-rs/pull/383/commits/ec94b512018aec6c5b8d5947426ccf781734673f

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: You made an accurate point. But, If it would cause a hard fork, it shouldn't be included in the list of possible supported upgrades by IBC-Go. 
Anyway, it is beyond my knowledge at the moment. We should either keep it for the sake of compliance with the IBC-go, but investigate it later?! or discard it and play it by ear if any request comes from our clients. WDYT?

docs/architecture/adr-006-upgrade-client-implementation.md
Farhad-Shabani: Per our discussion, we'll keep this check, unless there's a good reason or serious need for shorter unbounding periods

docs/architecture/adr-006-upgrade-client-implementation.md
plafer: Let's keep it for now 👍

plafer (approved):
🙏

PR #386, opened 2023-01-27
Organize host chain methods under a new `HostChainContext` trait 

### LINKED ISSUE #384: Extract validation step inside new client state creation and expose it as a separate method
## Improvement Summary

Separate validation from new client state creation method ([here](https://github.com/cosmos/ibc-rs/blob/ef52941e29aa2759e2861cdae8a7398b114cc5f2/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L83))

## Proposal

- Client states received via an IBC message should be verifiable without having to create a new one. There should be a separate `validate()` method exposed to encompass all the fields' validation checks.
- The inclusion of validation steps inside new() may also lead to implementors forgetting to perform validations
- One use case is for upgrading client process.

Farhad-Shabani:
Closes: #384 , #418

---
### Description

Following the discussion around client state validation methods that check parameters (fields) of the client state implementation. After a bit of thought, I came up with an idea (around `validate_self_client`) of combining all host-related methods under a single chain called "HostChainContext" and separating related host errors under the "HostError" enum, plus added some docstrings.
These methods do not have chain-specific arguments and returns, so ideally, any chain should be able to provide them.
There just remains one small cavity. Having the "unbonding_period" method in it, which looks more of Tendermint chains. Couldn't find a better way at the moment, and I addressed it by changing the return type into the `Option<Duration>`

**Note**: The change in this PR is also connected to #418. That should have been handled here, hopefully.
_____

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


plafer:
> I came up with an idea (around validate_self_client) of combining all host-related methods under a single chain called "HostChainContext"

`ValidationContext` and `ExecutionContext` are *all* the host-related methods; in the previous API, `Client/Connection/Channel{Keeper,Reader}` contained the host methods. What criterion did you use to determine whether a method should be moved to the `HostChainContext`? For example, isn't `consensus_state()` a host method too? Or `connection_end()`? Or any other method?

> These methods do not have chain-specific arguments and returns, so ideally, any chain should be able to provide them.

What do you mean by "chain-specific arguments and returns"? As designed, `ValidationContext` & `ExecutionContext` are implementable by *any* chain. To better understand where you're coming from, could you give an example of a method in `ValidationContext/ExecutionContext` that a hypothetical chain would not be able to implement?

crates/ibc/src/clients/ics07_tendermint/host_helpers.rs
```diff
@@ -1,131 +1,122 @@
...
-        let counterparty_client_state = TmClientState::try_from(counterparty_client_state)
-            .map_err(|_| ConnectionError::InvalidClientState {
+/// Provides an implementation of `validate_self_client` for Tendermint-based
+/// hosts.
+pub fn validate_self_tendermint_client<Ctx>(
+    ctx: &Ctx,
+    counterparty_client_state: Any,
+) -> Result<(), HostError>
+where
+    Ctx: HostChainContext,
```
plafer, line 19: `ValidateSelfClientContext` is designed to ask for an implementation of all the methods that it needs in its implementation; some only implementable by Tendermint hosts (such as `unbonding_period()`). I believe this is the root cause for why you had to change the signature to `Option<Duration>`; not all hosts can implement it.

This is why I don't think it is appropriate for `ValidationContext` to have `HostChainContext` as a supertrait. Hypothetically, say we also provided the implementation of `validate_self_client()` for NEAR hosts; would we need to have that `Context` be a supertrait of `ValidationContext` as well, requiring all hosts to implement NEAR-specific methods (and have the signature return an `Option` for those too)?

crates/ibc/src/core/ics04_channel/context.rs
```diff
@@ -127,10 +127,9 @@ pub trait ChannelReader {
 
     /// Returns the current timestamp of the local chain.
     fn host_timestamp(&self) -> Result<Timestamp, ChannelError> {
-        let pending_consensus_state = self
-            .pending_host_consensus_state()
-            .expect("host must have pending consensus state");
-        Ok(pending_consensus_state.timestamp())
+        let host_height = self.host_height()?;
+        let host_consensus_state = self.host_consensus_state(&host_height)?;
+        Ok(host_consensus_state.timestamp())
```
plafer, line 132: I don't think we can assume that this will work for all hosts. Basecoin-rs is an example: we added `pending_host_consensus_state()` because `host_consensus_state()` was not quite right. I think it should [simply be made abstract](https://github.com/cosmos/ibc-rs/issues/418#issuecomment-1426200776); each host can implement it however they see fit. We should do it in a separate PR though to keep PRs as self-contained as possible.

Farhad-Shabani:
> `ValidationContext` and `ExecutionContext` are _all_ the host-related methods; in the previous API, `Client/Connection/Channel{Keeper,Reader}` contained the host methods. What criterion did you use to determine whether a method should be moved to the `HostChainContext`? For example, isn't `consensus_state()` a host method too? Or `connection_end()`? Or any other method?

For a clearer picture of where it's instantiated, consider the below grouping:

- Group 1: Methods that retrieve state from the consensus engine (or we shall say from non-module parts) and validate a state against them
- Group 2: Methods that call the value of a state from paths (module-related parts) and validate them
- Group 3: Methods that set a new state by storing in the paths (module-related parts)

crates/ibc/src/clients/ics07_tendermint/host_helpers.rs
Farhad-Shabani: I understand your concern, but rather than the `unbonding_period()` method, which I also found it there to be a wart, other methods should be available by any chain

crates/ibc/src/core/ics04_channel/context.rs
Farhad-Shabani: Actually, moving the `host_timestamp()` methods brought a need to handle them here. but, sure alright.

Farhad-Shabani:
The idea had the aim of demonstrating how it could make implementation smoother, but looking at your comment, if it's not going to make it easier for us, then it wouldn't work for that purpose.

crates/ibc/src/core/ics04_channel/context.rs
plafer: what do you mean by "moving the `host_timestamp()` methods brought a need to handle them here"?

plafer:
Closing as discussed in sync meeting.

Issue #246, opened 2022-11-16
calls to `store_update_{time,height}` should not be done in client handlers


plafer:
Should be called in tendermint's implementation of `ClientState.initialise()`, as in [ibc-go](https://github.com/cosmos/ibc-go/blob/878986c1dfa3ace17eaba82729265a0b2d1934bc/modules/light-clients/07-tendermint/types/client_state.go#L189). We currently call it [in the `CreateClient` handler](https://github.com/cosmos/ibc-rs/blob/2bce00c3c96693cc4d05c7fc111d8f59ceadd649/crates/ibc/src/core/ics02_client/context.rs#L88-L97) and the [`UpdateClient` handler](https://github.com/cosmos/ibc-rs/blob/2bce00c3c96693cc4d05c7fc111d8f59ceadd649/crates/ibc/src/core/ics02_client/context.rs#L107-L116).

plafer:
Also, the "processed time" is the block time of the last header. Instead of storing it as a separate value, couldn't we instead retrieve it from the last stored `ConsensusState` (which might require extra bookkeeping of its own, such as the height of the last update)?

PR #393, opened 2023-02-01
Finish implementation of `Validation/ExecutionContext`

### LINKED ISSUE #322: Make `ValidationContext` and `ExecutionContext` implement the ICS-26 router

plafer:
Closes: #322 

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -1150,6 +1150,38 @@ impl Ics2ClientState for ClientState {
         )
     }
 
+    #[cfg(feature = "val_exec_ctx")]
+    fn new_verify_packet_acknowledgement(
```
Farhad-Shabani, line 1154: Good to cover these proof verifications with tests later ([#398](https://github.com/cosmos/ibc-rs/issues/398))

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -1150,6 +1150,38 @@ impl Ics2ClientState for ClientState {
...
+        proof: &CommitmentProofBytes,
+        root: &CommitmentRoot,
+        port_id: &PortId,
+        channel_id: &ChannelId,
+        sequence: Sequence,
+        ack: AcknowledgementCommitment,
+    ) -> Result<(), ClientError> {
+        let client_state = downcast_tm_client_state(self)?;
+        client_state.verify_height(height)?;
+        new_verify_delay_passed(ctx, height, connection_end)?;
```
Farhad-Shabani, line 1168: I was figuring out why we needed to introduce `ctx` field for `new_verify_packet_acknowledgement` method (and also some other similar verification methods), I found out we use it inside this `verify_delay_passed(...)` of `TmClientState`. Looking at the inside, It seems all the operations are  ctx-related and there isn't any client-specific process there. Couldn't we keep it at the `ValidationContext` level?

crates/ibc/src/mock/context.rs
```diff
@@ -1990,6 +1990,50 @@ mod tests {
 
                 Acknowledgement::try_from(vec![1u8]).unwrap()
             }
+
+            #[cfg(feature = "val_exec_ctx")]
+            fn on_timeout_packet_validate(
+                &self,
+                _packet: &Packet,
+                _relayer: &Signer,
+            ) -> Result<(), crate::core::ics04_channel::error::PacketError> {
```
Farhad-Shabani: Can we have the `PacketError` imported by `use crate::core::ics04_channel::error::PacketError;` under the `val_exec_ctx` feature flag?

crates/ibc/src/core/ics04_channel/handler/acknowledgement.rs
```diff
@@ -11,6 +11,130 @@ use crate::events::IbcEvent;
...
+            let next_seq_ack = ctx_a.get_next_sequence_ack(port_chan_id_on_a)?;
+
+            if packet.sequence != next_seq_ack {
+                return Err(PacketError::InvalidPacketSequence {
+                    given_sequence: packet.sequence,
+                    next_sequence: next_seq_ack,
+                }
+                .into());
+            }
+        }
```
Farhad-Shabani: Wouldn't be better to move this check before the new_verify_packet_acknowledgement call so that it stops in error cases before doing that resource-intensive proof verification step?

Farhad-Shabani, line 133: Wouldn't be better to move this check before the "Verify proofs" section so that it stops in error cases before doing that more resource-intensive proof verification step?

Farhad-Shabani (commented):
Just a few suggestions, otherwise, all set from my end. tnx :ok_hand:

crates/ibc/src/mock/context.rs
plafer: [69d14e5](https://github.com/cosmos/ibc-rs/pull/393/commits/69d14e52857e6a23bda46a677bf9c660f80940ec)

crates/ibc/src/core/ics04_channel/handler/acknowledgement.rs
plafer: I think we did it that way to make sure we'd get the same errors as ibc-go given the same input (see [ibc-go here](https://github.com/cosmos/ibc-go/blob/a4f7761981771b453fdb06203dba33a62b4b6317/modules/core/04-channel/keeper/packet.go#L467-L490)), but this is not a requirement, and I agree with you that this is better.

[51ddc41](https://github.com/cosmos/ibc-rs/pull/393/commits/51ddc415db14241790459208f74451628491ee6c)

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: Although most things inside the implementation use the `ValidationContext`, I don't think it belongs there. Fundamentally, the delay period is a Connection concern; not sure why it isn't implemented somewhere in the "connection" part of the stack.

I think this is worth opening a low-priority issue to look into it. There is probably a larger change to be made with how we do `ExecutionContext::store_update_time` on `CreateClient` (see #246)). That is, how we store/retrieve update time (perhaps through the latest `ClientState`?) will influence how we refactor this.

Farhad-Shabani: Right. It's more concerned about the connection part. I opened an issue for this #404 .
Thanks for the context

crates/ibc/src/core/ics04_channel/handler/acknowledgement.rs
Farhad-Shabani: 🙏🏻

plafer:
@Farhad-Shabani last changes are in (regarding ICS-20 validation for acknowledgement and timeout callbacks):

[8915502](https://github.com/cosmos/ibc-rs/pull/393/commits/8915502598f65dba07dabfa5ba1508e2a5b1e94f)
[e39aaa9](https://github.com/cosmos/ibc-rs/pull/393/commits/e39aaa9054040bc9701a8480a960c15df7888a26)

Farhad-Shabani (approved):
:pray:

PR #411, opened 2023-02-09
Implement `dispatch()` for `ValidationContext` and `ExecutionContext`

### LINKED ISSUE #395: Add `dispatch()` function for `Val/ExecContext`
Convenience function which calls `validate()` then `execute()`.

plafer:
Closes: #395

This is not an efficient implementation, as there are operations performed twice between `validate()` and `execute()`, but it is correct.


Farhad-Shabani (approved):
:ok_hand:

Issue #319, opened 2023-01-03
The function parameters in `ValidationContext` should be references, while the functions in `ExecutionContext` should take ownership


plafer:
Analogous to #304 for `ValidationContext` and `ExecutionContext`.

PR #419, opened 2023-02-11
Refactor store_* and get_* methods to take *Path structs instead

### LINKED ISSUE #382: `ExecutionContext`: all `store_*` methods should take a `Path` as input
We forgot to update [these](https://github.com/cosmos/ibc-rs/blob/8ed5c7e1492256805185a9fab1557c4af0f59ff8/crates/ibc/src/core/context.rs#L520-L560).

Farhad-Shabani:
Closes: #382 

## Description

I initiated reorganizing the `store_*` methods, but it quickly escalated to more changes. 

- The `store_*` methods take a `Path_*` argument while the `get_*` methods had a different approach, accepting the needed parameters directly. IMHO, we should take a consistent approach throughout the repo. I believe that using `Path_*` is easier to maintain, understand the purpose of args, and align with the [spec](https://github.com/cosmos/ibc/tree/main/spec/core/ics-024-host-requirements#keyvalue-store).
- This change also had an impact on the `verify_*` methods under the `ClientState` trait and allowed us to eliminate some of the `#[allow(clippy::too_many_arguments)]` annotations.
- I also revised the naming of some `*Path` structs to be in the singular form as each abstraction refers to a specific path. This previous naming caused confusion while working with the paths. (See also [ibc-go](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/core/24-host/keys.go#L129-L227))
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/context.rs
+++ b/crates/ibc/src/core/context.rs
@@ -519,14 +519,14 @@ pub trait ExecutionContext: ValidationContext {
     /// Stores the given connection_end at path
     fn store_connection(
         &mut self,
-        connection_path: ConnectionPath,
+        connection_path: &ConnectionPath,
         connection_end: ConnectionEnd,
     ) -> Result<(), ContextError>;
 
     /// Stores the given connection_id at a path associated with the client_id.
     fn store_connection_to_client(
         &mut self,
-        client_connection_path: ClientConnectionPath,
+        client_connection_path: &ClientConnectionPath,
         conn_id: ConnectionId,
     ) -> Result<(), ContextError>;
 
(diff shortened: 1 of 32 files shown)


crates/ibc/src/core/context.rs
```diff
@@ -534,14 +519,14 @@ pub trait ExecutionContext: ValidationContext {
     /// Stores the given connection_end at path
     fn store_connection(
         &mut self,
-        connections_path: ConnectionsPath,
+        connection_path: ConnectionPath,
```
plafer: Contrary to what was stated in #319, I think all path arguments should be passed by reference, since either the host will generate the path string with `to_string()`, or read the internal fields to access its store data structure.

Farhad-Shabani: Because of the recent change, I was hesitant to say that :). My thought was there should be a reservation. 
But yes, I also think so

Farhad-Shabani: [4babf0f](https://github.com/cosmos/ibc-rs/pull/419/commits/4babf0f95c141908dc3f8fd6b9579cd343fd31fe)

plafer (approved):
🙏

PR #439, opened 2023-02-16
Fix Remove expect() in token transfer's acknowledgement conversion

### LINKED ISSUE #389: Remove `expect()` in token transfer's acknowledgement conversion
I'm also not a fan of needing that comment. Note also that I added a test to check that this conversion doesn't return a non-empty vector.

The fix would look like:
```rust
impl From<TokenTransferAcknowledgement> for NonEmptyVec<u8> {
    // ...
}

impl From<TokenTransferAcknowledgement> for Acknowledgement {
    fn from(ack: TokenTransferAcknowledgement) -> Self {
        let v: NonEmptyVec<u8> = ack.into();

        // this can be an `into()` because the only possible error case
        // (empty vector) is not possible with `NonEmptyVec<u8>`
        v.into()
    }
}
```

_Originally posted by @plafer in https://github.com/cosmos/ibc-rs/pull/377#discussion_r1091107329_

DaviRain-Su:
Closes: #389 

## Description

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/ics04_channel/msgs/acknowledgement.rs
```diff
@@ -56,6 +56,16 @@ impl TryFrom<Vec<u8>> for Acknowledgement {
     }
 }
 
+pub struct NoEmptyVec<T> {
+    pub data: Vec<T>,
+}
```
Farhad-Shabani, line 61: Isn't this struct still capable of accepting empty vectors?

Farhad-Shabani (commented):
Thanks for your contribution! 
Just wanted to point out when converting the `TokenTransferAcknowledgement` to `Acknowledgement` (like [here](https://github.com/cosmos/ibc-rs/blob/52c0a861d650c8688ec0ef0796fe8c9613997b8c/crates/ibc/src/applications/transfer/context.rs#L294)), it's important to go through [this](https://github.com/cosmos/ibc-rs/blob/52c0a861d650c8688ec0ef0796fe8c9613997b8c/crates/ibc/src/applications/transfer/acknowledgement.rs#L67-L70) match block to ensure that the `Acknowledgement` is properly constructed. Having the `NoEmptyVec`, we are bypassing this step. 

Also note that the output of the mentioned match would never be empty indeed. Though by design we ended up with that ‍`expect()`. Seems we should try another approach.

plafer:
what is the status of this PR? @DaviRain-Su it seems like you didn't address Farhad's review. Is this still something you're planning to work on?

DaviRain-Su:
> what is the status of this PR? @DaviRain-Su it seems like you didn't address Farhad's review. Is this still something you're planning to work on?

present, I have no any idea.

plafer:
I will close this as stale.

Issue #429, opened 2023-02-14
Refactor unit tests


plafer:
Our unit tests make extensive use of [default values](https://github.com/cosmos/ibc-rs/blob/0f3e03f84859d13cf842ec718eed3a2d8817a4d4/crates/ibc/src/core/ics04_channel/handler/timeout.rs#L398), including [in "util" functions](https://github.com/cosmos/ibc-rs/blob/0f3e03f84859d13cf842ec718eed3a2d8817a4d4/crates/ibc/src/core/ics04_channel/packet.rs#L313) implicitly used, which creates hard to understand relationships, and ultimately makes it difficult to understand why a test succeeds or fails. For example, some validation checks probably work because we used the default value for our `ClientId`, which happens to be also used when, say, creating the "dummy" packet.

I've also seen tests such as [this one](https://github.com/cosmos/ibc-rs/blob/0f3e03f84859d13cf842ec718eed3a2d8817a4d4/crates/ibc/src/core/ics04_channel/handler/timeout.rs#L416-L423) which claims a reason for the test to fail (e.g. the client doesn't have a consensus state at the required height), but no client is even installed in that `MockContext`.

plafer:
After we refactor our errors with #269, each test should confirm that the right error was emitted as opposed to [using `Result::is_err()`](https://github.com/cosmos/ibc-rs/blob/0f3e03f84859d13cf842ec718eed3a2d8817a4d4/crates/ibc/src/core/ics04_channel/handler/timeout_on_close.rs#L431).

Farhad-Shabani:
**One possible solution regarding defaults:**
Found it effective and can result in a cleaner, easier-to-follow code, not just for default values, even for any new object construction within our tests, is to implement and use more helper methods. This can exist for each type in its own module under the `mod test_util`. Then simply be called everywhere needed. Similar to what we did [here](https://github.com/cosmos/ibc-rs/blob/ee16dc2da38808e3c68c9bdf9fef42f69b3b627d/crates/ibc/src/core/ics03_connection/msgs/conn_open_init.rs#L89-L127) for `MsgConnectionOpenInit`, a `new_valid_dummy` method or something like that can be defined for e.g `CleintId`, `Height`, etc

rnbguy:
I think #1074 was a duplicate of this. If that is the case, we can close this after fixing this title.

PR #441, opened 2023-02-17
Refactor connection handler unit tests to adapt with new API

### LINKED ISSUE #440: Refactor connection handler unit tests to adapt with new Validation/Execution API
## Summary

Part of #413
It should be noted along with adaptation connection unit tests with the recent changes of ADR05, we refactor these tests for better readability and less code duplication

Farhad-Shabani:
Closes: #440

### Remark

- Reusable components are kept inside the `core::ics03_connection::handler`. If we decide to apply such a writing test approach for the rest of the parts, it should be moved to a higher level inside `core::handler`
- The `Option<ContextError>` type for the `Expect::Failure` should be `ContextError` to know what error exactly we catch in a failure scenario. But here we aimed to fulfill the refactoring purpose and this (along with other tests) should be revised with the next improvement iteration
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/ics04_channel/handler/chan_open_ack.rs
+++ b/crates/ibc/src/core/ics04_channel/handler/chan_open_ack.rs
@@ -213,9 +213,7 @@ mod tests {
     use crate::core::ics03_connection::connection::ConnectionEnd;
     use crate::core::ics03_connection::connection::Counterparty as ConnectionCounterparty;
     use crate::core::ics03_connection::connection::State as ConnectionState;
-    use crate::core::ics03_connection::msgs::conn_open_init::test_util::get_dummy_raw_msg_conn_open_init;
     use crate::core::ics03_connection::msgs::conn_open_init::MsgConnectionOpenInit;
-    use crate::core::ics03_connection::msgs::conn_open_try::test_util::get_dummy_raw_msg_conn_open_try;
     use crate::core::ics03_connection::msgs::conn_open_try::MsgConnectionOpenTry;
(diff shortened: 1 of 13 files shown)


crates/ibc/src/core/ics04_channel/handler/chan_open_ack.rs
```diff
@@ -246,8 +244,7 @@ mod tests {
 
         let context = MockContext::default();
 
-        let msg_conn_init =
-            MsgConnectionOpenInit::try_from(get_dummy_raw_msg_conn_open_init(None)).unwrap();
+        let msg_conn_init = MsgConnectionOpenInit::new_dummy();
```
plafer, line 247: Can we remove all changes in the `ics04_channel` module? This will cause a merge conflict with #430

Farhad-Shabani: There still would be conflict, because of the `get_dummy_raw_msg_conn_open_*()` changes and it does not take `None` anymore, How if we merge this PR after #430?

plafer: hmm, in that case let's just merge as is. We need to merge this first because the last step of #430 is to remove `impl Readers/Keepers for MockContext`, which needs this. They won't be big conflicts anyways

plafer (approved):
thank you for these high-quality tests! We can use them as a baseline when we implement #429

PR #460, opened 2023-02-21
`ValidationContext::{next,prev}_consensus_state()` don't take a Path

plafer:
Reverts changes in #419 made to `ValidationContext::{next,prev}_consensus_state()`

## Description

for these 2 methods, a `Path` argument doesn’t make sense. We don’t know the exact path that we are querying; *e.g.* we are asking the host to give us the consensus state after a certain height.


Farhad-Shabani (approved):
Thank you for taking care of this 🙏🏻

Issue #376, opened 2023-01-26
Docstring every public item of the new `{Validation,Execution}Context` API


plafer:
To be done before we remove the feature flag

plafer:
We need to make sure also that a chain that doesn't need to distinguish between "validation" and "execution" (e.g. "Cosmos SDK-like" chains) should not have to read documentation about it. Perhaps we should even introduce a different trait name (which is equivalent to `ExecutionContext`) for those who only need to use `dispatch()`.

PR #468, opened 2023-02-24
Update send packet/transfer

### LINKED ISSUE #442: Adapt `send_packet()` and `send_transfer()` API
These APIs (and their traits) retain a flavor the old `Reader`/`Keeper` API. Changes to make include 
(but are not limited to):
 
+ traits `store_*` methods should use `Path` arguments
+ Remove `SendPacketResult`
+ Probably don't need the `Reader`/`Keeper` separation
+ `send_{transfer,packet}()` should not return a `HandlerResult`, or use `HandlerOutput`
  + they should instead use `ModuleExtras`
+ split into `*_validate()` and `*_execute()`

plafer:
Closes: #442

+ Separates `send_{packet,transfer}()` into `*_validation()` and `*_execution()`
+ Removes old `ModuleOutput`/`ModuleOutputBuilder` structs
+ Removes unused callback methods in `Module`

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
+++ b/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
@@ -5,63 +5,7 @@ use crate::applications::transfer::packet::PacketData;
 use crate::applications::transfer::{is_receiver_chain_source, TracePrefix};
 use crate::core::ics04_channel::handler::ModuleExtras;
 use crate::core::ics04_channel::packet::Packet;
-use crate::core::ics26_routing::context::ModuleOutputBuilder;
 use crate::prelude::*;
-
-// TODO: REMOVE BEFORE MERGE
-pub fn process_recv_packet<Ctx: 'static + TokenTransferExecutionContext>(
-    ctx: &mut Ctx,
-    output: &mut ModuleOutputBuilder,
-    packet: &Packet,
-    data: PacketData,
-) -> Result<(), TokenTransferError> {
-    if !ctx.is_receive_enabled() {
-        return Err(TokenTransferError::ReceiveDisabled);
-    }
-
-    let receiver_account = data
-        .receiver
-        .clone()
-        .try_into()
-        .map_err(|_| TokenTransferError::ParseAccountFailure)?;
-
-    if is_receiver_chain_source(
-        packet.port_on_a.clone(),
-        packet.chan_on_a.clone(),
-        &data.token.denom,
-    ) {
-        // sender chain is not the source, unescrow tokens
-        let prefix = TracePrefix::new(packet.port_on_a.clone(), packet.chan_on_a.clone());
-        let coin = {
-            let mut c = data.token;
-            c.denom.remove_trace_prefix(&prefix);
-            c
-        };
-
-        let escrow_address =
-            ctx.get_channel_escrow_address(&packet.port_on_b, &packet.chan_on_b)?;
-
-        ctx.send_coins(&escrow_address, &receiver_account, &coin)?;
-    } else {
-        // sender chain is the source, mint vouchers
-        let prefix = TracePrefix::new(packet.port_on_b.clone(), packet.chan_on_b.clone());
-        let coin = {
-            let mut c = data.token;
-            c.denom.add_trace_prefix(prefix);
-            c
-        };
-
-        let denom_trace_event = DenomTraceEvent {
-            trace_hash: ctx.denom_hash_string(&coin.denom),
-            denom: coin.denom.clone(),
-        };
-        output.emit(denom_trace_event.into());
-
-        ctx.mint_coins(&receiver_account, &coin)?;
-    }
-
-    Ok(())
-}
 
 pub fn process_recv_packet_execute<Ctx: TokenTransferExecutionContext>(
     ctx: &mut Ctx,
--- a/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
+++ b/crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
@@ -1,4 +1,4 @@
-use crate::applications::transfer::context::TokenTransferContext;
+use crate::applications::transfer::context::TokenTransferExecutionContext;
 use crate::applications::transfer::error::TokenTransferError;
 use crate::applications::transfer::events::DenomTraceEvent;
 use crate::applications::transfer::packet::PacketData;
@@ -8,7 +8,8 @@ use crate::core::ics04_channel::packet::Packet;
 use crate::core::ics26_routing::context::ModuleOutputBuilder;
 use crate::prelude::*;
 
-pub fn process_recv_packet<Ctx: 'static + TokenTransferContext>(
+// TODO: REMOVE BEFORE MERGE
+pub fn process_recv_packet<Ctx: 'static + TokenTransferExecutionContext>(
     ctx: &mut Ctx,
     output: &mut ModuleOutputBuilder,
     packet: &Packet,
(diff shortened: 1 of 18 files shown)


crates/ibc/src/applications/transfer/relay/send_transfer.rs
```diff
@@ -1,33 +1,109 @@
...
+where
+    C: TryInto<PrefixedCoin> + Clone,
+    Ctx: TokenTransferExecutionContext,
+{
+    send_transfer_validate(ctx_a, msg.clone())?;
+    send_transfer_execute(ctx_a, msg)
+}
+
+pub fn send_transfer_validate<C, Ctx>(
+    ctx_a: &Ctx,
```
Farhad-Shabani, line 25: A side opinion about our naming convention (my experience so far)

While using "_a" and "_b" for message fields has been super helpful, I think we can ease up on naming contexts. We only implement one context for each host and always interact with that one context. Keeping "ctx" instead of "ctx_a" and "ctx_b" would be just as simple, effective, without adding unnecessary suffixes. Yet, wherever there could be a confusion, we still see our naming convention around message related fields and variables

Farhad-Shabani, line 25: A side opinion about our naming convention (my experience so far)

While using `*_a` and `*_b` for message fields has been super helpful, I think we can ease up on our sensitivity for naming contexts. (in particualr convection contains `_on_a` and `_of_a` not `_a` ) We only implement one context for each host and always interact with that one context. Keeping "ctx" instead of "ctx_a" and "ctx_b" would be just as simple, effective, without adding unnecessary suffixes. Yet, wherever there could be a confusion, we still see our naming convention around on a message related field or variable.

Farhad-Shabani, line 25: A side opinion about our naming convention (my experience so far)

While using `*_a` and `*_b` for message fields has been super helpful, I think we can ease up on naming contexts. (in particualr convection contains `*_on_a` and `*_of_a` not `*_a`) We only implement one context for each host and always interact with that one context. Keeping `ctx` instead of `ctx_a` and `ctx_b` would be just as simple, effective, without adding unnecessary suffixes. Yet, wherever there could be a confusion, we still see our naming convention around on a message related field or variable.

Farhad-Shabani, line 28: A side opinion about our naming convention (my experience so far)

While the convention for message fields has been super helpful, I think we can ease up on naming contexts. (in particualr convection contains `*_on_a` and `*_of_a` , not `*_a`) We only implement one context for each host and always interact with that one context. Keeping `ctx` instead of `ctx_a` and `ctx_b` would be just as simple, effective, without adding unnecessary complexity or suffixes. Yet, wherever there could be a confusion, we still see our naming convention around on a message related field or variable.

Farhad-Shabani, line 25: A side opinion about our naming convention (my experience so far)

While using the convention for message fields has been super helpful, I think we can ease up on naming contexts. (in particular we have *_on_a and *_of_a , not *_a ) we only implement one context for each host and always interact with that one context. Keeping ctx instead of ctx_a and ctx_b would be just as simple, effective, without adding unnecessary complexity or suffixes. Yet, wherever there could be confusion, we still see our naming convention on a message related field or variable.

crates/ibc/src/core/ics04_channel/handler/send_packet.rs
```diff
@@ -94,27 +92,49 @@ pub fn send_packet(
...
-        channel_id: packet.chan_on_a.clone(),
-        seq: packet.sequence,
-        seq_number: next_seq_send_on_a.increment(),
-        commitment: ctx_a.packet_commitment(
+    ctx_a.store_packet_commitment(
+        &CommitmentPath {
+            port_id: packet.port_on_a.clone(),
+            channel_id: packet.chan_on_a.clone(),
+            sequence: packet.sequence,
+        },
```
Farhad-Shabani: Can we rewrite the `CommitmentPath` using the `new()` method?

crates/ibc/src/applications/transfer/relay/send_transfer.rs
```diff
@@ -1,33 +1,109 @@
...
+use crate::core::ics04_channel::handler::send_packet::{send_packet_execute, send_packet_validate};
 use crate::core::ics04_channel::packet::Packet;
 use crate::core::ics24_host::path::{ChannelEndPath, SeqSendPath};
 use crate::events::ModuleEvent;
-use crate::handler::{HandlerOutput, HandlerOutputBuilder};
 use crate::prelude::*;
 
-/// This function handles the transfer sending logic.
-/// If this method returns an error, the runtime is expected to rollback all state modifications to
-/// the `Ctx` caused by all messages from the transaction that this `msg` is a part of.
```
Farhad-Shabani: Did this no longer correspond with the changes?

Farhad-Shabani (approved):
:ok_hand:

crates/ibc/src/applications/transfer/relay/on_recv_packet.rs
```diff
@@ -1,68 +1,13 @@
...
-
-        ctx.mint_coins(&receiver_account, &coin)?;
-    }
-
-    Ok(())
-}
-
-pub fn process_recv_packet_execute<Ctx: TokenTransferContext>(
+pub fn process_recv_packet_execute<Ctx: TokenTransferExecutionContext>(
     ctx: &mut Ctx,
```
Farhad-Shabani, line 11: `ctx` or `ctx_b`

crates/ibc/src/applications/transfer/relay/send_transfer.rs
plafer: Sorry, I'm not sure I understand. Are you suggesting we use `ctx` her here instead of `ctx_a`?

crates/ibc/src/core/ics04_channel/handler/send_packet.rs
plafer: [d07c6e8](https://github.com/cosmos/ibc-rs/pull/468/commits/d07c6e8e2051616ca535e16cfab1ab64d1b5e616)

crates/ibc/src/applications/transfer/relay/send_transfer.rs
Farhad-Shabani: Right. in general, I meant. Not just here

crates/ibc/src/applications/transfer/relay/send_transfer.rs
plafer: I added it back [2f1fce1](https://github.com/cosmos/ibc-rs/pull/468/commits/2f1fce19a5a14214682a09d5d19acf31470e4b48)

However this is true in general with `Validation/ExecutionContext`, so we should remember to document it more widely with #376.

crates/ibc/src/applications/transfer/relay/send_transfer.rs
plafer: This is following our convention (specifically [this part](https://github.com/cosmos/ibc-rs/blob/2f1fce19a5a14214682a09d5d19acf31470e4b48/crates/ibc/src/core/mod.rs#L16-L18)). I like this use of the convention, for example [here](https://github.com/cosmos/ibc-rs/blob/2f1fce19a5a14214682a09d5d19acf31470e4b48/crates/ibc/src/applications/transfer/relay/send_transfer.rs#L39), having `ctx_a` makes it easy to know I need to select `port_on_a` and `chan_on_a` to get the `ChannelEndPath` on "this chain" (the sending chain).

WDYT?

crates/ibc/src/applications/transfer/relay/send_transfer.rs
Farhad-Shabani: Agreed 👍🏻

crates/ibc/src/applications/transfer/relay/send_transfer.rs
Farhad-Shabani: Let's go with the present (ctx_a) format 👍🏻

PR #517, opened 2023-03-09
revert ibc-proto-rs and tendermint-rs upgrade

plafer:
tendermint-abci v0.30.0 doesn't support tendermint v0.34.0, which creates a problem for our integration tests with basecoin-rs. We can upgrade tendermint-rs at a later time.


Farhad-Shabani (approved):

plafer:
Note that before we move to tendermint v0.30.0, we need to make sure that none of our users are still using
1. tendermint-abci
2. tendermint v0.34

Issue #508, opened 2023-03-08
[Transfer App] Return Result instead of boolean for `is_send_enabled` and `is_receive_enabled‍‍` methods


Farhad-Shabani:
## Problem Definition

The `is_send_enabled` and `is_receive_enabled` methods under the `TokenTransferValidationContext` return a boolean value whose sole purpose is to propagate errors. This violates Rust's error-handling convention of using the Result type for error reporting.
This implementation was suitable as long as no specific error reporting was required, but it can become an issue as our codebase grow and require more detailed error reporting.

plafer:
> whose sole purpose is to propagate errors.

can you expand on that? IMO, the purpose is to [fail the transfer](https://github.com/cosmos/ibc-rs/blob/d0d0683e9257fd83ef2e00897b29adf5e8445072/crates/ibc/src/applications/transfer/relay/send_transfer.rs#L35-L37) if for any reason the hosts wants to freeze transfers. In other words, it's sort of a "kill switch" on transfers.

Farhad-Shabani:
> can you expand on that? IMO, the purpose is to [fail the transfer](https://github.com/cosmos/ibc-rs/blob/d0d0683e9257fd83ef2e00897b29adf5e8445072/crates/ibc/src/applications/transfer/relay/send_transfer.rs#L35-L37) if for any reason the hosts want to freeze transfers. In other words, it's sort of a "kill switch" on transfers.

Indeed, it is. What I meant:
- It's actually uncommon to use a `boolean` as a return type instead of `Result` to handle failures. In our case, we return a `boolean` and just afterward use it to handle failures. Why not just fail right away?
- Also, this way implementors have more control over the message of propagated error and can keep it consistent with their own project

Issue #270, opened 2022-11-29
Make core `Error` enums less specific


plafer:
Follow-up to #255.

Our `Error` enums have too many variants. We should drastically condense them following the ideas described [here](https://github.com/cosmos/ibc-rs/pull/255#issuecomment-1325805412).
```[tasklist]
### Tasks
- [ ] #1320 
- [ ] https://github.com/cosmos/ibc-rs/issues/1249
- [ ] https://github.com/cosmos/ibc-rs/issues/950
```

plafer:
A thought: ideally, internal functions that do use an `Error` enum would return an `Error` enum where the variants capture only the possible errors for that function. For example, [`verify_delay_passed()`](https://github.com/cosmos/ibc-rs/blob/3e139844c587ec96a7ca74f24128258b9880f2fb/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L233-L259)'s `Error` would only have the `TimestampOverflow`,`NotEnoughTimeElapsed`, and `NotEnoughBlocksElapsed` variants (or less if we don't care to be that specific). This provides better documentation of what can go wrong, and callers won't write code for errors that cannot happen.

plafer:
In #164, there was some push-back against using `anyhow` for our errors. However, looking at projects that use ibc-rs, none of them actually make use of our `Error` enum:

+ [Nomic's errors](https://github.com/nomic-io/orga/blob/beba8496d68eb81dd1e3c664de6a0dcacf371494/src/error.rs#L30) are modeled as `String`
+ [Namada make heavy use](https://github.com/anoma/namada/blob/d257f3ad7d77a0788bf0ba812d63e57d3a7009c8/core/src/ledger/ibc/context/execution.rs#L332) of the `Other(String)` error variant,
+ [Penumbra use](https://github.com/penumbra-zone/penumbra/blob/963ab0b887a1d24878be844dfac89717ee1aba17/component/src/action_handler/actions/ibc_action/msg/update_client.rs#L19) `anyhow` in their own implementation of IBC handlers

In other words, programmatic logic based on which error occurred is non-existent in all these projects. This tells me that the most important use case for our errors is aiding in debugging; either host chain implementations, or relayers such as Hermes.

[...]

PR #531, opened 2023-03-14
Supersede `ClientState` proof verification methods with generic interfaces

### LINKED ISSUE #530: [ICS02] Replace specific `verify_functions` with generic `verify_membership` and `verify_non_membership`
## Summary

In continuation of #404, part of #173, aiming toward https://github.com/cosmos/ibc/issues/684 implementation in IBC-rs
Replace specific verify_functions of `ClientState` trait with generic `verify_membership` and `verify_non_membership` methods

## Proposal
By separating the `verify_conn_delay_passed` process from the proof verification steps and isolating the concern, it has become apparent that the proof verification method in `ics02-clienstate` is carrying out a very similar task. With some minor refactoring, we can make use of generic interfaces and impose fewer implementation requirements on the builders.

### LINKED ISSUE #545: `ClientState`: replace `frozen_height()` and `is_frozen()` with `confirm_not_frozen()`
`frozen_height()` and `is_frozen()` together really serve one purpose: letting the core handlers know if the the client is frozen. We should replace it instead with just one method that does just that:

```rust
fn confirm_not_frozen(&self) -> Result<(), ClientError>;
```

Farhad-Shabani:
Closes: #530
Closes: #545
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -764,181 +753,31 @@ impl Ics2ClientState for ClientState {
...
-        verify_membership(
-            client_state,
-            prefix,
-            proof,
-            root,
-            channel_end_path.clone(),
-            value,
-        )
+        client_state.verify_height(proof_height)?;
+        verify_membership(client_state, prefix, proof, root, path, value)
```
plafer: I would remove this function and just inline the implementation, since this is the only place it's being used

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -764,181 +753,31 @@ impl Ics2ClientState for ClientState {
...
-        proof: &CommitmentProofBytes,
-        root: &CommitmentRoot,
-        receipt_path: &ReceiptPath,
-    ) -> Result<(), ClientError> {
-        let client_state = downcast_tm_client_state(self)?;
-        client_state.verify_height(height)?;
-
-        verify_non_membership(client_state, prefix, proof, root, receipt_path.clone())
+        client_state.verify_height(proof_height)?;
+        verify_non_membership(client_state, prefix, proof, root, path)
```
plafer: same here, I would just inline this

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -764,181 +753,31 @@ impl Ics2ClientState for ClientState {
...
-        prefix: &CommitmentPrefix,
-        proof: &CommitmentProofBytes,
-        root: &CommitmentRoot,
-        receipt_path: &ReceiptPath,
-    ) -> Result<(), ClientError> {
-        let client_state = downcast_tm_client_state(self)?;
-        client_state.verify_height(height)?;
-
-        verify_non_membership(client_state, prefix, proof, root, receipt_path.clone())
+        client_state.verify_height(proof_height)?;
```
plafer: our `verify_height` has a check on the frozen height, whereas [ibc-go doesn't](https://github.com/cosmos/ibc-go/blob/02d8975773583212a0fd4fcae2481ec7caf3b654/modules/light-clients/07-tendermint/client_state.go#L222-L227). Rather it just [rejects the transaction when the client is frozen](https://github.com/cosmos/ibc-go/blob/02d8975773583212a0fd4fcae2481ec7caf3b654/modules/core/03-connection/keeper/verify.go#L117-L119). It also makes more sense to me that `verify_membership` is not expected to do a client frozen height check.

Looking at our handlers, it seems like this `verify_height` is our only "is client frozen" check for connection handlers; all other handlers have the check. I think we should: 
1. add client frozen checks to connection handlers
2. remove the frozen check in this function
3. inline the function since it will only contain the height check

WDYT?

plafer (commented):
review still WIP, but got this so far!

plafer:
could you also run the integration tests on basecoin-rs, given that this is a critical change?

crates/ibc/src/core/ics04_channel/packet.rs
```diff
@@ -101,6 +103,18 @@ impl From<Sequence> for u64 {
...
+
+    fn try_from(seq: Sequence) -> Result<Self, Self::Error> {
+        let mut value = Vec::new();
+        u64::from(seq)
+            .encode(&mut value)
+            .expect("buffer size too small");
+        Ok(value)
+    }
+}
+
```
plafer: I don't think we should have this here; the protobuf way is one way to encode a `u64` to a byte array; this suggests that it is "the right way to do it".

plafer: we should also return an error instead of the `expect()`

crates/ibc/src/core/ics04_channel/error.rs
```diff
@@ -23,6 +24,8 @@ pub enum ChannelError {
     UnknownState { state: i32 },
     /// channel order type unknown: `{type_id}`
     UnknownOrderType { type_id: String },
+    /// invalid channel end error: `{0}`
```
plafer:
```suggestion
    /// invalid channel: `{0}`
```
I would also just take a `String` instead of a `ProtoError` to make the error more versatile.

crates/ibc/src/core/ics02_client/consensus_state.rs
```diff
@@ -55,6 +55,15 @@ pub fn downcast_consensus_state<CS: ConsensusState>(h: &dyn ConsensusState) -> O
...
 
+impl TryFrom<Box<dyn ConsensusState>> for Vec<u8> {
+    type Error = ClientError;
+
+    fn try_from(value: Box<dyn ConsensusState>) -> Result<Self, Self::Error> {
+        let value = value.encode_vec().map_err(ClientError::Encode)?;
+        Ok(value)
+    }
+}
+
```
plafer: same here, this implementation is just one of many ways to serialize a `Box<dyn ConsensusState>`, and so shouldn't live here

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -116,97 +109,27 @@ pub trait ClientState:
...
-        expected_counterparty_channel_end: &ChannelEnd,
-    ) -> Result<(), ClientError>;
-
-    /// Verify the client state for this chain that it is stored on the counterparty chain.
-    fn verify_client_full_state(
+    // Verify_membership is a generic proof verification method which verifies a
+    // proof of the existence of a value at a given Path at the specified height.
+    fn verify_membership(
         &self,
         proof_height: Height,
```
plafer: At first glance it seems odd to me that we are doing a height check as part of `verify_membership`, where the name suggests we're only doing a Merkle tree proof verification.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: e1f40ecd4977a513feb00acab41c3fd9d44f380a

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: e1f40ecd4977a513feb00acab41c3fd9d44f380a

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: That's indeed correct 👌🏻 
It was on my mind to cover it in another PR, but applied it now :). A bit different from what you mentioned here, see if you like it?

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: It is fixed because of refactoring height validations in [6258916](https://github.com/cosmos/ibc-rs/pull/531/commits/6258916640f39e7ec417145646df32308c0f40f0)

crates/ibc/src/core/ics02_client/consensus_state.rs
Farhad-Shabani: fcd8b791d29a82e838e7724c58ae4e804502bc29

crates/ibc/src/core/ics04_channel/error.rs
Farhad-Shabani: fcd8b791d29a82e838e7724c58ae4e804502bc29

crates/ibc/src/core/ics04_channel/packet.rs
Farhad-Shabani: fcd8b791d29a82e838e7724c58ae4e804502bc29

crates/ibc/src/core/ics04_channel/error.rs
```diff
@@ -23,6 +23,8 @@ pub enum ChannelError {
     UnknownState { state: i32 },
     /// channel order type unknown: `{type_id}`
     UnknownOrderType { type_id: String },
+    /// invalid channel end error: `{channel_end}`
```
plafer:
```suggestion
    /// invalid channel end: `{channel_end}`
```

Farhad-Shabani:
Let's see how [things](https://github.com/informalsystems/basecoin-rs/pull/86) will go for the basecoin-rs
Surprisingly, I only changed the branch for the ibc-rs dependency there!

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -42,14 +42,28 @@ pub trait ClientState:
     /// Latest height the client was updated to
     fn latest_height(&self) -> Height;
 
-    /// Freeze status of the client
-    fn is_frozen(&self) -> bool {
-        self.frozen_height().is_some()
+    /// Check if the given proof has a valid height for the client
+    fn validate_proof_height(&self, proof_height: Height) -> Result<(), ClientError> {
```
plafer: helper methods where the implementation must not be modified by the user should be private functions, similar to `verify_conn_delay_passed`; so `validate_proof_height` and `assert_not_frozen` should be private

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -42,14 +42,28 @@ pub trait ClientState:
...
     /// Frozen height of the client
     fn frozen_height(&self) -> Option<Height>;
 
+    /// Assert that the client is not frozen
+    fn assert_not_frozen(&self) -> Result<(), ClientError> {
+        if let Some(frozen_height) = self.frozen_height() {
+            return Err(ClientError::ClientFrozen { frozen_height });
+        }
+        Ok(())
+    }
```
plafer: I prefer `is_frozen()` than this interface because here we're forcing the error to be `ClientError::ClientFrozen`, whereas that might not be the most appropriate error everywhere. In general, I think a `Result<(), Error>` should be used when there is more than one error possible, otherwise a `bool` is more appropriate. WDYT?

(note that this should be a private function, as mentioned in other comment)

crates/ibc/src/core/ics03_connection/handler/conn_open_try.rs
```diff
@@ -58,6 +58,10 @@ where
                 .map_err(|_| ConnectionError::Other {
                     description: "failed to fetch client state".to_string(),
                 })?;
+
+        client_state_of_a_on_b.assert_not_frozen()?;
```
plafer: we're missing an `assert_not_frozen()` in `conn_open_init`

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
```diff
@@ -71,13 +71,8 @@ where
...
         let client_state_of_a_on_b = ctx_b.client_state(client_id_on_b)?;
 
-        // The client must not be frozen.
-        if client_state_of_a_on_b.is_frozen() {
-            return Err(PacketError::FrozenClient {
-                client_id: client_id_on_b.clone(),
-            }
-            .into());
-        }
+        client_state_of_a_on_b.assert_not_frozen()?;
```
plafer: Everywhere that we get the `ClientState` in a validation function, we need to `assert_not_frozen()`. Perhaps we could have a `get_client_state_validation()` private function which would run the `assert_not_frozen()` check after getting the `ClientState` from the host?

crates/ibc/src/core/ics04_channel/handler/chan_open_try.rs
```diff
@@ -48,6 +48,10 @@ where
     {
         let client_id_on_b = conn_end_on_b.client_id();
         let client_state_of_a_on_b = ctx_b.client_state(client_id_on_b)?;
+
+        client_state_of_a_on_b.assert_not_frozen()?;
```
plafer: Similarly, the `chan_open_init`, `chan_close_init` and `chan_close_confirm` need an `assert_not_frozen` check as well

plafer: note that using a `get_client_state_validation()` would solve this issue

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: I see them differently. Both `assert_not_frozen` and `validate_proof_height` can get different implementations from the default. e.g. in the Parachains, the `proof_height` can have the same height as the `update_client` message. The frozen clients may also freeze with timestamps in some places. I think this one needs more investigation later if you agree

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: As mentioned [here](https://github.com/cosmos/ibc-rs/pull/531#discussion_r1136151279), potentially there could be other error variants. thought, IMHO, I think returning a boolean is not a convention in places like here.

crates/ibc/src/core/ics04_channel/handler/chan_open_try.rs
Farhad-Shabani: Good catch 👌🏻 
These were missed before and it came to light

crates/ibc/src/core/ics02_client/client_state.rs
plafer: > in the Parachains, the proof_height can have the same height as the update_client message

If that's the case, this implementation should go in the tendermint light client, and `validate_proof_height` should be abstract.

> The frozen clients may also freeze with timestamps in some places.

Not in my understanding; a client freezes when a misbehaviour is detected, which *always* happens at a precise height. Also IBC [mandates](https://github.com/cosmos/ibc/blob/main/spec/core/ics-002-client-semantics/README.md#clientstate) the concept of "height", so we know for sure that all IBC chains will have this concept. Therefore, all "client freezes" will have a precise height associated with it, and thus our current implementation of `assert_not_frozen`/`is_frozen` is the only correct one.

crates/ibc/src/core/ics02_client/client_state.rs
plafer: > IMHO, I think returning a boolean is not a convention in places like here.

What makes you say that? Using `bool` definitely is idiomatic in Rust. For example, see [tokio](https://docs.rs/tokio/latest/tokio/io/struct.Ready.html#method.is_empty), [`std::vec::Vec::is_empty`](https://docs.rs/tokio/latest/tokio/io/struct.Ready.html#method.is_empty), [`Option::is_some`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some), [`std::path::Path::is_file`](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file), etc etc etc

Farhad-Shabani: Just to clarify, I didn't mean all over places in Rust. I was referring to _places like this_ and hoped we were on the same page from our [previous](https://github.com/cosmos/ibc-rs/issues/508#issuecomment-1460572951) discussion, so I didn't elaborate. Apologies if I confused you. 

Here are my two cents:

In the places `is_frozen` returns a `bool`, just afterward we only use that true/false to handle a failure. 
This resulted in:
1. Instead of only one implementation and simply using `?` to propagate the error, we have to take care the “true” condition (error variant) everywhere **(One vs Multiple impls)**
2. In action, having to handle multiple places, we easily can end up with inconsistent implementations (e.g. see [here](https://github.com/cosmos/ibc-rs/blob/45d5f0d25f323eb423d5511e2488be8104e1c39f/crates/ibc/src/core/ics02_client/handler/update_client.rs#L33-L35), and [here](https://github.com/cosmos/ibc-rs/blob/707182148a7df5916307af112128a6354dd9808b/crates/ibc/src/hosts/tendermint.rs#L30-L35)) and different yet similar error variants.(e.g. see [here](https://github.com/cosmos/ibc-rs/blob/088861bf573634d3179709e414b3176e7a893091/crates/ibc/src/core/ics04_channel/error.rs#L102-L103), [here](https://github.com/cosmos/ibc-rs/blob/088861bf573634d3179709e414b3176e7a893091/crates/ibc/src/core/ics04_channel/error.rs#L79-L80), [here](https://github.com/cosmos/ibc-rs/blob/d0d0683e9257fd83ef2e00897b29adf5e8445072/crates/ibc/src/clients/ics07_tendermint/error.rs#L88-L92), and [here](https://github.com/cosmos/ibc-rs/blob/088861bf573634d3179709e414b3176e7a893091/crates/ibc/src/core/ics02_client/error.rs#L24-L25)) **(Inconsistency)**

Further, since `assert_not_frozen` is only the by-product of misbehavior detection, I still see it as a chain and client-specific process. Builders should be able to include their own desirable error message to reflect the reason for the client freezing.

  3. This approach gives more control over their error handling. Otherwise, we should assume the reason/message or only can provide minimal info about the error. **(Customizable Errors)**
    

(Side note: I didn’t go through well-defined errors in this PR, since a serious and thorough evaluation is needed on that end) 

Also, note that the [examples](https://github.com/cosmos/ibc-rs/pull/531#discussion_r1136179670) you gave were implemented on **structs/enums** where end users can't have another implementation and just call and use them. Here, _in case of failures/errors_, if you don't want to impose your error system - unlike us :) - it's actually better to just return true/false and let users manage the error variant themselves.

One last thing, regarding your [rationale](https://github.com/cosmos/ibc-rs/pull/531#discussion_r1136173995). I agree with the height mandate, but still protocol does not restrict you to freeze a client only with height, and e.g. it is possible to set a state on a certain timestamp in the Solana ecosystem. 
That was why I liked the way `ibc-go` handle this and assume less about the clients giving a more generic client-neutral [`status()`](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/core/exported/client.go#L52-L53) API. It was on my mind, so I asked you to investigate this in a later issue.

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: See [here](https://github.com/cosmos/ibc-rs/pull/531#discussion_r1137077485) please

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
Farhad-Shabani: I'm not sure if I caught you correctly. Wouldn't this only replace a line with `client_state.assert_not_frozen()` with another`get_client_state_validation()`? Can you a bit expand?

crates/ibc/src/core/ics02_client/client_state.rs
plafer: There are a lot of moving parts here (error system, API design, `bool` vs `Result`), which makes it hard to have a discussion about just one issue at a time. I’ll do my best to isolate concerns.

## Mechanism for telling whether a client is frozen

`is_frozen()` was *designed* to only be a convenience function that calls `frozen_height`, which is why I was suggesting we make it private. IIUC, you’re suggesting to move away from that design, and make `is_frozen()`/`assert_not_frozen()` the primary method that hosts implement to tell us when a client is frozen. That can work too; however if this is the case, then `frozen_height()` should be removed. Otherwise we have 2 potentially inconsistent mechanisms in our API for telling if a client is frozen (`frozen_height` could return `None`, while `assert_not_frozen` could return `Err(…)`).

About ibc-go status API: we should consider adopting a similar API, but let’s not discuss that now. Whatever we come out of this discussion should be self-consistent; whether or not we switch to a “client status API” is a separate topic.

**Conclusion from this section:** let’s remove `frozen_height()`

## Whether users should control the return error from `assert_not_frozen`

I can see now that there could be value in letting users construct the `Error` for when a client is frozen, as they could include more information about why the client is frozen. However, note that we're hitting a limitation of our error system once again, because the `ClientError::ClientFrozen` only takes a `Height` (no custom error message possible). To include a custom error message, they would need to use `ClientError::Other`, but that defeats the purpose of having an `Error` enum at all. This is just another example of why we need to improve our error system.

**Conclusion from this section:** let's keep `assert_not_frozen` with the current signature. Since `frozen_height` is removed, it will need to be an abstract method. Users how want to customize the error message can use `ClientError::Other` until we improve the error system.

## About the name `assert_not_frozen`

To me names of the form `assert_*` suggest that the function will panic on error. How about `confirm_not_frozen`? Or something else?

## Side notes

> I agree with the height mandate, but still protocol does not restrict you to freeze a client only with height, and e.g. it is possible to set a state on a certain timestamp in the Solana ecosystem.

I agree that the protocol does not restrict you to decide whether a client is frozen based on height; my point was that no matter how a chain comes to a conclusion about when a client becomes frozen, there will always be a height associated with that moment. I was saying this in the context of explaining why the `frozen_height()` interface was sound. But now we’re moving away from the `frozen_height` API, so this is not so important anymore.


> In action, having to handle multiple places, we easily can end up with inconsistent implementations (e.g. see here, and here) and different yet similar error variants.(e.g. see here, here, here, and here) (Inconsistency)

This is a great example of why our error system needs to be improved (#270).

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
plafer: I'm suggesting to replace

```rust
let client_state_of_a_on_b = ctx_b.client_state(client_id_on_b)?;
client_state_of_a_on_b.assert_not_frozen()?;
```

with 

```rust
// call site
let client_state_of_a_on_b = get_client_state_validation(&ctx_b);


// pseudo-code implementation
fn get_client_state_validation(ctx: &Ctx) {
    let client_state_of_a_on_b = ctx_b.client_state(client_id_on_b)?;
    client_state_of_a_on_b.assert_not_frozen()?;

    client_state_of_a_on_b
}
```

We can probably come up with a better name

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: I opened https://github.com/cosmos/ibc-rs/issues/536 for this. It even looks like we should also move away from `assert_not_frozen` method. If it's ok with you? here I only revise the naming and keep the rest to address in  #536.

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
Farhad-Shabani: > let client_state_of_a_on_b = ctx_b.client_state(client_id_on_b)?;
 client_state_of_a_on_b.assert_not_frozen()?;

Isn't the above lines self-explaining enough? If we weren't to use `client_state_of_a_on_b` afterward, having this function would make sense more. but, honestly, I don't see value adding a further function here

crates/ibc/src/core/ics02_client/client_state.rs
plafer: why not remove `frozen_height` as well?

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
plafer: > Isn't the above lines self-explaining enough?

My concern is that after a few refactors we'd forget to check if the client is frozen. Basically, since in validation, we always want to check that the client is not frozen right after receiving it, this new function would be a tool to not forget that.

>  If we weren't to use client_state_of_a_on_b afterward, having this function would make sense more.

What do you mean? `get_client_state_validation` would return the `ClientState` that we just fetched.

Basically I'm looking for a means to ensure that we never forget to do that check over time. Another potential solution would be to ensure that we have a unit test for each handler that will fail if we forget that check.

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: The changes we've had here were around the `validate_*` methods though inevitably we touched `is_fronzen` (bc of error variants). I am suggesting this since it is concerned with a different logical change that can also impact (remove) `confirm_not_frozen` and `expired` methods, for the sake of following the changes and a better changelog.

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
Farhad-Shabani: I understand your concern, but I am afraid we achieve this by having this function. We still should be cautious to not miss `get_client_state_validation` wherever needed. Definitely, well-written tests will avoid this issue. 👌🏻

Farhad-Shabani: Interestingly, with this refactor, we caught test failure in places we missed introducing a `Client` to the `MockContext`.  [425ccb2](https://github.com/cosmos/ibc-rs/pull/531/commits/425ccb2f66280cd0ef8c6234bdc22500711c42a4)

crates/ibc/src/core/ics02_client/client_state.rs
plafer: We didn't *have* to touch `is_frozen`; I agree this would have been better handled in a different PR. I think it'd better to just revert the `assert_not_frozen()` changes; we expect to change it again, I am not convinced that this state is better than the previous, and this change is not covered in the changelog. But ultimately I think we had some good discussions around it, and we have a better idea of where we want to go with it.

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
plafer: > Definitely, well-written tests will avoid this issue. 

I'm good with that. Unless you already verified that we have these tests, we should create an issue to track it.

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: >  I am not convinced that this state is better than the previous

Note that changes around the `assert_not_frozen` differs. We ended up with the `assert_not_frozen` since changes in this PR required handling  **failures/errors** everywhere and I thought you agreed given the above reasoning. (**What should be the signature?**)

This change is different from the logical decision that should be made around keeping or dropping **status-related** interfaces (`asssert_not_frozen`, `frozen_height` and `is_expired`) or even gathering them under one interface named `status()` method. (vs **achieve chain-neutral status APIs**)

> and different yet similar error variants.(e.g. see [here](https://github.com/cosmos/ibc-rs/blob/088861bf573634d3179709e414b3176e7a893091/crates/ibc/src/core/ics04_channel/error.rs#L102-L103), [here](https://github.com/cosmos/ibc-rs/blob/088861bf573634d3179709e414b3176e7a893091/crates/ibc/src/core/ics04_channel/error.rs#L79-L80), [here](https://github.com/cosmos/ibc-rs/blob/d0d0683e9257fd83ef2e00897b29adf5e8445072/crates/ibc/src/clients/ics07_tendermint/error.rs#L88-L92), and [here](https://github.com/cosmos/ibc-rs/blob/088861bf573634d3179709e414b3176e7a893091/crates/ibc/src/core/ics02_client/error.rs#L24-L25)) (Inconsistency)

Reverting `assert_not_frozen` would revert all the above error variants and everywhere we should use a couple of `map_err` and deal with this inconsistency

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
Farhad-Shabani: It just captured places we missed the client, but `MockContext` does not cover frozen/expired conditions. 
Opened this https://github.com/cosmos/ibc-rs/issues/538.

crates/ibc/src/core/ics03_connection/connection.rs
```diff
@@ -113,6 +112,17 @@ pub struct ConnectionEnd {
...
+    type Error = ConnectionError;
+
+    fn try_from(value: ConnectionEnd) -> Result<Self, Self::Error> {
+        let value = value
+            .encode_vec()
+            .map_err(ConnectionError::InvalidConnectionEnd)?;
+        Ok(value)
+    }
+}
+
```
plafer: oops we missed that one; for the same reasons as the others, we should remove this `TryFrom`

crates/ibc/src/core/ics04_channel/handler/recv_packet.rs
```diff
@@ -70,13 +71,8 @@ where
...
         let client_state_of_a_on_b = ctx_b.client_state(client_id_on_b)?;
 
-        // The client must not be frozen.
-        if client_state_of_a_on_b.is_frozen() {
-            return Err(PacketError::FrozenClient {
-                client_id: client_id_on_b.clone(),
-            }
-            .into());
-        }
+        client_state_of_a_on_b.confirm_not_frozen()?;
```
plafer, line 74: ibc-go only checks if client is frozen for `send_packet` (of all packet-related handlers). Waiting on response from them for why that is

plafer: wrong; all the handlers do make that check as expected

crates/ibc/src/core/ics04_channel/handler/timeout.rs
```diff
@@ -119,26 +121,33 @@ where
             }
             let seq_recv_path_on_b =
                 SeqRecvPath::new(&msg.packet.port_id_on_b, &msg.packet.chan_id_on_b);
-            client_state_of_b_on_a.verify_next_sequence_recv(
-                msg.proof_height_on_b,
+
```
plafer, line 126: on line 82, we should be checking if the client is frozen

crates/ibc/src/core/ics03_connection/handler/conn_open_try.rs
```diff
@@ -58,6 +59,10 @@ where
                 .map_err(|_| ConnectionError::Other {
                     description: "failed to fetch client state".to_string(),
                 })?;
+
+        client_state_of_a_on_b.confirm_not_frozen()?;
```
plafer, line 63: `conn_open_init` also should have that check

crates/ibc/src/core/ics04_channel/handler/chan_open_init.rs
```diff
@@ -21,6 +21,10 @@ where
     // An IBC connection running on the local (host) chain should exist.
     let conn_end_on_a = ctx_a.connection_end(&msg.connection_hops_on_a[0])?;
 
+    // let client_id_on_a = conn_end_on_a.client_id();
+    // let client_state_of_b_on_a = ctx_a.client_state(client_id_on_a)?;
+    // client_state_of_b_on_a.confirm_not_frozen()?;
```
plafer: this should not be commented out

crates/ibc/src/core/ics03_connection/connection.rs
Farhad-Shabani: Also, we can keep them inside a private module, so use the benefit of this conversion internally. WDYT?
```rust
mod private {
use super::*;
-----
}

plafer: I can see the value of having these conversions internally yep. But I'd do helper functions with different names; e.g. `ConnectionEnd::proto_encode_vec()` or something like that. Again, because I'd reserve the use of `TryFrom` only for when there's one way to do the conversion

crates/ibc/src/core/ics04_channel/handler/chan_open_init.rs
Farhad-Shabani: [828b40a](https://github.com/cosmos/ibc-rs/pull/531/commits/828b40a5a04d351041b190841fd281e9ac68f467)

crates/ibc/src/core/ics04_channel/handler/timeout.rs
Farhad-Shabani: [828b40a](https://github.com/cosmos/ibc-rs/pull/531/commits/828b40a5a04d351041b190841fd281e9ac68f467)

crates/ibc/src/core/ics03_connection/connection.rs
Farhad-Shabani: [bfe0004](https://github.com/cosmos/ibc-rs/pull/531/commits/bfe0004cadf558ff5c0fad84f6fdd75e7a0489f5)

plafer (approved):
🎉

PR #542, opened 2023-03-16
Fix Pass ChannelId by value instead of reference since it impls Copy

### LINKED ISSUE #537: [Crate] Pass `ChannelId` by value instead of reference since it impls `Copy`
## Summary

Part of #25, originated from https://github.com/informalsystems/hermes/pull/2071#discussion_r845063824

We should revisit the codebase and check wherever this comment applies pass the `ChannelId` by value instead of getting the reference since it impls the `Copy`

DaviRain-Su:
Closes: #537 

## Description

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -412,9 +412,8 @@ impl Default for PortId {
     feature = "borsh",
     derive(borsh::BorshSerialize, borsh::BorshDeserialize)
 )]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ChannelId(String);
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ChannelId(u64);
```
kevinji, line 416: Doesn't this break compatibility with other IBC chains? I don't think [the spec](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/README.md) requires that the channel ID be a `u64`, only that it's an `Identifier`, [which is defined as](https://github.com/cosmos/ibc/tree/main/spec/core/ics-024-host-requirements#paths-identifiers-separators)

> An `Identifier` is a bytestring used as a key for an object stored in state, such as a connection, channel, or light client.
>
> Identifiers MUST be non-empty (of positive integer length).
> 
> Identifiers MUST consist of characters in one of the following categories only:
> - Alphanumeric
> - `.`, `_`, `+`, `-`, `#`
> - `[`, `]`, `<`, `>`

plafer: yes indeed, I did some digging into the history of this issue. It was created when `ChannelId` was a `u64`. However we changed it to `String` in [this PR](https://github.com/informalsystems/hermes/pull/2361), for exactly these reasons. I will close this PR and the issue as "won't do".

Thank you @DaviRain-Su for ultimately bringing this to light; sorry for needing to throw away all this code.

plafer:
See https://github.com/cosmos/ibc-rs/pull/542#discussion_r1144928669

PR #556, opened 2023-03-17
Fix timeout validation for timeout height

### LINKED ISSUE #443: implement missing timeout packet test
[This one](https://github.com/cosmos/ibc-rs/blob/ccd17e00c5b5281fd1ee6cfb0a24da2a199efbe4/crates/ibc/src/core/ics04_channel/handler/timeout.rs#L470-L473).

### LINKED ISSUE #555: Validating timeout failed when height timeout
## Bug Summary
If `timeout_height_on_b` in a packet reached, the timeout handling failed.

## Details
The handling returns `PacketTimeoutHeightNotReached` error if the timeout height has passed.
```rust
        if msg
            .packet
            .timeout_height_on_b
            .has_expired(msg.proof_height_on_b)
        {
            return Err(PacketError::PacketTimeoutHeightNotReached {
                timeout_height: msg.packet.timeout_height_on_b,
                chain_height: msg.proof_height_on_b,
            }
            .into());
        }
```
It should return an error only when both the timeout height and timestamp have not been reached.

## Version
0.32.0

yito88:
Closes: #555
Closes: #443

## Description
Only when both the timeout height and timestamp have not passed yet, the timeout handler returns an error `PacketTimeoutNotReached`.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/ics04_channel/handler/timeout.rs
+++ b/crates/ibc/src/core/ics04_channel/handler/timeout.rs
@@ -347,10 +347,8 @@ mod tests {
         let res = validate(&context, &msg);
 
         assert!(
-            res.unwrap_err()
-                .to_string()
-                .starts_with("ICS04 Packet error: Packet timeout"),
-            "Validation should fail because both the timeout height and timestamp have not reached yet"
+            res.is_err(),
+            "Validation should fail because the timeout height was reached, but the timestamp hasn't been reached. Both the height and timestamp need to be reached for the packet to be considered timed out"
         )
     }
 
--- a/crates/ibc/src/core/ics04_channel/handler/timeout.rs
+++ b/crates/ibc/src/core/ics04_channel/handler/timeout.rs
@@ -350,7 +350,7 @@ mod tests {
             res.unwrap_err()
                 .to_string()
                 .starts_with("ICS04 Packet error: Packet timeout"),
-            "Validation should fail because the timeout height has not reached yet"
+            "Validation should fail because both the timeout height and timestamp have not reached yet"
         )
     }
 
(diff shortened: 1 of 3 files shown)


crates/ibc/src/core/ics04_channel/handler/timeout.rs
```diff
@@ -306,9 +292,66 @@ mod tests {
...
+                Height::new(0, 4).unwrap(),
+            )
+            .unwrap();
+
+        let res = validate(&context, &msg);
+
+        assert!(
+            res.unwrap_err()
+                .to_string()
+                .starts_with("ICS04 Packet error: Packet timeout"),
```
plafer: As inconvenient as it is, can we match on the error instead of matching on the string representation?

(Another good example why our current errors are difficult to use: #270).

crates/ibc/src/core/ics04_channel/handler/timeout.rs
```diff
@@ -306,9 +292,66 @@ mod tests {
...
+            "Validation should fail because both the timeout height and timestamp have not reached yet"
```
plafer:
```suggestion
            "Validation should fail because the timeout height was reached, but the timestamp hasn't been reached. Both the height and timestamp need to be reached for the packet to be considered timed out"
```

crates/ibc/src/core/ics04_channel/handler/timeout.rs
plafer: or just `is_err()` if it's too unwieldy

plafer:
~~@yito88 any idea why the CI suddenly fails?~~

this is happening in other PRs too, will investigate

plafer (approved):
Thank you!

PR #561, opened 2023-03-22
commit CI `no-std-check`'s Cargo.lock

plafer:


romac:
I think the title and comment are a bit misleading because with this change only the `no-std-check` CI job will use the lockfile. All other jobs will not use any lockfile whatsoever.

plafer:
@romac see if the new version is more clear

romac (approved):

PR #568, opened 2023-03-27
Improve `MsgTransfer`

### LINKED ISSUE #567: Improve `MsgTransfer` struct
Our current [`MsgTransfer`](https://github.com/cosmos/ibc-rs/blob/8cc943d67b5aa28b09e4d1c94a563f9f35e53608/crates/ibc/src/applications/transfer/msgs/transfer.rs#L27) struct has 2 flaws.

1. In our conversion from the raw type, we [don't fully convert](https://github.com/cosmos/ibc-rs/blob/8cc943d67b5aa28b09e4d1c94a563f9f35e53608/crates/ibc/src/applications/transfer/msgs/transfer.rs#L83) the token field. The conversion is instead [completed](https://github.com/cosmos/ibc-rs/blob/8cc943d67b5aa28b09e4d1c94a563f9f35e53608/crates/ibc/src/applications/transfer/relay/send_transfer.rs#L57-L60) in the `send_transfer` function.
2. #559 requires us to add a `memo` field to both `MsgTransfer` and [`PacketData`](https://github.com/cosmos/ibc-rs/blob/8cc943d67b5aa28b09e4d1c94a563f9f35e53608/crates/ibc/src/applications/transfer/packet.rs#L17C3-L21). This is not [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Instead of copy/pasting a subset of fields from `MsgTransfer` to `PacketData`, `MsgTransfer` should instead contain a `PacketData` field.

Note that this is an example of us changing a domain type, which currently requires a breaking change (#324).

plafer:
Closes: #567

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/breaking-changes/567-msgtransfer.md
+++ b/.changelog/unreleased/breaking-changes/567-msgtransfer.md
@@ -1,0 +1,2 @@
+- Improve MsgTransfer struct
+  ([#567](https://github.com/cosmos/ibc-rs/issues/567))
--- a/crates/ibc/src/applications/transfer/msgs/transfer.rs
+++ b/crates/ibc/src/applications/transfer/msgs/transfer.rs
@@ -1,8 +1,8 @@
 //! This is the definition of a transfer messages that an application submits to a chain.
 
+use crate::applications::transfer::packet::PacketData;
 use crate::prelude::*;
 
-use ibc_proto::cosmos::base::v1beta1::Coin;
 use ibc_proto::google::protobuf::Any;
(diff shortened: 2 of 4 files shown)


Farhad-Shabani (approved):
All set.
There is just an inconsistency with the naming of some fields and methods I remember you also mentioned this before. Some places we use "token" and other times we use "coin". Do we need to deal with that in a later PR?

plafer:
yes I can always do that in the next PR 👍

Issue #536, opened 2023-03-15
`ClientState`: consider replacing `confirm_frozen` and `expired` methods with a "status" instead


Farhad-Shabani:
## Problem Statement
Context: https://github.com/cosmos/ibc-rs/pull/531#discussion_r1137077485

Currently, the `ClientState` trait [exposes](https://github.com/cosmos/ibc-rs/blob/d0d0683e9257fd83ef2e00897b29adf5e8445072/crates/ibc/src/core/ics02_client/client_state.rs#L52-L62) three interfaces: `is_frozen`, `expired`, and `frozen_height` for checking the client status. However, these methods appear to make some underlying assumptions about the characteristics of the host chain. 
Further investigation is required to ensure their compatibility with any chains. To address this issue, it is suggested that we adopt the approach taken by `IBC-go`, which provides a more practical and robust [`status()`](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/core/exported/client.go#L52-L53) API.

Issue #324, opened 2023-01-05
Make `Msg*` structs `pub(crate)`


plafer:
This will allow us to make changes to our message domain types on breaking change.

plafer:
After further analysis, this is a pretty big change that will affect our core API. We currently expect the user to construct a `MsgEnvelope` and call `dispatch()` (AFAIK no project uses `deliver()`). So we're currently exposing all our domain `Msg` types so that users can construct one of them, wrap it in `MsgEnvelope`, and call `dispatch()`.

A better API would be to ask the users to pass in the protobuf `Message` types (i.e. the "raw" types) directly; this way we can make all our domain types private and change them without any breaking change. Presumably, projects will be using the raw types directly in their transaction format anyways, so I don't expect any friction there. Concretely, I suggest we repurpose the `Msg` trait to be implemented on all raw `Msg` types, and make `validate()`/`execute()` take a generic type bounded by `Msg`.

This being said, we should wait until we have fully removed the legacy `Reader`s and `Keeper`s to avoid refactoring code that will be removed soon.

blocked: #279

plafer:
Note that the proto struct definitions can sometimes have a field that changes name from one version to the next, since protobuf's encoding is [based on field numbers](https://protobuf.dev/programming-guides/encoding/#structure) (i.e. the field names can change without changing what the struct serializes to).

This actually happened with `MsgUpdateClient`: from ibc-go v6 to v7, the [`header`](https://github.com/cosmos/ibc-go/blob/1391a52b168f647eca04da2ac7208cdfe5db9353/proto/ibc/core/client/v1/tx.proto#L51) field was renamed to [`client_message`](https://github.com/cosmos/ibc-go/blob/7fb65c76000f207ece1b83d8950e3aaff3dfc0e7/proto/ibc/core/client/v1/tx.proto#L51).

One possible solution is to have a thin wrapper around these structs that only keeps the naming constant; whenever a field name changes in the protobuf struct, we map that new name to the analogous field with the "old name".

plafer:
Another reason why I am in favor of making the domain `Msg` types private is that the representation we have at any one point in time is tied to other decisions specific to the your library. 

[...]

PR #584, opened 2023-03-30
Replace `check_header_and_update_state` with an architecture similar to ibc-go

### LINKED ISSUE #535: [ICS02] Split `check_misbehaviour_and_update_state` and `check_header_and_update_state` into two sets of validation/execution functions
## Problem Statement

Part of #173, in continuation of implementing [IBC-go/ADR006](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-006-02-client-refactor.md) and getting `check_header_and_update_state()` and  `check_misbehaviour_and_update_state()` compatible with the Validation/Execution design ([ADR005](https://github.com/cosmos/ibc-rs/blob/main/docs/architecture/adr-005-handlers-redesign.md))

```[tasklist]
### Related Issues
- [ ] https://github.com/cosmos/ibc-rs/issues/583
- [ ] https://github.com/cosmos/ibc-rs/issues/585
- [ ] https://github.com/cosmos/ibc-rs/issues/589
- [ ] https://github.com/cosmos/ibc-rs/issues/598
- [ ] https://github.com/cosmos/ibc-rs/issues/601
```

### LINKED ISSUE #583: tendermint light client: `check_{header,misbehaviour}_and_update_state()` missing `trusted_validator_set` hash check
## Bug Summary
Problem is [here](https://github.com/cosmos/ibc-rs/blob/20d682fcb0652e27f8dd6e3808e4c7c337e11790/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L500-L501).

We need to add a check `hash(header.trusted_validator_set) == trusted_consensus_state.next_validators_hash`. This corresponds to [ibc-go's `checkTrustedHeader()`](https://github.com/cosmos/ibc-go/blob/ae27fa5960d523e4476776308ecd0b4cd3377671/modules/light-clients/07-tendermint/update.go#L206-L223) function.

## Description

The light client only keeps the hash of validator sets on chain, as the full set would be too costly to store. Therefore, we rely on the relayer to supply the full validator sets for us when updating a client (in this particular case, [stored in `Header.trusted_validator_set`](https://github.com/cosmos/ibc-rs/blob/20d682fcb0652e27f8dd6e3808e4c7c337e11790/crates/ibc/src/clients/ics07_tendermint/header.rs#L33)).

We should never blindly trust what a relayer gives us though. In this case, we should make sure that the relayer passed us the right validator by checking that its hash is the same as the hash we have stored on chain at `trusted_consensus_state.next_validators_hash`. 

Today, we are missing that check, so effectively we're blindly trusting the relayer for that field.

### LINKED ISSUE #585: `check_header_and_update_state`: missing `Header.height()` vs `Header.trusted_height` check
Problem is [here](https://github.com/cosmos/ibc-rs/blob/ab04536627e306d7182560d80ea3931f83835218/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L487-L511). 

The `tendermint-light-client`'s verifier doesn't ensure that `Header.height() > Header.trusted_height`, so we need to make that check in ibc-rs. In words, the trusted height (representing the height of the header already on chain for which this client update is based on) must be smaller than height of the new header that we're installing.

See [equivalent check](https://github.com/cosmos/ibc-go/blob/dcda04438eb1c51fe70b8e67e3295a77aca2e489/modules/light-clients/07-tendermint/update.go#L84-L89) in ibc-go.

### LINKED ISSUE #589: `check_misbehaviour_and_update_state()`: commit verification might use the wrong `chainId`
In ibc-go, the `chainID`'s `revision_number` is substituted with the header's `revision_number` before validator set commit verification ([here](https://github.com/cosmos/ibc-go/blob/fdd664698d79864f1e00e147f9879e58497b5ef1/modules/light-clients/07-tendermint/misbehaviour_handle.go#L153-L160)). Our current implementation does not do that ([here](https://github.com/cosmos/ibc-rs/blob/5ffffc595778fbde40591abe616e253da7d6b215/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L283)).

### LINKED ISSUE #598: Update client: missing timestamp monotonicity checks to freeze the client
[These checks](https://github.com/cosmos/ibc-go/blob/15cb75017adea8a1dfa08bfa1fea0351f58a82b3/modules/light-clients/07-tendermint/misbehaviour_handle.go#L46-L53) in ibc-go result in the client being frozen when they fail. We currently make them [here](https://github.com/cosmos/ibc-rs/blob/845cf489213bd0d28b7788cdcd79436284f94f23/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L540-L595), but the failure of these checks results in the transaction failing as opposed to the client being frozen.

### LINKED ISSUE #601: `check_header_and_update_state()` always updates the `ClientState.latest_height`
`check_header_and_update_state()` updates it's state [here](https://github.com/cosmos/ibc-rs/blob/845cf489213bd0d28b7788cdcd79436284f94f23/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L598). The new `ClientState` is updated in `with_header()`, which [always sets the height](https://github.com/cosmos/ibc-rs/blob/845cf489213bd0d28b7788cdcd79436284f94f23/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L182-L185) of the new header as `latest_height`. This is incorrect in the case where a header comes in and is not the last header we've seen.

Note: ibc-go [does it properly](https://github.com/cosmos/ibc-go/blob/ae27fa5960d523e4476776308ecd0b4cd3377671/modules/light-clients/07-tendermint/update.go#L146-L148).

plafer:
While implementing the new API, I found several bugs. I implicitly fix those by not re-introducing them in the new API implementation.

**Main issue:**
Closes: #535 

**Bugs fixed:**
Closes: #583 
Closes: #585
Closes: #589
Closes: #598
Closes: #601

## Description

+ [x] tendermint light client methods to implement:
  + [x] `verify_client_message`
  + [x] `check_for_misbehaviour`
  + [x] `update_state_on_misbehaviour`
  + [x] `update_state`
+ [x] `update_client::validate` calls `client_state.verify_client_message`, similar to [here in ibc-go](https://github.com/cosmos/ibc-go/blob/5242ee83eb5f686a367fead20b09803ac481b5ad/modules/core/02-client/keeper/client.go#L53-L66)
+ [x] `update_client::execute` calls `client_state.{check_for_misbehaviour, update_state_on_misbehaviour, update_state}`, similar to [here in ibc-go ](https://github.com/cosmos/ibc-go/blob/5242ee83eb5f686a367fead20b09803ac481b5ad/modules/core/02-client/keeper/client.go#L68-L104)
+ [x] cleanup
  + [x] remove now unused `ClientState::{check_header_and_update_state}` and `ClientState::{check_misbehaviour_and_update_state}`, and all unused `with_*` functions
  + [x] remove now unused `misbehaviour::{validate, execute}`
  + [x] remove checks from `TmMisbehaviour::new()`; this should be a light constructor.
+ [x] Adapt `MsgUpdateClient`
  + [x] `MsgSubmitMisbehaviour` and `MsgUpdateClient` are merged into the "new"  `MsgUpdateClient`
  + [x] Add a `TryFrom<RawMsgSubmitMisbehaviour> for MsgUpdateClient`
  + [x] Update `TryFrom<Any> for MsgEnvelope`
  + [x] Rename `MsgUpdateClient.header` to `MsgUpdateClient.client_message`, and add `update_kind` enum field

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


plafer:
Created #600; I won't implement in this PR but it's a very desirable feature which we should have.

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -1,19 +1,14 @@
...
 use crate::core::ics02_client::msgs::update_client::MsgUpdateClient;
 use crate::events::IbcEvent;
 
 use crate::core::context::ContextError;
 
-use crate::core::ics24_host::path::{ClientConsensusStatePath, ClientStatePath};
-
 use crate::core::{ExecutionContext, ValidationContext};
 
 pub(crate) fn validate<Ctx>(ctx: &Ctx, msg: MsgUpdateClient) -> Result<(), ContextError>
```
ancazamfir, line 14: Question about `validate()` vs `execute()`. If there is a Tx that includes two update client messages what is the order for these function calls:

1. `validate(m1), execute(m1), validate(m2), execute(m2)` or
2. `validate(m1), validate(m2), execute(m1), execute(m2)` 

Since `validate()` doesn't update `ctx`, the latter will not work well with the code the way it is right now.
Looks like we offer:
- `validate()` and `execute()` entypoints which could lead to 2
- `dispatch()` which will give us 1.

In ibc-go there are checks done in `ValidateBasic()`, checks that can be done on the message itself, independent of the context. I think the current equivalent of the ibc-go `ValidateBasic()` is in `ibc-rs` domain type conversion but we don't seem to do some of those checks there.

crates/ibc/src/clients/ics07_tendermint/client_state/update_client.rs
```diff
@@ -0,0 +1,162 @@
...
+                }
+                .to_string(),
+            });
+        }
+
+        // We also need to ensure that the trusted height (representing the
+        // height of the header already on chain for which this client update is
+        // based on) must be smaller than height of the new header that we're
+        // installing.
+        if header.height() <= header.trusted_height {
```
ancazamfir, line 39: move to `TryFrom<RawHeader>`?

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
```diff
@@ -0,0 +1,163 @@
...
+            if duration_since_consensus_state >= self.trusting_period {
+                return Err(Error::ConsensusStateTimestampGteTrustingPeriod {
+                    duration_since_consensus_state,
+                    trusting_period: self.trusting_period,
+                }
+                .into());
+            }
+        }
+
+        // ensure that 2/3 of trusted validators have signed the new header
```
ancazamfir:
```suggestion
        // ensure the client trust_threshold overlap between the validator sets of the trusted and untrusted states.
```

crates/ibc/src/clients/ics07_tendermint/misbehaviour.rs
```diff
@@ -25,37 +24,6 @@ pub struct Misbehaviour {
...
-        verifier
-            .verify_validator_sets(&untrusted_state_1)
-            .into_result()?;
-        verifier
-            .verify_validator_sets(&untrusted_state_2)
-            .into_result()?;
-
-        verifier.verify_commit(&untrusted_state_1).into_result()?;
-        verifier.verify_commit(&untrusted_state_2).into_result()?;
-
```
ancazamfir, line 58: We need these checks (`verify_validator_sets` and `verify_commit`) and I don't see them in the updated code. Maybe add them to `TryFrom<RawHeader>` or `try_from(raw: RawMisbehaviour)` as they don't need a context.

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn check_misbehaviour_header(
+        &self,
+        header: &TmHeader,
+        trusted_consensus_state: &TmConsensusState,
+        current_timestamp: Timestamp,
+    ) -> Result<(), ClientError> {
+        // ensure correctness of the trusted next validator set provided by the relayer
+        check_header_trusted_next_validator_set(header, trusted_consensus_state)?;
+
+        // ensure header timestamp is within trusted period from the trusted consensus state
```
ancazamfir: The code is correct but the comment is not. We could be receiving old headers still within the trusted ~~consensus state~~ period while the trusted consensus state is expired. 
```suggestion
        // ensure trusted consensus state is within trusting period
```

crates/ibc/src/core/ics02_client/msgs/update_client.rs
```diff
@@ -3,32 +3,36 @@
...
+    /// (e.g 2 headers with the same height in Tendermint)
+    Misbehaviour,
+}
 
 /// A type of message that triggers the update of an on-chain (IBC) client with new headers.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MsgUpdateClient {
     pub client_id: ClientId,
-    pub header: Any,
+    pub client_message: Any,
```
ancazamfir, line 37: Why do we keep the "raw" here and not `Msg`. Could convert during `TryFrom<RawMsgUpdateClient>` and `TryFrom<RawMsgSubmitMisbehaviour>`

crates/ibc/src/clients/ics07_tendermint/client_state/update_client.rs
```diff
@@ -0,0 +1,162 @@
...
+                //    the “previous header”
+                {
+                    let maybe_prev_cs = ctx.prev_consensus_state(client_id, &header.height())?;
+
+                    if let Some(prev_cs) = maybe_prev_cs {
+                        // New header timestamp cannot occur *before* the
+                        // previous consensus state's height
+                        let prev_cs = downcast_tm_consensus_state(prev_cs.as_ref())?;
+
+                        if header.signed_header.header().time < prev_cs.timestamp {
```
ancazamfir:
```suggestion
                        if header.signed_header.header().time <= prev_cs.timestamp {
```

crates/ibc/src/clients/ics07_tendermint/client_state/update_client.rs
```diff
@@ -0,0 +1,162 @@
...
+                //    that its timestamp is less than the “next header”
+                if header.height() < self.latest_height() {
+                    let maybe_next_cs = ctx.next_consensus_state(client_id, &header.height())?;
+
+                    if let Some(next_cs) = maybe_next_cs {
+                        // New (untrusted) header timestamp cannot occur *after* next
+                        // consensus state's height
+                        let next_cs = downcast_tm_consensus_state(next_cs.as_ref())?;
+
+                        if header.signed_header.header().time > next_cs.timestamp {
```
ancazamfir:
```suggestion
                        if header.signed_header.header().time >= next_cs.timestamp {
```

crates/ibc/src/clients/ics07_tendermint/client_state/update_client.rs
ancazamfir: We need to think this through a bit. The same code is also in `ibc-go`. Here is where I have doubts:
We are here after we ran light client verification for header agains the trusted consensus state (tcs) and it must be true that:
   - `tcs.height < header.height` and
   - `tcs.timestamp < header.timestamp`
 
so if `tcs == prev_cs` we lost earlier the opportunity to freeze the client if `tcs.timestamp >= header.timestamp`. So I am not sure why we do that monotonicity check during verification (this is done as part of `validate_against_trusted` during `self.verifier.verify()` and not skip it to catch things here.

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
```diff
@@ -0,0 +1,163 @@
...
+
+            self.verifier
+                .validate_against_trusted(
+                    &untrusted_state,
+                    &trusted_state,
+                    &options,
+                    current_timestamp.into_tm_time().unwrap(),
+                )
+                .into_result()?;
+        }
```
ancazamfir: I think we should skip this check for misbehaviour.  In `ibc-go` some of the check are done in `verifyNewHeaderAndVals()`.
I am also not convinced we should keep it fully for update.
This function in `tendermint-rs` does:
-  Ensure the latest trusted header hasn't expired - done before in `check_misbehaviour_header()`
-  Ensure the header isn't from a future time - should be skipped because
  `if h1.h > h2.h && h1.t <= h2.t then ...`  h2 is fla
-  Check that the untrusted block is more recent than the trusted state - should be skipped maybe since this is another form of FLA (we need to think more about this)
-  Check that the chain-id of the untrusted block matches that of the trusted state - done before in `check_misbehaviour_headers_consistency()`

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -32,41 +28,9 @@ where
 
     client_state.confirm_not_frozen()?;
```
ancazamfir, line 29: we should also cover the case where the client is not frozen but still cannot be updated because it is not `active`, ie. the latest trusted consensus state has expired.

ancazamfir (changes requested):
Went over the changes and made some comments. Here are some notes I took during the review, tried to list all checks we do. I think we should have a sync call to go over them as I am not sure myself on some aspects. 

We should document in some place all the checks required for the two cases and the decisions on when/ where they should be done (domain type conversion, validation, execution, ibc-rs vs tendermint-rs, expensive vs cheap checks etc). 

- checks for Header (`TryFrom<RawHeader>`), looks ok. it fails if:
 - `h.signed_header` missing (ibc-rs)
 - `h.signed_header.header` or `h.signed_header.commit` missing (check done in `tendermint-rs`)
 - `h.validator_set` missing
 - `h.trusted_height` missing
 - `h.trusted_validators` missing
 - `h.signed_header.header.chain_id != h.trusted_height.revision_number`


- In `TryFrom<RawMsgUpdateClient>` and `TryFrom<RawMsgSubmitMisbehaviour>` we don't convert to domain types so there is almost no validation. 
**We should consider adding the validation that does not require context and that it is "cheap"**

`validate()`:
 - client must be active (**checking frozen is not enough**)

 - `verify_header()`:
    - convert to Header (see checks above)
    - chain_id/ revision number check again (**could be skipped**)
    - `h.height > h.trusted_height` (**could be done in `TryFrom<RawHeader>`)**
    - `h.trusted_next_validator_set.hash() == trusted_consensus_state.next_validators_hash` (done in `check_header_trusted_next_validator_set()`)
    - `self.verifier.verify` calls (listed here to compare with `verify_misbehaviour`):  
  ```
        ensure_verdict_success!(self.verify_validator_sets(&untrusted));
        ensure_verdict_success!(self.validate_against_trusted(&untrusted, &trusted, options, now)); --> not sure all is required
        ensure_verdict_success!(self.verify_commit_against_trusted(&untrusted, &trusted, options));
        ensure_verdict_success!(self.verify_commit(&untrusted));
 ```
- `verify_misbehaviour()`:
  - converts the misbehaviour msg to domain types, header verification (see checks in Header)
  - retrieves trusted states, ts1 and ts2 must exist
  - `check_misbehaviour_headers_consistency()`
    - `h1` and `h2` must have same `chain_id`
    - `h1.height >= h2.height`
  - `check_misbehaviour_header(h, ts)` for  both`h1` and`h2` (with their trusted states):
    - `check_header_trusted_next_validator_set()` (see above)
     - `ts` not expired
     - `verify_commit_against_trusted()` -> `ts` and `h` have `trust_threshold` overlap
     - `validate_against_trusted()`
        - // Ensure the latest trusted header hasn't expired - **done above already**
        - // Ensure the header isn't from a future time - **should be skipped**
        `if h1.h > h2.h && h1.t <= h2.t then`... h2 is fla
        - // Check that the untrusted block is more recent than the trusted state - **should be skipped ??**
        - // Check that the chain-id of the untrusted block matches that of the trusted state - **done above already**
     - `verify_commit()` - **MISSING**
     - `verify_validator_sets()` - **MISSING**, does the following:
       -    `h.validators.hash == h.signed_header.header.validators_hash` (done in header `ValidateBasic()` for update in ibc-go)
       -  `h.next_validators.hash == h.signed_header.header.next_validators_hash` (**not checked** in `ibc-go`?)
       -  `h.signed_header.header.hash ==  h.signed_header.commit.block_id.hash (done in ValidateBasic() for header in comet, also checks commit vs header height)
       -  `valid_commit` - checks at least one signature

crates/ibc/src/core/ics02_client/handler/update_client.rs
plafer: > If there is a Tx that includes two update client messages what is the order for these function calls

It must always be 1, which is as you mentioned the behavior from dispatch. I added a [clarifying comment](https://github.com/cosmos/ibc-rs/pull/584/commits/478ed8ce12a4f2aba5f04facb49ec7dbfcfda1d3). This is also how Namada processes messages.

> I think the current equivalent of the ibc-go ValidateBasic() is in ibc-rs domain type conversion but we don't seem to do some of those checks there.

The goal is to do all validation in `validate()`, and no validation in `execute()`. If we put validation checks in domain type conversion, then these will be run also in `execute()`, which we're trying to avoid. I'd like us to move away from putting any checks in domain type conversions for this reason.

crates/ibc/src/clients/ics07_tendermint/client_state/update_client.rs
plafer: See relevant comment [here](https://github.com/cosmos/ibc-rs/pull/584#discussion_r1161736624).

Generally, I find putting all the validation in one place makes it easier to reason about the correctness of the code. I think ibc-go split up some validation into `ValidateBasic` for extra validation to run in `CheckTx`. Maybe we could expose an API for that purpose? Note that we'll probably end up making the [message domain types private](https://github.com/cosmos/ibc-rs/issues/324), so it would probably make sense to wait for that change before expose such an API (as the API would work on the raw message type)?

crates/ibc/src/clients/ics07_tendermint/misbehaviour.rs
plafer: We were already doing them in tendermint-rs ([call here](https://github.com/cosmos/ibc-rs/blob/b43507c95f64d850b537e4a13b934e081c4e0685/crates/ibc/src/clients/ics07_tendermint/client_state/update_client.rs#L92-L94)): [`verify_validator_sets`](https://github.com/informalsystems/tendermint-rs/blob/db9cb212974937a38fb64af4f1170c3070b10dc6/light-client-verifier/src/verifier.rs#L267) and [`verify_commit`](https://github.com/informalsystems/tendermint-rs/blob/db9cb212974937a38fb64af4f1170c3070b10dc6/light-client-verifier/src/verifier.rs#L270).

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
plafer: hmm, not sure I follow. The model that I have is

```
(1)                       (2)                        (3) end of trusting period
+--------------------------+--------------------------+
trusted CS             new header 
```

My comment was trying to express: `(1) <= (2) <= (3)`. IIUC, your comment takes a different but equivalent perspective:

```
(1)                       (2)                        (3) new header
+--------------------------+--------------------------+
end of trusting period    trusted CS 
```
That is you look at the trusting period as "going back" from `now` (3) whereas I see it as "going forward" from the trusted CS.

Is this correct? Then I don't understand

> while the trusted consensus state is expired.

what does it mean for a trusted consensus state to be expired?

crates/ibc/src/core/ics02_client/msgs/update_client.rs
plafer: I answered [here](https://github.com/cosmos/ibc-rs/issues/324#issuecomment-1501915686) as this is relevant to the discussion there

crates/ibc/src/core/ics02_client/handler/update_client.rs
plafer: let's do this in #536, which we could do right after this PR

crates/ibc/src/clients/ics07_tendermint/misbehaviour.rs
ancazamfir: But I don't see `verify_header()` being called for misbehaviour (??)

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
ancazamfir:
```
(1)                       (2)                        (3) current_timestamp
+--------------------------+--------------------------+
trusted CS             new header 
```
the comment says `(2) - (1) < trusting_period` but the code does `(3) - (1) < trusting_period`
Note: `(2) - (1) < trusting_period` check is not correct as it may still pass if we update with an older header while trusted cs is expired.

> what does it mean for a trusted consensus state to be expired?

it means it cannot be trusted anymore.

crates/ibc/src/clients/ics07_tendermint/misbehaviour.rs
plafer: blehh, that is very true

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
plafer: So ultimately it seems like we want to

1. remove this block
2. add `verify_validator_sets` and `verify_commit` checks

?

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
plafer: ah, yes, get it now

crates/ibc/src/clients/ics07_tendermint/misbehaviour.rs
```diff
@@ -25,37 +24,6 @@ pub struct Misbehaviour {
 
 impl Misbehaviour {
     pub fn new(client_id: ClientId, header1: Header, header2: Header) -> Result<Self, Error> {
```
Farhad-Shabani: Here we won’t have errors anymore and can go with just returning `Self`

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -177,15 +179,9 @@ impl ClientState {
...
+    pub fn with_header(self, header: TmHeader) -> Result<Self, Error> {
         Ok(ClientState {
-            latest_height: Height::new(
-                self.latest_height.revision_number(),
-                h.signed_header.header.height.into(),
-            )
-            .map_err(|_| Error::InvalidHeaderHeight {
-                height: h.signed_header.header.height.value(),
-            })?,
+            latest_height: max(header.height(), self.latest_height),
```
Farhad-Shabani, line 183: Creating a new `ClientState` without using `new` method skips the `latest_height` checks. Of course, we're confident in the correctness of `latest_height` where `with_header` is called in our implementation (as it passes validations beforehand), but it's still publicly exposed and susceptible to errors!

(On a side note, this is one of the reasons I generally advocate separating new instance creation and validation methods and prefer IBC-go [approach](https://github.com/cosmos/ibc-go/blob/5a67efc496125823f6ca998048ce6c95dcc98427/modules/light-clients/07-tendermint/client_state.go#L112) in this regard)

crates/ibc/src/core/handler.rs
```diff
@@ -6,6 +6,13 @@ use super::{
 };
 
 /// Entrypoint which only performs message validation
+///
+/// If a transaction contains `n` messages `m_1` ... `m_n`, then
+/// they MUST be processed as follows:
+///     validate(m_1), execute(m_1), ..., validate(m_n), execute(m_n)
+/// That is, the state transition of message `i` must be applied before
+/// message `i+1` is validated. This is equivalent to calling
+/// `dispatch()` on each successively.
```
Farhad-Shabani, line 15: As a suggestion, couldn’t we just make `validate()` and `execute()` entrypoints hidden in the first place. Maybe, keeping them under a feature flag and only giving the access if a relevant feature is enabled?

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -342,262 +263,111 @@ impl Ics2ClientState for ClientState {
...
+        let frozen_client_state = self
+            .clone()
+            .with_frozen_height(Height::new(0, 1).unwrap())
+            .into_box();
 
-        Ok(UpdatedState {
-            client_state: client_state.with_header(header.clone())?.into_box(),
-            consensus_state: TmConsensusState::from(header).into_box(),
-        })
+        ctx.store_client_state(ClientStatePath::new(client_id), frozen_client_state)?;
```
Farhad-Shabani, line 367: I see we aimed to follow similar practice as `IBC-go` for this function, but calling a provable storage here is not consistent with the design pattern we took in other places. Anywhere else, when a new **provable state** is obtained upon a process, it is only set within an `execute()` handler. As well, This line doesn't appear to be a client-specific operation. This function can return new states (e.g. using `UpdatedState`), and not leave the storing to users and hope they'll do it right.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -342,262 +263,111 @@ impl Ics2ClientState for ClientState {
...
+                client_id.clone(),
+                new_client_state.latest_height(),
+                ctx.host_height()?,
+            )?;
+
+            ctx.store_consensus_state(
+                ClientConsensusStatePath::new(client_id, &new_client_state.latest_height()),
+                new_consensus_state,
+            )?;
+            ctx.store_client_state(ClientStatePath::new(client_id), new_client_state)?;
```
Farhad-Shabani, line 348: Same [comment](https://github.com/cosmos/ibc-rs/pull/584#discussion_r1163595719) applies here

crates/ibc/src/core/ics02_client/msgs/update_client.rs
```diff
@@ -3,32 +3,36 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum UpdateClientKind {
+    /// this is the typical scenario where a new header is submitted to the client
+    /// to update the client
+    UpdateHeader,
+    /// this is the scenario where misbehaviour is submitted to the client
+    /// (e.g 2 headers with the same height in Tendermint)
+    Misbehaviour,
+}
```
Farhad-Shabani: To my understanding, including `UpdateClientKind` here in ICS02 and its usage as an argument in some of the `ClientState` interfaces, requires it be client agnostic, flexible enough. So, users be able to add/remove their desired variants. This way, we implicitly assume the correct variants of a `client_message` while it is mainly adapted for the CometBFT clients. We neither have such a type within spec nor similar implementation in IBC-go which make me to not be sure about it.

- How about taking a closer look, and consider moving away from using the `UpdateClientKind` enum and `update_kind` fields? Couldn’t we add a `decode_client_message` method for the `ClientState` trait and calling it wherever needed? I am suggesting this for the same reason we added a `decode_client_state` API under the `ValidationContext` (The design required us!)

- A prerequisite for doing so is to merge [`misbehavior`](https://github.com/cosmos/ibc-rs/blob/a7231167efd2edb8fd44aadf4b23583b0158f27e/crates/ibc/src/core/ics02_client/misbehaviour.rs#L9) and [`Header`](https://github.com/cosmos/ibc-rs/blob/a7231167efd2edb8fd44aadf4b23583b0158f27e/crates/ibc/src/core/ics02_client/header.rs#L19) traits and define a new `ClientMessage` interface. Something similar to what already live in [IBC-go](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/core/exported/client.go#L153-L161). This specific merge anyhow looks necessary and should be done along with refactoring others (See [here](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-006-02-client-refactor.md#merge-headermisbehaviour-interface-and-rename-to-clientmessage)).
    
- Also, this way, looking at the [`misbehavior.ValidateBasic()`](https://github.com/cosmos/ibc-go/blob/5a67efc496125823f6ca998048ce6c95dcc98427/modules/light-clients/07-tendermint/misbehaviour.go#L46) and [`header.ValidateBasic()`](https://github.com/cosmos/ibc-go/blob/143e5d85e9d2fc5fc2fc53adc42127a439ee2b79/modules/light-clients/07-tendermint/header.go#L50) in IBC-go, some of `verify_client_message` checks could be moved under a `validate_basic` method.

Farhad-Shabani (commented):
I left some suggestions. About the different checks discussed here in particular, I didn't go into much detail yet, I assumed there might be updates based on your discussion with Anca.
Greatly appreciate you handling this PR!

crates/ibc/src/clients/ics07_tendermint/misbehaviour.rs
plafer: [1cffe09](https://github.com/cosmos/ibc-rs/pull/584/commits/1cffe09dad8b0f167252b089b33bea340a267ed6)

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: > Creating a new `ClientState` without using `new` method skips the `latest_height` checks.

It's not possible to create a `ClientState` object outside the module without calling `new()` because `ClientState` has some private fields. So we know the checks will always run

plafer: > (On a side note, this is one of the reasons I generally advocate separating new instance creation and validation methods and prefer IBC-go [approach](https://github.com/cosmos/ibc-go/blob/5a67efc496125823f6ca998048ce6c95dcc98427/modules/light-clients/07-tendermint/client_state.go#L112) in this regard)

I consider that a code smell in most cases. Invariants about an object field should be checked as you're constructing an object; otherwise in your code you never know if an object you receive is valid. It's just another source of bugs. A similar concept in C++ is [RAII](https://en.cppreference.com/w/cpp/language/raii) (basically objects constructed should always be valid and ready to use without needing to call another function after the constructor).

One scenario where we might want to do checks outside the constructor though is if some invariant checks are very expensive, and we only want to run them in `validate()` (and not `execute()`). Otherwise I think we should stick to the general principle above.

crates/ibc/src/core/handler.rs
plafer: That's a good idea. Created https://github.com/cosmos/ibc-rs/issues/613

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: The justification for this is [here](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-006-02-client-refactor.md#new-clients-may-want-to-do-batch-updates) in the ADR (and links to a [concrete reason](https://github.com/cosmos/ibc-go/issues/284#issuecomment-1005583679) why this is useful for polkadot).

> Anywhere else, when a new provable state is obtained upon a process, it is only set within an execute() handler. 

The model of "1 update client message - 1 client/consensus state update" no longer applies for `MsgUpdateClient` and `MsgSubmitMisbehaviour`. We want the light clients to perform the state updates to enable them to do it on their own schedule.

> This function can return new states (e.g. using UpdatedState),

What do you mean? The function no longer returns `UpdatedState`.

> and not leave the storing to users and hope they'll do it right.

In ibc-rs we have 2 types of "users": light client devs, and chain devs. Light clients are part of the ibc-rs system; if you use a buggy light client, then yes the whole system will be buggy. We happen to be the ones developing the tendermint light client, but the goal is to enable other teams to write light clients as well, and yes we rely on them to write quality light clients. Ultimately, it is always the chain dev's responsibility to audit any code they choose to integrate.

Note that this is not different from any other concern taken care of by light clients; we previously relied on the light client to give us the appropriate `UpdatedState`. Us storing it in `execute()` ourselves is a false sense of security; we still need the light client to give us the proper object to store. All that is changed is that the light client now controls not only the object that will be stored, but also *when* it will be stored.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: https://github.com/cosmos/ibc-rs/pull/584#discussion_r1164231704

crates/ibc/src/core/ics02_client/msgs/update_client.rs
plafer: > including UpdateClientKind here in ICS02 and its usage as an argument in some of the ClientState interfaces, requires it be client agnostic, flexible enough. So, users be able to add/remove their desired variants.

There is no flexibility to be had here. This is an internal type, meant to hold either a `Header` or a `Misbehaviour` message (where the specific "runtime type" depends on the light client); nothing else. The only reason we group them together is that we process both messages in the same way in [`update_client`](https://github.com/cosmos/ibc-rs/blob/b10f4baea722f3942f81fdeb610c84ad4e664f92/crates/ibc/src/core/ics02_client/handler/update_client.rs#L14-L85); this applies to only these 2 cases.

> Couldn’t we add a decode_client_message method for the ClientState trait and calling it wherever needed? I am suggesting this for the same reason we added a decode_client_state API under the ValidationContext (The design required us!)

This is different. We need `decode_client_state()` because by design, we don't want to have to know about all concrete `ClientState`s that exist. So when the relayer gives us a concrete `ClientState` in `CreateClient`, we don't know how to decode it *by design*. We rely on the host for that, since the host knows about all the light clients that it supports.

Here, *we create the domain `MsgUpdateClient`*. The `update_kind` field is just so that we remember what "client message" *we put in there*; and we will only ever put one of 2 things: an `UpdateHeader` or a `Misbehaviour`.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
Farhad-Shabani: > It's not possible to create a ClientState object outside the module without calling new() because ClientState has some private fields. So we know the checks will always run

Right. By the way, good to have a note here so that calling `with_header` (in case needed later anywhere else in the repo, specifically for tests) should be done with caution.

Farhad-Shabani: > I consider that a code smell in most cases. Invariants about an object field should be checked as you're constructing an object; otherwise in your code you never know if an object you receive is valid. It's just another source of bugs. A similar concept in C++ is [RAII](https://en.cppreference.com/w/cpp/language/raii) (basically objects constructed should always be valid and ready to use without needing to call another function after the constructor).

> One scenario where we might want to do checks outside the constructor though is if some invariant checks are very expensive, and we only want to run them in validate() (and not execute()). Otherwise I think we should stick to the general principle above.

I see, that's another way to look at it. Though, already IBC-go is doing that differently. 
Nevertheless, for now, we don't really need to worry about it. We came up with a good context here to keep in mind, in case needed to take a closer look. 🙏🏻

plafer: > Though, already IBC-go is doing that differently.

Note that we mainly use ibc-go to infer what the IBC protocol is (since the standard is out of date/inaccurate, as previously stated). The patterns used in the code and general code style are irrelevant to us. We sometimes borrow some architectural ideas (such as what we did with this PR), but that's about it.

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -67,19 +68,54 @@ pub trait ClientState:
...
+
+    /// Updates and stores as necessary any associated information for an IBC
+    /// client, such as the ClientState and corresponding ConsensusState. Upon
+    /// successful update, a list of consensus heights is returned. It assumes
+    /// the client_message has already been verified.
+    ///
+    /// Post-condition: on success, the return value MUST contain at least one
+    /// height.
+    fn update_state(
+        &self,
```
plafer, line 103: consider making this `self` instead

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -67,19 +68,54 @@ pub trait ClientState:
...
+        ctx: &mut dyn ExecutionContext,
+        client_id: &ClientId,
+        client_message: Any,
+        update_kind: &UpdateClientKind,
+    ) -> Result<Vec<Height>, ClientError>;
+
+    /// update_state_on_misbehaviour should perform appropriate state changes on
+    /// a client state given that misbehaviour has been detected and verified
+    fn update_state_on_misbehaviour(
+        &self,
```
plafer, line 113: `self` here too

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: opened #615

crates/ibc/src/core/ics02_client/msgs/update_client.rs
plafer: changed some names and improved the docstrings, see if you like it [ea2d588](https://github.com/cosmos/ibc-rs/pull/584/commits/ea2d588b84ef630858aea948920d3d758e1665d8)

crates/ibc/src/core/ics02_client/client_state.rs
plafer: This doesn't work since `Self` is a `dyn ClientState`, and so cannot be moved (it must always live behind a reference).

crates/ibc/src/core/ics02_client/msgs/update_client.rs
plafer: and I forgot we had the `Misbehavior` trait; it was not used, so I just removed it [af10978](https://github.com/cosmos/ibc-rs/pull/584/commits/af109786f551da630fb3b6ae9a03944e064c9a7f)

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
plafer: As discussed, for now we will keep the `validate_against_trusted` check here

plafer: Since we want all the calls in `self.verifier.verify()`, I just call that directly.

[d439574](https://github.com/cosmos/ibc-rs/pull/584/commits/d4395749ddd011f8fea0b23cd721dbf0dfe61232)

crates/ibc/src/clients/ics07_tendermint/misbehaviour.rs
plafer: See https://github.com/cosmos/ibc-rs/pull/584#discussion_r1164687443

crates/ibc/src/core/ics02_client/msgs/update_client.rs
Farhad-Shabani: > changed some names and improved the docstrings, see if you like it [ea2d588](https://github.com/cosmos/ibc-rs/pull/584/commits/ea2d588b84ef630858aea948920d3d758e1665d8)

Better now 🙏🏻 

> and I forgot we had the Misbehavior trait; it was not used, so I just removed it [af10978](https://github.com/cosmos/ibc-rs/pull/584/commits/af109786f551da630fb3b6ae9a03944e064c9a7f)

I couldn't find any usage for this [`Header`](https://github.com/cosmos/ibc-rs/blob/b10f4baea722f3942f81fdeb610c84ad4e664f92/crates/ibc/src/clients/ics07_tendermint/header.rs#L84-L92)  implementation, too
And, basically, It's not clear what's the [`Header`](https://github.com/cosmos/ibc-rs/blob/ef52941e29aa2759e2861cdae8a7398b114cc5f2/crates/ibc/src/core/ics02_client/header.rs#L18-L42) trait role. Only serving the `mock` tests

crates/ibc/src/clients/ics07_tendermint/client_state/misbehaviour.rs
ancazamfir: I opened https://github.com/informalsystems/tendermint-rs/issues/1294 to discuss a new misbehaviour header verification API. 
Maybe open a new issue here and link ^ ?

plafer: https://github.com/cosmos/ibc-rs/issues/616

crates/ibc/src/core/ics02_client/msgs/update_client.rs
plafer: Created an issue for it as it is also used in `RelayerContext`, which is currently used in our mock code, which is not test-only. I'd like to investigate moving the whole mock code to `tests`, and not exporting it anymore, as it's not clear what value it currently offers, and couples our test code with non-test code in annoying ways.

https://github.com/cosmos/ibc-rs/issues/617

Farhad-Shabani (approved):
Great work 👌🏻

plafer:
@ancazamfir I will wait for your approval as well before merging

ancazamfir (approved):

PR #586, opened 2023-03-31
ics26_routing: Impl `Protobuf<Any>` for `MsgEnvelope`

kevinji:
Closes: #XXX

## Description

Implement `Protobuf<Any>` for `MsgEnvelope` so that serialization of `MsgEnvelope` types is possible. This matches the already-existing `TryFrom<Any>` implementation.

Using `msg.encode_vec().unwrap()` is safe because `.encode_vec()` will never return an error, as the buffer will always be constructed to be the correct size. I will submit a follow-up PR to ibc-proto to update the return type of `encode_vec()`.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/feature/586-msg-envelope-to-protobuf.md
+++ b/.changelog/unreleased/feature/586-msg-envelope-to-protobuf.md
@@ -1,0 +1,1 @@
+- Impl `Protobuf<Any>` for `MsgEnvelope` ([#586](https://github.com/cosmos/ibc-rs/issues/586))
--- a/crates/ibc/src/core/ics26_routing/msgs.rs
+++ b/crates/ibc/src/core/ics26_routing/msgs.rs
@@ -13,6 +13,7 @@ use crate::core::ics04_channel::msgs::{
     chan_open_init, chan_open_try, recv_packet, timeout, timeout_on_close, ChannelMsg, PacketMsg,
 };
 use crate::core::ics26_routing::error::RouterError;
+use crate::tx_msg::Msg as _;
 use ibc_proto::protobuf::Protobuf;
 
 /// Enumeration of all messages that the local ICS26 module is capable of routing.
@@ -23,6 +24,8 @@ pub enum MsgEnvelope {
     Channel(ChannelMsg),
     Packet(PacketMsg),
 }
+
+impl Protobuf<Any> for MsgEnvelope {}
 
 impl TryFrom<Any> for MsgEnvelope {
     type Error = RouterError;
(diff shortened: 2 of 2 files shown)


kevinji:
Reference to upstream PR: cosmos/ibc-proto-rs#73

Farhad-Shabani:
Thanks for this improvement suggestion.

Note that since each message type provides a [`to_any`](https://github.com/cosmos/ibc-rs/blob/060d80fc66c1de1a82c64ac193f7329ed3f2d326/crates/ibc/src/tx_msg.rs#L23) method, we can extract the domain message from `MsgEnvelop` and apply `domain_msg.to_any()`

However, from what I've seen, the typical conversion flow to work with the IBC-rs implementation is from `Any` to `MsgEnvelop`, and the reverse direction may be rarely used which makes me to think there is no need to retain it in the IBC-rs boundary. Nonetheless, the follow-up PR to ibc-proto may still work and if your project requires that conversion, please let us know what it is used for. we are happy to help.

Farhad-Shabani:
And a quick request. I’d appreciate for such cases if you open a GitHub issue in advance, so we could keep our changelog and plannings consistent. This helps us to search, track and prioritize our issues in a more efficient way. 🙏🏻 🌷

kevinji:
Will do regarding issues! To respond to the use case: I need to serialize `MsgEnvelope` within my code so I can create a client that passes them to our IBC handler, so I thought that this would be the easiest way to do so. If you would prefer not to include this serialization I can just copy-paste what I have here directly into my code.

Farhad-Shabani:
> Will do regarding issues! To respond to the use case: I need to serialize `MsgEnvelope` within my code so I can create a client that passes them to our IBC handler, so I thought that this would be the easiest way to do so. If you would prefer not to include this serialization I can just copy-paste what I have here directly into my code.

In fact, [this](https://github.com/cosmos/ibc-rs/pull/588#issuecomment-1492545375) comment applies here as well. Since we are going towards making our msg structs private (issue #324), we'd prefer only to maintain necessary parts/conversions needed to work with the _IBC-rs implementation_.

kevinji:
I think this should be fine as long as there's some way to call `dispatch` after the change to make things private happens. Is the idea to depend solely on the `ibc-proto`-generated raw message types then?

plafer:
> Is the idea to depend solely on the ibc-proto-generated raw message types then?

That's right. We need to make changes to the domain types from time to time, which is a breaking change every time (which won't work when we're on version 1.x).

kevinji:
Got it, in that case I'll close this PR and await the API changes once the msg structs are private.

Issue #592, opened 2023-03-31
[ICS02] Unclear role of `ClientTypePath` and respectively `store_client_type` method


Farhad-Shabani:
## Problem Statement

1. Looking into the [ICS24](https://github.com/cosmos/ibc/tree/main/spec/core/ics-024-host-requirements#path-space), there should be a provable store for **`ClientType`** in the **`ClientTypePath`** and provable stores are [required](https://github.com/cosmos/ibc/tree/main/spec/core/ics-024-host-requirements#keyvalue-store) to provide methods needed for getting, setting, and deleting a value.

2. According to the [ICS02](https://github.com/cosmos/ibc/tree/main/spec/core/ics-002-client-semantics#create) [**`ClientTypePath`**](https://github.com/cosmos/ibc-rs/blob/d5a44165c880b520bfb410ee693aa44ca8e6be7d/crates/ibc/src/core/ics24_host/path.rs#L50) is utilized during client creation to ensure that the client type is absent in the respective path before proceeding.

3. Our API only offer the [**`store_client_type`**](https://github.com/cosmos/ibc-rs/blob/021005f0dd41cc62e1c7732202ab6eeb17fa4801/crates/ibc/src/core/context.rs#L444-L449) method, which stores a client type without any usage or providing a means to retrieve or remove it. Although we have yet to encounter a need for this interface among consumers like Namada, Basecoin, and Mock, Having it necessitate to add a **`get_client_type`** interface to enable us checking the existence of a client type in the respective path (as specified in the specs).

4. Meanwhile, with the counter mechanism in place, which never fails and increments the client ID, we can be confident that every client creation process results in a new path. This means we don't need to check whether the client already exists in the **`ClientTypePath`** and the whole interaction with the **`ClientTypePath`** becomes unnecessary.

5. I checked the IBC-go implementation in this regard and couldn’t find any **`ClientTypePath`** there. (I have reached out to them for clarification) 

## Proposal
**Considering points 4 & 5, it seems logical to eliminate the **`store_client_type`** interface and move away from `ClientTypePath` even though it is somewhat inconsistent with the specs!**

## Remarks
[...]

PR #602, opened 2023-04-04
Disallow creation of new Tendermint client state instance with a frozen height

### LINKED ISSUE #178: Disallow creation of frozen clients
## Summary of Bug

We're currently not checking in the create-client handler if the client being created is frozen and I don't think it's in the [spec](https://github.com/cosmos/ibc/tree/main/spec/core/ics-002-client-semantics#create) either. Note the spec requires the [Tendermint client-state to be initialized with a `FrozenHeight: null`](https://github.com/cosmos/ibc/blob/main/spec/client/ics-007-tendermint-client/README.md#client-initialisation), but doesn't enforce this in the handler.

____

## For Admin Use

- [ ] Not duplicate issue

Farhad-Shabani:
Closes: #178
______

### PR author checklist:
- [x] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/clients/ics07_tendermint/client_state.rs
+++ b/crates/ibc/src/clients/ics07_tendermint/client_state.rs
@@ -82,7 +82,6 @@ impl ClientState {
         proof_specs: ProofSpecs,
         upgrade_path: Vec<String>,
         allow_update: AllowUpdate,
-        frozen_height: Option<Height>,
     ) -> Result<ClientState, Error> {
         if chain_id.as_str().len() > MaxChainIdLen {
             return Err(Error::ChainIdTooLong {
@@ -172,7 +171,7 @@ impl ClientState {
             proof_specs,
             upgrade_path,
             allow_update,
-            frozen_height,
+            frozen_height: None,
             verifier: ProdVerifier::default(),
         })
     }
@@ -704,9 +703,6 @@ impl Ics2ClientState for ClientState {
         let upgraded_tm_client_state = TmClientState::try_from(upgraded_client_state)?;
         let upgraded_tm_cons_state = TmConsensusState::try_from(upgraded_consensus_state)?;
 
-        // Frozen height is set to None fo the new client state
-        let new_frozen_height = None;
-
         // Construct new client state and consensus state relayer chosen client
         // parameters are ignored. All chain-chosen parameters come from
         // committed client, all client-chosen parameters come from current
@@ -721,7 +717,6 @@ impl Ics2ClientState for ClientState {
             upgraded_tm_client_state.proof_specs,
             upgraded_tm_client_state.upgrade_path,
             self.allow_update,
-            new_frozen_height,
         )?;
 
         // The new consensus state is merely used as a trusted kernel against
@@ -860,9 +855,13 @@ impl TryFrom<RawTmClientState> for ClientState {
         // In `RawClientState`, a `frozen_height` of `0` means "not frozen".
         // See:
         // https://github.com/cosmos/ibc-go/blob/8422d0c4c35ef970539466c5bdec1cd27369bab3/modules/light-clients/07-tendermint/types/client_state.go#L74
-        let frozen_height = raw
+        if raw
             .frozen_height
-            .and_then(|raw_height| raw_height.try_into().ok());
+            .and_then(|h| Height::try_from(h).ok())
+            .is_some()
+        {
+            return Err(Error::FrozenHeightNotAllowed);
+        }
 
         // We use set this deprecated field just so that we can properly convert
         // it back in its raw form
@@ -882,7 +881,6 @@ impl TryFrom<RawTmClientState> for ClientState {
             raw.proof_specs.into(),
             raw.upgrade_path,
             allow_update,
-            frozen_height,
         )?;
 
         Ok(client_state)
@@ -951,16 +949,20 @@ impl From<ClientState> for Any {
 
 #[cfg(test)]
 mod tests {
+    use crate::clients::ics07_tendermint::header::test_util::get_dummy_tendermint_header;
     use crate::prelude::*;
     use crate::Height;
     use core::time::Duration;
     use test_log::test;
 
+    use ibc_proto::google::protobuf::Any;
+    use ibc_proto::ibc::core::client::v1::Height as RawHeight;
     use ibc_proto::ics23::ProofSpec as Ics23ProofSpec;
 
     use crate::clients::ics07_tendermint::client_state::{
         AllowUpdate, ClientState as TmClientState,
     };
+    use crate::clients::ics07_tendermint::error::Error;
     use crate::core::ics02_client::client_state::ClientState;
     use crate::core::ics02_client::trust_threshold::TrustThreshold;
     use crate::core::ics23_commitment::specs::ProofSpecs;
@@ -1133,7 +1135,6 @@ mod tests {
                 p.proof_specs,
                 p.upgrade_path,
                 p.allow_update,
-                None,
             );
 
             assert_eq!(
@@ -1199,7 +1200,6 @@ mod tests {
                 p.proof_specs,
                 p.upgrade_path,
                 p.allow_update,
-                None,
             )
             .unwrap();
             let client_state = match test.setup {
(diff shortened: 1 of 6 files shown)


crates/ibc/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -860,9 +855,13 @@ impl TryFrom<RawTmClientState> for ClientState {
...
         // https://github.com/cosmos/ibc-go/blob/8422d0c4c35ef970539466c5bdec1cd27369bab3/modules/light-clients/07-tendermint/types/client_state.go#L74
-        let frozen_height = raw
+        if raw
             .frozen_height
-            .and_then(|raw_height| raw_height.try_into().ok());
+            .and_then(|h| Height::try_from(h).ok())
+            .is_some()
+        {
+            return Err(Error::FrozenHeightNotAllowed);
+        }
```
plafer, line 864: I don't think this is the right approach; the domain type should be allowed to be created as frozen. What we want in this PR is to disallow a client to be already frozen in the context of the `CreateClient` handler (see other comment as to whether we actually want that).

If we were to want this behavior, I think it should go in `ClientState::initialise()`

plafer (commented):
I don't think we want to disallow the creation of frozen clients (see my earlier comment [on the issue](https://github.com/cosmos/ibc-rs/issues/178#issuecomment-1380757603)). ibc-go and the spec allow for it; I don't see why we should reject it. There might be an obscure use case for it; but in the majority of cases, we would just allow a useless client to be created.

Farhad-Shabani:
> I don't think we want to disallow the creation of frozen clients (see my earlier comment [on the issue](https://github.com/cosmos/ibc-rs/issues/178#issuecomment-1380757603)).

Of course, I saw your comment. Note that just before [this](https://github.com/cosmos/ibc/pull/956) very recent PR in the IBC repo (that was triggered by [our investigation](https://github.com/cosmos/ibc-rs/issues/592) around `ClientTypePath`), the spec [was initializing](https://github.com/cosmos/ibc/commit/ba9c4f82e0c706761e5b4be5a4fbc270357e09e1#diff-41a81331fa27dcef562c77b90b8068c2bf506add8735c89a822eb04659734322L187) a `ClientState` only by the `frozenHeight = null`, which is raised in #136 as another matter to investigate. 

> I don't think this is the right approach; the domain type should be allowed to be created as frozen.

Given the above reason, I disallowed creating a client with a frozen height. Though, from what I see in some other places of the spec (e.g, [here](https://github.com/cosmos/ibc/blob/ed849c7bacf16204e9509f0f0df325391f3ce25c/spec/core/ics-024-host-requirements/README.md#client-state-validation) for validating self client) similar condition - _`FrozenHeight` must be null_ - is still enforcing. Besides, in IBC-go a [new instance](https://github.com/cosmos/ibc-go/blob/5a67efc496125823f6ca998048ce6c95dcc98427/modules/light-clients/07-tendermint/client_state.go#L22-L40) of Tendermint client state is always created with `frozenHeight = clienttypes.ZeroHeight()` not supporting client creation by taking in a frozen height.

Given the above references and considering the primary purpose of having a frozen height which relates to misbehavior detection, I was looking at the `frozen_height` as a state that should be only controlled by the IBC handler. But, nevertheless, If you still think? the allowance is not in conflict with the spec, potentially could be other functionalities for the `frozen_height`, and sometime later it will make sense to create such a client, that’s a different story.

plafer:
Right, so things are a bit confusing on the ibc-go repo, and this is really a technicality.

> Note that just before https://github.com/cosmos/ibc/pull/956 very recent PR in the IBC repo (that was triggered by https://github.com/cosmos/ibc-rs/issues/592 around ClientTypePath), the spec [was initializing](https://github.com/cosmos/ibc/commit/ba9c4f82e0c706761e5b4be5a4fbc270357e09e1#diff-41a81331fa27dcef562c77b90b8068c2bf506add8735c89a822eb04659734322L187) a ClientState only by the frozenHeight = null, which is raised in https://github.com/cosmos/ibc-rs/pull/136 as another matter to investigate.

Unfortunately, I think the spec needs to be treated as a general reference and not a source of truth; ibc-go is really the source of truth.

> Besides, in IBC-go a [new instance](https://github.com/cosmos/ibc-go/blob/5a67efc496125823f6ca998048ce6c95dcc98427/modules/light-clients/07-tendermint/client_state.go#L22-L40) of Tendermint client state is always created with frozenHeight = clienttypes.ZeroHeight() not supporting client creation by taking in a frozen height.

This `NewClientState` function is only called in the upgrade code [here](https://github.com/cosmos/ibc-go/blob/ae27fa5960d523e4476776308ecd0b4cd3377671/modules/light-clients/07-tendermint/upgrade.go#L102-L105). The `CreateClient` handler gets the initial `ClientState` [here](https://github.com/cosmos/ibc-go/blob/ae27fa5960d523e4476776308ecd0b4cd3377671/modules/core/02-client/keeper/client.go#L16-L18), where the `ClientState` was built by the relayer, and could be frozen. I don't see any checks in the `CreateClient` code to ensure that the client is frozen.

> Though, from what I see in some other places of the spec (e.g, [here](https://github.com/cosmos/ibc/blob/ed849c7bacf16204e9509f0f0df325391f3ce25c/spec/core/ics-024-host-requirements/README.md#client-state-validation) for validating self client) similar condition - FrozenHeight must be null - is still enforcing.

`ValidateSelfClient` is called in the connection handshake code, and indeed there it is necessary for the client to not be frozen. This is what I meant by "it will be a useless client"; no connections will be able to use it.

How I see it:
1. Whether or not we allow for a frozen client to be created is an arbitrary choice (either `CreateClient` fails, or the subsequent connection opening handshake fails); I prefer sticking to what ibc-go does, for consistency's sake. For example, if given the same frozen client, an ibc-go chain fails at the connection opening stage, while an ibc-rs chain fails at the `CreateClient` phase, it can cause unnecessary confusion to the developer debugging the situation.
2. We should ensure that we disallow the client to be frozen after an upgrade

Farhad-Shabani:
> Unfortunately, I think the spec needs to be treated as a general reference and not a source of truth; ibc-go is really the source of truth.

> Whether or not we allow for a frozen client to be created is an arbitrary choice (either CreateClient fails, or the subsequent connection opening handshake fails); I prefer sticking to what ibc-go does, for consistency's sake. For example, if given the same frozen client, an ibc-go chain fails at the connection opening stage, while an ibc-rs chain fails at the CreateClient phase, it can cause unnecessary confusion to the developer debugging the situation.

Reached out to the IBC-go team to see what was the reason for allowing the creation of such a client. It is replied that a client is considered to be frozen if it has a **non-zero frozen height** (i.e.[frozen height is used as a boolean](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/light-clients/07-tendermint/client_state.go#L85)) and they will open an issue to add a check for this in the [07-tendermint client state](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/light-clients/07-tendermint/client_state.go#L112) `Validate` to make sure that frozen height is zero during a new client instance creation.

Another thing caught my attention; IBC-go always sets `FrozenHeight` to 1 in [misbehavior cases](https://github.com/cosmos/ibc-go/blob/143e5d85e9d2fc5fc2fc53adc42127a439ee2b79/modules/light-clients/07-tendermint/update.go#L197-L203). Basically, `FrozenHeight` is always 0 or 1, acting like a boolean, and the height number doesn't matter. (They’ll fix this soon, see [here](https://github.com/cosmos/ibc-go/issues/145#issuecomment-1490567966)) Again, makes no sense that users be able to set a predetermined frozen height.

crates/ibc/src/clients/ics07_tendermint/client_state.rs
plafer: given the latest [comment](https://github.com/cosmos/ibc-rs/pull/602#issuecomment-1504050779), let's move forward with this change since ibc-go will also adopt the change. Thus, only this applies:

> If we were to want this behavior, I think it should go in `ClientState::initialise()`

Farhad-Shabani: Not sure if I feel you correctly:

- Initializing an instance of a client happens through [`decode_client_state`](https://github.com/cosmos/ibc-rs/blob/b10f4baea722f3942f81fdeb610c84ad4e664f92/crates/ibc/src/core/ics02_client/handler/create_client.rs#L68) call. And with the current implementation, we won’t end up creating a Tm client with frozen height through that process. Of course, before that, we neither pass a raw ClientState with non-frozen height.
- Plus, from what I see, `intialise` method serves to return a `TmConsensusState`, and considering the [recent spec change](https://github.com/cosmos/ibc/commit/ba9c4f82e0c706761e5b4be5a4fbc270357e09e1#diff-41a81331fa27dcef562c77b90b8068c2bf506add8735c89a822eb04659734322R172), `initialize` would not deal with returning a `ClientState` anymore and mainly handle setting the states.
- And, the needed checks that exist inside `initialize` there in spec, are basically covered through `ClientState::new` call on our side

plafer (approved):

PR #605, opened 2023-04-05
Check if `ClientStatePath` is empty during client creation

### LINKED ISSUE #604: [ICS02] `ClientStatePath` is not checked during client creation
## Problem Statment

- The [ICS02](https://github.com/cosmos/ibc/tree/main/spec/core/ics-002-client-semantics#create) specifies that for handling a client creation request, the process should check if the submitted `ClientState` already exists in the `ClientStatePath`!

    > abortTransactionUnless(provableStore.get(clientStatePath(identifier)) === null)

- Note that an incorrect use of `SetClientState` can result in a `ClientState` being set to the same store as the one that is going to be set by a `MsgCreateClient`.  

## In IBC-go

- There is a [`GetClientState`](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/core/02-client/keeper/keeper.go#L79-L89) method, but **not** used to perform this check. Instead they create an isolated store by [`StoreClient`](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/core/02-client/keeper/client.go#L32) for this purpose.
- This can't be applied to IBC-rs, as we do not access to the store.
 

## Acceptance Criteria
Check `ClientStatePath` to ensure the received `ClientState` via a `MsgCreateClient` does not exist

Farhad-Shabani:
Closes: #604
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/ics02_client/handler/create_client.rs
+++ b/crates/ibc/src/core/ics02_client/handler/create_client.rs
@@ -37,13 +37,19 @@ where
 
     let client_type = client_state.client_type();
 
-    let _client_id = ClientId::new(client_type, id_counter).map_err(|e| {
+    let client_id = ClientId::new(client_type, id_counter).map_err(|e| {
         ClientError::ClientIdentifierConstructor {
             client_type: client_state.client_type(),
             counter: id_counter,
             validation_error: e,
         }
     })?;
+
+    let stored_client_state = ctx.client_state(&client_id);
+    match stored_client_state {
+        Err(ContextError::ClientError(ClientError::ClientStateNotFound { client_id: _ })) => (),
+        _ => return Err(ClientError::ClientStateAlreadyExists { client_id }.into()),
+    };
 
     Ok(())
 }
--- a/crates/ibc/src/core/ics02_client/handler/create_client.rs
+++ b/crates/ibc/src/core/ics02_client/handler/create_client.rs
@@ -45,10 +45,8 @@ where
         }
     })?;
 
-    let stored_client_state = ctx.client_state(&client_id);
-    match stored_client_state {
-        Err(ContextError::ClientError(ClientError::ClientStateNotFound { client_id: _ })) => (),
-        _ => return Err(ClientError::ClientStateAlreadyExists { client_id }.into()),
+    if ctx.client_state(&client_id).is_ok() {
+        return Err(ClientError::ClientStateAlreadyExists { client_id }.into());
     };
 
     Ok(())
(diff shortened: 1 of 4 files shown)


crates/ibc/src/core/ics02_client/handler/create_client.rs
```diff
@@ -37,14 +37,20 @@ where
...
         ClientError::ClientIdentifierConstructor {
             client_type: client_state.client_type(),
             counter: id_counter,
             validation_error: e,
         }
     })?;
 
+    let stored_client_state = ctx.client_state(&client_id);
+    match stored_client_state {
+        Err(ContextError::ClientError(ClientError::ClientStateNotFound { client_id: _ })) => (),
```
plafer: I wouldn't rely on users actually returning `ClientStateNotFound`. As we do [elsewhere](https://github.com/cosmos/ibc-rs/blob/caff666ced4c95ff04f59c798f5e4870e0dcd44a/crates/ibc/src/core/context/timeout.rs#L78), we should just check `is_err()` here. I would wait for my [suggested error system redesign](https://github.com/cosmos/ibc-rs/issues/270#issuecomment-1470734150) to be implemented before we match on specific errors returned from users

Farhad-Shabani: I see, but I'd like to point out that relying on the `is_err()` still leaves the validation step dependent on the users. wouldn't be much better. If `client_state()` fails due to any other reasons such as "broken storage" or "unable to fetch from storage", etc... the validation will wrongly pass, which is why I created issue #607. 
I think, this is a significant one needing to be addressed even before we begin to refactor our errors.

plafer: Note that using `is_err()` achieves the exact [same behavior from ibc-go](https://github.com/cosmos/ibc-go/blob/ae27fa5960d523e4476776308ecd0b4cd3377671/modules/core/04-channel/keeper/packet.go#L210). Being more specific about which error occurred would be good for sure, but I consider using `is_err()` as at least following the same standard as ibc-go.

Farhad-Shabani: Yes, though IBC-go's situation is better than ours. The `nil` is handled internally, users' preferences in error picking do not affect that. Plus, the storage is in a more known and clearer position with respect to the entire IBC implementation, which makes it less susceptible to errors, but it still suffers from this problem.
np, I went with the `is_err` approach for this PR ([c9eba3e](https://github.com/cosmos/ibc-rs/pull/605/commits/c9eba3ec7bc0c051dd8a742526d61edb8d99726f)) Though, the discussed issue remains there.

plafer (approved):
👌

PR #624, opened 2023-04-14
Remove `IbcEventType`

### LINKED ISSUE #623: Fix `IbcEventType`
The current assumption in our events is that each variant of `IbcEvent` has one and only one "event type" (which eventually makes it to the `abci::Event::kind` field). This is not true for `IbcEvent::AppModule`. For example, only in the token transfer app, it can have types `ibc_transfer`, `fungible_token_packet`, [etc](https://github.com/cosmos/ibc-go/blob/d3038f8317bbf4a9f43c578f7b6d4a6a5c4ebda5/modules/apps/transfer/types/events.go#L5-L9).

In order to fix this, we should 
1. Remove `IbcEventType` (since not all event types are known by ibc-core)
2. let each event variant define its own event type

plafer:
Closes: #623

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/breaking-changes/623-remove-ibceventtype.md
+++ b/.changelog/unreleased/breaking-changes/623-remove-ibceventtype.md
@@ -1,0 +1,1 @@
+- Remove `IbcEventType` ([#623](https://github.com/cosmos/ibc-rs/issues/623))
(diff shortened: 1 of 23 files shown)


Farhad-Shabani (approved):
Looks good to me :pray:

PR #626, opened 2023-04-14
Refactor and fix version validation in connection and channel handshakes

### LINKED ISSUE #625: [ICS03][ICS04] Refactor and fix version validation in connection and channel handshakes
## Summary

Part of #233

Found confusing validation codebase around `Version` type, in particular in places we call `contains()` and **unsure if checks are done correctly**. Also, there is a bit higher processing cost within `pick_version` method.
Streamlining and maintaining the code will ensure all validation steps are completed.

Farhad-Shabani:
Closes: #625

(As a side note, this PR streamlined also a few errors, showing that some of our error confusion comes from how certain logic is implemented)
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/ics03_connection/connection.rs
```diff
@@ -331,8 +331,12 @@ impl ConnectionEnd {
     }
 
     /// Getter for the list of versions in this connection end.
-    pub fn versions(&self) -> &[Version] {
-        &self.versions
+    pub fn versions(&self) -> Result<&[Version], ConnectionError> {
+        // Note: With the current implementation, only one version is supported per connection.
+        if self.versions.len() != 1 {
+            return Err(ConnectionError::InvalidVersionLength);
+        }
```
plafer: I don't think ibc-go enforces this, right? If not, neither should we

crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs
```diff
@@ -20,9 +19,7 @@ where
     client_state_of_b_on_a.confirm_not_frozen()?;
 
     if let Some(version) = msg.version {
-        if !ctx_a.get_compatible_versions().contains(&version) {
-            return Err(ConnectionError::VersionNotSupported { version }.into());
-        }
+        version.verify_version_supported(&ctx_a.get_compatible_versions())?;
```
plafer: I find what we had before more readable. I read both as:

+ Before: ensure that the message's version is in the list of "compatible versions"
+ After: ensure that the message's version is supported, based on the list of compatible versions.
    + I'm not sure just by reading this what the requirement for "being supported" is

plafer: Ah, it seems like we were not doing the right check (looking at `verify_version_supported()` and ibc-go's `IsSupportedVersion`

crates/ibc/src/core/ics03_connection/version.rs
```diff
@@ -32,9 +32,33 @@ pub struct Version {
...
+        let maybe_version_supported = supported_versions
+            .iter()
+            .find(|sv| sv.identifier == self.identifier)
+            .ok_or(ConnectionError::VersionNotSupported {
+                version: self.clone(),
+            })?;
+
+        for feature in self.features.iter() {
+            maybe_version_supported.verify_feature_supported(feature.to_string())?;
+        }
```
plafer: ibc-go also [ensures](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/core/03-connection/types/version.go#L80) that there is at least 1 feature.

crates/ibc/src/core/ics03_connection/version.rs
```diff
@@ -32,9 +32,33 @@ pub struct Version {
...
+        Ok(())
+    }
+
     /// Checks whether or not the given feature is supported in this version
-    pub fn is_supported_feature(&self, feature: String) -> bool {
-        self.features.contains(&feature)
+    pub fn verify_feature_supported(&self, feature: String) -> Result<(), ConnectionError> {
+        if feature.trim().is_empty() {
```
plafer: this check is redundant, since for an empty feature, the next check will fail (i.e. `self.features` does not contain `""`)

crates/ibc/src/core/ics03_connection/version.rs
```diff
@@ -95,29 +119,22 @@ pub fn get_compatible_versions() -> Vec<Version> {
...
+    let mut intersection: Vec<Version> = counterparty_versions
+        .iter()
+        .filter(|v| v.verify_version_supported(supported_versions).is_ok())
+        .cloned()
+        .collect();
+
     if intersection.is_empty() {
         return Err(ConnectionError::NoCommonVersion);
     }
+
```
plafer, line 147: ibc-go only returns the features that are contained in both the supported version, and the proposed version [here](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/core/03-connection/types/version.go#L161)

crates/ibc/src/applications/transfer/context.rs
```diff
@@ -134,12 +135,9 @@ pub fn on_chan_open_init_validate(
...
 
-    if !version.is_empty() && version != &Version::new(VERSION.to_string()) {
-        return Err(TokenTransferError::InvalidVersion {
-            expect_version: Version::new(VERSION.to_string()),
-            got_version: version.clone(),
-        });
-    }
+    version
+        .verify_version_supported(Version::new(VERSION.to_string()))
+        .map_err(ContextError::from)?;
```
plafer: This is a different check (i.e. introduces a bug). Previously, an empty version is allowed; now it is not

crates/ibc/src/core/ics03_connection/connection.rs
Farhad-Shabani: - The way a connection is created, neither us nor IBC-go will end up with a connection supporting multiple versions (`Vec<Version>`). Since [`get_compatible_versions`](https://github.com/cosmos/ibc-rs/blob/351efaeeda16621d2321832ecf953a00259067fe/crates/ibc/src/core/ics03_connection/version.rs#L117-L120) returns a vector of size 1, and the [`pick_version`](https://github.com/cosmos/ibc-go/blob/b07741cb95d3de9650564086ec3db55d1b854991/modules/core/03-connection/types/version.go#L146-L174) also matches a single version. 

- Though, already the `get_compatible_versions` API allows hosts to introduce multiple versions. Then, storing a `ConnectionEnd` on host with different versions is possible (bc of [here](https://github.com/cosmos/ibc-rs/blob/351efaeeda16621d2321832ecf953a00259067fe/crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs#L35-L37)), which is not good looking at how a channel is initialized. The `MsgChannelOpenTry::validate()` method [doesn't know](https://github.com/cosmos/ibc-rs/blob/351efaeeda16621d2321832ecf953a00259067fe/crates/ibc/src/core/ics04_channel/handler/chan_open_init.rs#L28) which connection version should be checked for ordering.

I kept this condition as we are not supposed to see multiples upon calling `versions`. We had this before as `InvalidVersionLengthConnection` error. Perhaps, we should drop it after figuring out how to manage various versions or not allowing users for the `Vec<Version>` in `get_compatible_versions`. WDYT?

crates/ibc/src/core/ics03_connection/version.rs
Farhad-Shabani: [fdab426](https://github.com/cosmos/ibc-rs/pull/626/commits/fdab426831e6f160d5646ad19c1349ed26b4021b)

crates/ibc/src/core/ics03_connection/version.rs
Farhad-Shabani: [fdab426](https://github.com/cosmos/ibc-rs/pull/626/commits/fdab426831e6f160d5646ad19c1349ed26b4021b)

crates/ibc/src/core/ics03_connection/version.rs
Farhad-Shabani: ~As far as I see, going through `filter(|v| v.verify_version_supported(supported_versions).is_ok())` ensure this check.~
Got it. I'll fix this

crates/ibc/src/applications/transfer/context.rs
Farhad-Shabani: 👌🏻 [46b9338](https://github.com/cosmos/ibc-rs/pull/626/commits/46b933883fa15efdec087561c29c014c7ef37b2d)

crates/ibc/src/core/ics03_connection/version.rs
Farhad-Shabani: [3d53f13](https://github.com/cosmos/ibc-rs/pull/626/commits/3d53f13a3a73dd23798ba317594f73664564991c)

plafer:
Note: made minor name changes, see if you like [aa91e71](https://github.com/cosmos/ibc-rs/pull/626/commits/aa91e71137ef8bfc15e584f84a7e160d1016daee)

plafer:
Could you also update basecoin to run the integration tests on this branch?

crates/ibc/src/core/ics03_connection/connection.rs
plafer: I looked into it a bit more. The problem is that `versions` has different semantics based on which handshake step it is in:
+ after INIT: `versions` holds the compatible versions
    + This was given by the relayer, or a set of default values
+ after TRY,ACK,CONFIRM: `versions` holds THE version (exactly one) that was chosen by the handshake

Therefore, to accomodate both cases, we currently store a `Vec<Version>`. My takeaways:
1. There's probably a better way to model this using an `enum` which differentates between the 2 cases.
2. For now though, we should update the check to: `if !matches!(self.state, State::Init) && self.versions.len() != 1` (and update the comment because this is hard to read)

plafer: But then the invariant is enforced at "read time", which is not very elegant. A better way is to enforce the invariant in `new()`; when we create the object, we ensure that it is valid, and then we don't have to check this everytime we read (and add a `?`)

Farhad-Shabani: See if this works? [63c50a3](https://github.com/cosmos/ibc-rs/pull/626/commits/63c50a34179e091703194dc9f17c516a53136360)

crates/ibc/src/core/ics03_connection/connection.rs
```diff
@@ -502,9 +509,14 @@ impl State {
         }
     }
 
+    /// Returns true if this connection state matches the expected.
+    pub fn state_matches(&self, expected: &State) -> bool {
+        self == expected
+    }
```
plafer: Just noticed this; this is a reimplementation of the standard `PartialEq::eq`. [3fa02e6](https://github.com/cosmos/ibc-rs/pull/626/commits/3fa02e6808ff51ca162f91b6d9fca90df2a7ca0d)

plafer (approved):
👌

Let's wait for the integration tests to finish and then we can merge https://github.com/informalsystems/basecoin-rs/pull/95

PR #627, opened 2023-04-15
Bump ibc-proto to v0.29.0 and add `memo` field to `PacketData`

### LINKED ISSUE #559: ICS-20: add memo field
We should also support the memo field, added in ibc-go with https://github.com/cosmos/ibc-go/pull/2595.

Currently, if an ibc-go chain sends a packet with a memo, it will fail on the ibc-rs chain (and the user will be able to recover the funds after the timeout). So supporting it in ibc-rs would make interoperability with ibc-go chains better.

kevinji:
Closes: #559

## Description

This PR bumps ibc-proto to v0.29.0, which also introduces the `memo` field that is added to `PacketData` for ICS-20, and tendermint to v0.30.0.

Most of the other changes are due to `Protobuf::encode_vec` no longer returning a `Result`.

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/applications/transfer/events.rs
+++ b/crates/ibc/src/applications/transfer/events.rs
@@ -1,5 +1,5 @@
 use crate::applications::transfer::acknowledgement::TokenTransferAcknowledgement;
-use crate::applications::transfer::{Amount, PrefixedDenom, MODULE_ID_STR};
+use crate::applications::transfer::{Amount, Memo, PrefixedDenom, MODULE_ID_STR};
 use crate::events::ModuleEvent;
 use crate::prelude::*;
 use crate::signer::Signer;
@@ -23,6 +23,7 @@ pub struct RecvEvent {
     pub receiver: Signer,
     pub denom: PrefixedDenom,
     pub amount: Amount,
+    pub memo: Memo,
     pub success: bool,
 }
 
@@ -33,6 +34,7 @@ impl From<RecvEvent> for ModuleEvent {
             receiver,
             denom,
             amount,
+            memo,
             success,
         } = ev;
         Self {
@@ -43,6 +45,7 @@ impl From<RecvEvent> for ModuleEvent {
                 ("receiver", receiver).into(),
                 ("denom", denom).into(),
                 ("amount", amount).into(),
+                ("memo", memo).into(),
                 ("success", success).into(),
             ],
         }
@@ -54,6 +57,7 @@ pub struct AckEvent {
     pub receiver: Signer,
     pub denom: PrefixedDenom,
     pub amount: Amount,
+    pub memo: Memo,
     pub acknowledgement: TokenTransferAcknowledgement,
 }
 
@@ -64,6 +68,7 @@ impl From<AckEvent> for ModuleEvent {
             receiver,
             denom,
             amount,
+            memo,
             acknowledgement,
         } = ev;
         Self {
@@ -74,6 +79,7 @@ impl From<AckEvent> for ModuleEvent {
                 ("receiver", receiver).into(),
                 ("denom", denom).into(),
                 ("amount", amount).into(),
+                ("memo", memo).into(),
                 ("acknowledgement", acknowledgement).into(),
             ],
         }
@@ -103,6 +109,7 @@ pub struct TimeoutEvent {
     pub refund_receiver: Signer,
     pub refund_denom: PrefixedDenom,
     pub refund_amount: Amount,
+    pub memo: Memo,
 }
 
 impl From<TimeoutEvent> for ModuleEvent {
@@ -111,6 +118,7 @@ impl From<TimeoutEvent> for ModuleEvent {
             refund_receiver,
             refund_denom,
             refund_amount,
+            memo,
         } = ev;
         Self {
             kind: EVENT_TYPE_TIMEOUT.to_string(),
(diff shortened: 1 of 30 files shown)


plafer:
We were holding off upgrading to tendermint-rs v0.30 because the `tendermint-abci` crate [no longer supports](https://github.com/informalsystems/tendermint-rs/blob/main/CHANGELOG.md#breaking-changes-1) Tendermint Core v0.34. I am not sure if any project uses `tendermint-abci` though...

Though many projects use `tower-abci` now, and they have a `tendermint-30` ready. @erwanor are you waiting on us to upgrade for you to merge that branch? If so then maybe we take the risk to break projects that do use `tendermint-abci`; we'll have to at some point anyways...

plafer:
Also @kevinji thank you for the PR! Next time could you have one PR/issue please? Reviewing the "memo field" concern is difficult here because the diff is cluttered with tendermint-rs's upgrade related things. It also makes for a cleaner changelog when each issue has its own PR.

But anyways, this is still very appreciated as is 😁

kevinji:
@plafer Hey I'm happy to split this PR up but I didn't find an easy way to do so since we needed to bump `ibc-proto` to pick up the new `memo` field in the raw types. Would it be better to just add the `Memo` type in a separate PR without using it yet, and then start using the `memo` field in the PR that bumps `ibc-proto` and Tendermint?

plafer:
Hmmm I don't have a good enough understanding of the specifics yet to say; let's keep it as is for this PR, and I'll see if I can suggest how I would have split it up after reviewing (for future reference)

romac:
FYI if you end up bumping ibc-proto and tendermint-rs, I just released ibc-proto 0.30.0 which brings with it tendermint-proto 0.31.1.

kevinji:
I just amended the PR so ibc-proto is now bumped to v0.30.

crates/ibc/src/applications/transfer/memo.rs
```diff
@@ -0,0 +1,47 @@
...
+        scale_info::TypeInfo
+    )
+)]
+#[cfg_attr(
+    feature = "borsh",
+    derive(borsh::BorshSerialize, borsh::BorshDeserialize)
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Memo(String);
```
plafer, line 21: We need to add the memo to the events as well:

+ `TransferEvent` ([here](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/apps/transfer/keeper/msg_server.go#L52) in ibc-go)
+ `RecvEvent` ([here](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/apps/transfer/ibc_module.go#L203) in ibc-go)
+ `AckEvent` ([here](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/apps/transfer/ibc_module.go#L250) in ibc-go)
+ `TimeoutEvent` ([here](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/apps/transfer/ibc_module.go#L297) in ibc-go)

plafer: Would it be possible to split the `Memo` part into a new PR, where we also upgrade ibc-proto-rs to v0.28.0 (first version with the memo field), and no tendermint-rs upgrade?

Essentially I don't think we should depend on tendermint v0.30+ until tower-abci releases a version which supports it; otherwise, it forces our users to use an unreleased version of tower-abci (probably their branch `tendermint-30`). In the meantime though we should still be able to support the memo field in ics 20

erwanor: hey @plafer, just cut a release for `tower-abci` with support for `0.30` - sorry for the delay, this slipped through, if I forget again next time feel free to bump with a comment in the PR for max visibility :)

plafer: Awesome thanks @erwanor! In this case, @kevinji , I think what's left for this PR is:
+ downgrade to use tendermint-rs v0.30 (to be compatible with tower-abci v0.6.0)
+ downgrade to use ibc-proto v0.29 (highest version which depends on tendermint-proto v0.30)
+ Add the memo to the events as mentioned above

Let me know if you need any help with any of this

kevinji:
Just responded to review comments, sorry for the delay!

crates/ibc/src/applications/transfer/events.rs
```diff
@@ -23,6 +23,7 @@ pub struct RecvEvent {
     pub receiver: Signer,
     pub denom: PrefixedDenom,
     pub amount: Amount,
+    pub memo: Memo,
```
plafer, line 26: `TimeoutEvent` will also need the same `memo` field (see [ibc-go](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/apps/transfer/ibc_module.go#L297))

plafer (changes requested):
Looks good, except we're only missing the `TimeoutEvent::memo` field!

kevinji:
Just pushed a new change adding that field.

plafer (approved):
🙏

PR #642, opened 2023-04-20
Remove `store_client_type` interface

### LINKED ISSUE #592: [ICS02] Unclear role of `ClientTypePath` and respectively `store_client_type` method

Farhad-Shabani:
Closes: #592
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/breaking-changes/592-remove-store-client-type.md
+++ b/.changelog/unreleased/breaking-changes/592-remove-store-client-type.md
@@ -1,0 +1,2 @@
+- Remove `store_client_type` interface as it is not included in the IBC spec anymore.
+  ([#592](https://github.com/cosmos/ibc-rs/issues/592))
--- a/crates/ibc/src/core/context.rs
+++ b/crates/ibc/src/core/context.rs
@@ -32,7 +32,6 @@ use core::time::Duration;
 use ibc_proto::google::protobuf::Any;
 
 use crate::core::ics02_client::client_state::ClientState;
-use crate::core::ics02_client::client_type::ClientType;
 use crate::core::ics02_client::consensus_state::ConsensusState;
 use crate::core::ics03_connection::connection::ConnectionEnd;
 use crate::core::ics03_connection::version::{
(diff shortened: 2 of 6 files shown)


plafer (approved):
👌

PR #661, opened 2023-05-05
Separate handler methods from context APIs as free functions 

### LINKED ISSUE #539: [Context] Separate `validate()` and `execute()` handlers for better modularity and API coherence
## Problem Statement

The `ValidationContext` and `ExecutionContext` traits now include default implementations for their `validate()` and `execute()` methods. However, these methods use functions that are only public to the crate. This makes it nearly impossible for someone to use a different implementation and renders keeping them as API meaningless.

Moreover, looking at the operations they do, it does not make sense to keep these two (validate/execute) along with the others. They deal with the **handler** functionality and route an incoming message to the appropriate validation/execution handler, while others mainly interact with **storage** for retrieving or setting a state. It would be better to keep these two roles separate to achieve more modularity.

### LINKED ISSUE #613: Put top-level `validate()` and `execute()` entrypoints under a feature flag
As a suggestion, couldn’t we just make `validate()` and `execute()` entrypoints hidden in the first place. Maybe, keeping them under a feature flag and only giving the access if a relevant feature is enabled?

_Originally posted by @Farhad-Shabani in https://github.com/cosmos/ibc-rs/pull/584#discussion_r1163594735_

Farhad-Shabani:
Closes: #539
Closes: #613

Basecoin CI check: [3dc0c23](https://github.com/informalsystems/basecoin-rs/pull/97/commits/3dc0c2310d02571076635079d17f0d9e3d603055)
______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/entrypoint.rs
```diff
@@ -0,0 +1,42 @@
...
+/// Entrypoint which only performs message validation
+///
+/// If a transaction contains `n` messages `m_1` ... `m_n`, then
+/// they MUST be processed as follows:
+///     validate(m_1), execute(m_1), ..., validate(m_n), execute(m_n)
+/// That is, the state transition of message `i` must be applied before
+/// message `i+1` is validated. This is equivalent to calling
+/// `dispatch()` on each successively.
+#[cfg(feature = "val-exec-entry")]
+pub fn validate<Ctx>(ctx: &Ctx, message: Any) -> Result<(), RouterError>
```
plafer: I think we should have all entry points use `MsgEnvelope`. I don't like `Any` because it hides the real type, is arbitrary (why `Any` as opposed to simply using the actual `MsgEnvelope` that we expect?), and if users actually have a `MsgEnvelope` in an `Any`, they only need to add 1 line of code before calling our entrypoints.

crates/ibc/Cargo.toml
```diff
@@ -50,6 +50,9 @@ upgrade_client = []
 mocks = ["tendermint-testgen", "tendermint/clock", "cfg-if", "parking_lot"]
 mocks-no-std = ["cfg-if"]
 
+# This feature grants access to additional `validate` and `execute` entrypoints.
+val-exec-entry = []
```
plafer: Why have a feature? I find it has many downsides
1. it's a crutch against proper documentation. All this does currently is effectively "hide" 2 functions, with no binary size improvements that conditional compilation is typically used for. Don't you think there's a way to properly document these functions such that it's clear which one a given user should use? 
2. Our docs completely hide them from the documentation. This makes discoverability of these entrypoints more difficult. Some projects have a "X feature flag required" notes in their docs, but we don't (and I currently don't know how to do it).

crates/ibc/src/core/mod.rs
```diff
@@ -33,12 +33,17 @@ pub mod events;
 pub mod timestamp;
 
 mod context;
+mod entrypoint;
 mod handler;
```
plafer: `handler` and `entrypoint` are basically the same thing; I don't think we need both

crates/ibc/Cargo.toml
Farhad-Shabani: The goal was to not expose them to the public, b/c of issue discussed in #584 (Since they may be misapplied and considered the rare use case, didn't want to be in immediate access). Nevertheless, it's been fixed [4d1ca0a](https://github.com/cosmos/ibc-rs/pull/661/commits/4d1ca0a825b2f37be264c55e2cc9caf5bead0b20)

crates/ibc/src/core/mod.rs
Farhad-Shabani: [4d1ca0a](https://github.com/cosmos/ibc-rs/pull/661/commits/4d1ca0a825b2f37be264c55e2cc9caf5bead0b20)

crates/ibc/src/core/entrypoint.rs
Farhad-Shabani: [4d1ca0a](https://github.com/cosmos/ibc-rs/pull/661/commits/4d1ca0a825b2f37be264c55e2cc9caf5bead0b20)

plafer (approved):
👌

Issue #420, opened 2023-02-13
Provide a "upgrade client proposal" handler for Tendermint chains


Farhad-Shabani:
## Problem Definition

The work done so far and covered within [ADR06](https://github.com/cosmos/ibc-rs/blob/main/docs/architecture/adr-006-upgrade-client-implementation.md) around upgrading a counterparty client mainly reflects how to handle submitted `MsgUpgradeClient` by the relayer and substitute the upgraded client and consensus states on the counterparty chain. 
But, there is another side to this process (see [ibc-go](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/core/02-client/keeper/proposal.go#L71-L102)) and it is where the newly proposed client and consensus states should be stored in the _upgrade path_ (related to #141) of the upgrading chain. 
We haven't clarified yet how the incoming chain-level upgrade proposal should be handled within the IBC-rs core and a specific implementation for Tendermint chains seems to be necessary.

Farhad-Shabani:
### Updates 

- We cannot assume about how hosts may reach a consensus on upgrading a client, which could be via a governance model, multi-sig, etc. Accordingly, we can’t write a generic `UpgradeProposal` handler.
- The minimum we should do is expose getter/setter APIs under our context traits, so hosts can store and retrieve upgraded client/consensus states in/from `UpgradeClientPath`
- For Comet-based chains, we might be able to code a very similar operation to IBC-go and maintain it within IBC-rs as some ready-to-use functions. But we'd have to add a minimal gov module to basecoin-rs [first](https://github.com/informalsystems/basecoin-rs/issues/100). (**Blocked**)

Issue #296, opened 2022-12-13
Enable light clients to specify the methods they need the host to implement


plafer:
Currently, `check_header_and_update_state()` [takes a `ClientReader`](https://github.com/cosmos/ibc-rs/blob/7d128f7d9855b2735bbfd340417ca09ded174ced/crates/ibc/src/core/ics02_client/client_state.rs#L84) as a context object. Hence, any methods from the context that we needed in the tendermint light  client implementation, we added to `ClientReader` (and more recently, `ValidationContext`). This is problematic because:

1. Light clients other than tendermint might need a different set of methods from the host
2. Hosts may not want to run have a tendermint light client, and it might not always be clear how to implement the methods that were required only by the tendermint light client (see [this](https://github.com/cosmos/ibc-rs/issues/17#issuecomment-1261306040)).

We should think of different ways to address this problem; namely to allow light clients to impose the set of methods that they need the host to implement.

plafer:
Similarly, the packet verification functions [take a `ChannelReader`](https://github.com/cosmos/ibc-rs/blob/7f62c1b6fe0d123defe8bda7a707bed2edd65c70/crates/ibc/src/core/ics02_client/client_state.rs#L182) (soon to be `ValidationContext`). `ChannelReader` is the wrong trait here; the methods it provides are not necessarily the right ones for all light clients.

plafer:
I made this work in nightly rust (see [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=56d2d7bae2f7363498abca0f9508f8a5)). I left it as a future exercise to make it work in stable rust.

Notice however that this solution requires `ClientState` to be generic (i.e. not trait object safe). This might be a motivation to go back on our decision to use dynamic dispatch for light client types in [ADR 4](https://github.com/cosmos/ibc-rs/blob/8d5178d0cdc0329fa89f5050807b8aa0290f64a2/docs/architecture/adr-004-light-client-crates-extraction.md), and instead switch to static dispatch. This is a major change in our API, but we might be forced to do it to support this feature.

```rust
#![feature(return_position_impl_trait_in_trait)]

////////////////////////////////////////
// ibc-rs crate
////////////////////////////////////////
pub mod ibc_rs {
    /// the ICS-2 ClientState trait
    pub trait ClientState<ClientCtx> {
        fn check_header(&self, ctx: &ClientCtx);
    }

    /// Our ValidationContext, now parameterized by a `ClientCtx`,
[...]

PR #673, opened 2023-05-10
Fix upgrade-client issues and refactor relevant tests

### LINKED ISSUE #385: Validate "upgrade_client" implementation by adding unit and integration tests
## Improvement Summary

Validate "upgrade_client" implementation (ADR-06) by adding unit and integration tests

## Proposal

- [ ] Write all the necessary unit tests and correspond them with `IBC-go`
- [ ] Perform integration and hot tests using `basecoin-rs` and `Hermes`
- [ ] Make sure that we are building the right upgrade paths

### LINKED ISSUE #671: [ICS07] try_from `RawClientState` conversion should exclude `ClientState::new` checks
## Problem Statement

It is possible that `MsgUpgradeClient` contains a valid `ClientState`, but it won't necessarily pass checks within [`ClientState::new()`](https://github.com/cosmos/ibc-rs/blob/ee16dc2da38808e3c68c9bdf9fef42f69b3b627d/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L660-L670) when parsed from `Any`. E.g. `trust_level` might be zero, see how [Hermes](https://github.com/informalsystems/hermes/blob/a285c015c604819b213c1582d901e5ef140d3f47/crates/relayer/src/upgrade_chain.rs#L110-L114) creates the `ClientState` for an upgrade proposal.

The ClientState validity checks for the upgrade_client  must be done only after a [`new_client_state`](https://github.com/cosmos/ibc-rs/blob/ee16dc2da38808e3c68c9bdf9fef42f69b3b627d/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L479-L489) is built within the `update_state_with_upgrade_client` process

### LINKED ISSUE #672: [ICS07] Client/ConsensusState bytes for upgrade_client validation should be derived from prost
## Bug Summary

`verify_upgrade_client` method encodes (e.g. [here](https://github.com/cosmos/ibc-rs/blob/ee16dc2da38808e3c68c9bdf9fef42f69b3b627d/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L428-L429)) upgraded_client/cons_state incorrectly.

## Proposal

We should instead use `prost::Message`:
```Rust
        let mut client_state_value = Vec::new();
        upgraded_client_state
            .encode(&mut client_state_value)
            .map_err(ClientError::Encode)?;
```

## Version
v0.40.0

Farhad-Shabani:
Closes: #385
Closes: #671
Closes: #672

## Description
Check out Basecoin-rs [PR99](https://github.com/informalsystems/basecoin-rs/pull/99) for the new integration test covering client upgrades.

**Note:**

- With the current implementation, all comet-based chains can upgrade their clients on hosts using IBC-rs, but hosts for handling an upgrade proposal must figure out the implementation suits their architecture, but can use the newly defined APIs here for setting/getting upgraded_client/cons_state.

- The new APIs haven't been implemented for `MockContext` yet because both it and its storage will be undergoing changes. Additionally, the methods aren't callable anywhere for testing purposes due to the lack of an upgrade proposal handler.

______

### PR author checklist:
- [x] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/context.rs
+++ b/crates/ibc/src/core/context.rs
@@ -25,7 +25,6 @@ use crate::core::ics24_host::identifier::ConnectionId;
 use crate::core::ics24_host::path::{
     AckPath, ChannelEndPath, ClientConnectionPath, ClientConsensusStatePath, ClientStatePath,
     CommitmentPath, ConnectionPath, ReceiptPath, SeqAckPath, SeqRecvPath, SeqSendPath,
-    UpgradeClientPath,
 };
 use crate::core::router::Router;
 use crate::core::timestamp::Timestamp;
@@ -116,12 +115,6 @@ pub trait ValidationContext: Router {
     /// Returns the ClientState for the given identifier `client_id`.
     fn client_state(&self, client_id: &ClientId) -> Result<Box<dyn ClientState>, ContextError>;
 
-    /// Returns the upgraded `ClientState` for the height that upgrade is planned.
-    fn upgraded_client_state(
-        &self,
-        client_upgrade_path: &UpgradeClientPath,
-    ) -> Result<Box<dyn ClientState>, ContextError>;
-
     /// Tries to decode the given `client_state` into a concrete light client state.
     fn decode_client_state(&self, client_state: Any) -> Result<Box<dyn ClientState>, ContextError>;
 
@@ -147,12 +140,6 @@ pub trait ValidationContext: Router {
         client_id: &ClientId,
         height: &Height,
     ) -> Result<Option<Box<dyn ConsensusState>>, ContextError>;
-
-    /// Returns the upgraded `ConsensusState` for the height that upgrade is planned.
-    fn upgraded_consensus_state(
-        &self,
-        upgraded_cons_state_path: &UpgradeClientPath,
-    ) -> Result<Box<dyn ConsensusState>, ContextError>;
 
     /// Returns the current height of the local chain.
     fn host_height(&self) -> Result<Height, ContextError>;
@@ -285,25 +272,11 @@ pub trait ExecutionContext: ValidationContext {
         client_state: Box<dyn ClientState>,
     ) -> Result<(), ContextError>;
 
-    /// Sets the upgraded `ClientState` in the upgrade path
-    fn store_upgraded_client_state(
-        &mut self,
-        upgraded_client_state_path: UpgradeClientPath,
-        upgraded_client_state: Box<dyn ClientState>,
-    ) -> Result<(), ContextError>;
-
     /// Called upon successful client creation and update
     fn store_consensus_state(
         &mut self,
         consensus_state_path: ClientConsensusStatePath,
         consensus_state: Box<dyn ConsensusState>,
-    ) -> Result<(), ContextError>;
-
-    /// Sets the upgraded `ConsensusState` in the upgrade path
-    fn store_upgraded_consensus_state(
-        &mut self,
-        upgraded_cons_state_path: UpgradeClientPath,
-        upgraded_consensus_state: Box<dyn ConsensusState>,
     ) -> Result<(), ContextError>;
 
     /// Called upon client creation.
--- a/crates/ibc/src/core/context.rs
+++ b/crates/ibc/src/core/context.rs
@@ -25,6 +25,7 @@ use crate::core::ics24_host::identifier::ConnectionId;
 use crate::core::ics24_host::path::{
     AckPath, ChannelEndPath, ClientConnectionPath, ClientConsensusStatePath, ClientStatePath,
     CommitmentPath, ConnectionPath, ReceiptPath, SeqAckPath, SeqRecvPath, SeqSendPath,
+    UpgradeClientPath,
 };
 use crate::core::router::Router;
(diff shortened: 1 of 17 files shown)


crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
```diff
@@ -24,13 +21,6 @@ where
 
     ctx.validate_message_signer(&signer)?;
 
-    // Temporary has been disabled until we have a better understanding of some design implications
```
plafer, line 27: I can't find where ibc-go looks if the `old_client_state` is expired; can you point me to it? 

(line 49, GH won't let me put a comment there since it wasn't modified in this PR)

crates/ibc/src/core/context.rs
```diff
@@ -272,13 +285,27 @@ pub trait ExecutionContext: ValidationContext {
         client_state: Box<dyn ClientState>,
     ) -> Result<(), ContextError>;
 
+    /// Sets the upgraded `ClientState` in the upgrade path
+    fn store_upgraded_client_state(
```
plafer: All new upgrade-related methods on `ValidationContext` and `ExecutionContext` are never used. Are they necessary? ibc-go doesn't seem to have any special methods to store an "upgraded" client/consensus state vs a "normal" one

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
Farhad-Shabani: ofc, [Here](https://github.com/cosmos/ibc-go/blob/c3b12b11f1f60e0c7ba45ccc4c0cd8ae9365e538/modules/core/02-client/keeper/client.go#L125-L127) upon calling the `GetClientStatus` method, it goes through the [status check](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/modules/core/02-client/keeper/keeper.go#L414)

plafer: Ah right. The "client expired" check is a [tendermint one though](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/light-clients/07-tendermint/client_state.go#L97-L98), so we should have it in the tendermint client instead

crates/ibc/src/core/context.rs
Farhad-Shabani: I believe we need them. Ofc, not used yet, but after fixing [this issue](https://github.com/informalsystems/basecoin-rs/issues/100) for basecoin, it will be called for handling upgrade proposals. It's the minimal interface for hosts to interact with the provable stores at the `ClientUpgradePath`, or otherwise they should take care of that in their boundary.
And [here](https://github.com/cosmos/ibc-go/blob/9c6edc4bb828e994645d32108bc4610fe065eb71/modules/core/02-client/types/expected_keepers.go#L17-L26) is the relevant keepers in ibc-go

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
Farhad-Shabani: I see your point. Just note that, with the current design, having `expired` and even `confirm_not_frozen` as client-wide interfaces, I suggest we mend this while going through #536 and wait for changes from #296
Particularly given that, the `expired` needs a `host_timestamp` from context for calculation, which is not available within the `verify_upgrade_client`

crates/ibc/src/core/context.rs
plafer: `ValidationContext` and `ExecutionContext` are the contexts specific to handling all messages in `MsgEnvelope`. This is something else. We'd need some kind of [`HandleUpgradeProposal`](https://github.com/cosmos/ibc-go/blob/295bab01f633633e0d6136ba9b974c0e248edbe5/modules/core/02-client/keeper/proposal.go#L73) contexts (validation & execution), which chains can use to do upgrade proposals. However, I believe you had trouble creating such an abstraction?

crates/ibc/src/core/ics02_client/handler/upgrade_client.rs
plafer: Makes sense

crates/ibc/src/core/context.rs
Farhad-Shabani: Hmm, my perspective was a bit different. As we surfaced all the APIs that interact with provable stores within the Validation/Execution contexts, I've placed them there. 
In terms of the abstraction, to have the ability to go over the correctness and functionality of that handler in one go (rather than trial & error), I postponed that for the moment until I add a very simple gov module to the basecoin-rs or maybe even something as basic as possible.
Alright. If so, let me drop them for now, and add them back while finishing #420

Farhad-Shabani: [736a7d2](https://github.com/cosmos/ibc-rs/pull/673/commits/736a7d2e770d46d49e23f3f708fa3340c247c0f7)

plafer (approved):
👌

PR #683, opened 2023-05-17
Static `ClientState` API

### LINKED ISSUE #296: Enable light clients to specify the methods they need the host to implement

### LINKED ISSUE #534: [ICS02] Examine removing `dyn ValidationContext` type from `check_header_and_update_state()` and `check_misbehaviour_and_update_state()`
## Summary
Part of #25

Examine the possibility of moving away from passing `dyn ValidationContext` type at `check_header_and_update_state()` and `check_misbehaviour_and_update_state()`

## Description

This requires us to pass a `Sized` object and restricts us from exploring more generic ideas (Like incorporating associated or generic types within `ValidtionContext` / `Module` / `Router` traits or using blanket implementation to get the needed interface available for objects that implement `ClientState`/ `Module` traits) 
As well, it does not seem efficient to pass all the context while needed data comes in via one or two methods.

### LINKED ISSUE #614: [ICS07] Investigate updating the store in `ClientState::initialize()` instead of `create_client::execute()`
## Summary

Looking at the [spec](https://github.com/cosmos/ibc/tree/main/spec/client/ics-007-tendermint-client#client-initialisation), we are missing the following check & setting upon initializing a Tendermint client state:

- assert(clientState.trustLevel >= 1/3 && clientState.trustLevel <= 1)
- provableStore.set("clients/{identifier}/clientState", clientState)

plafer:
Closes: #296
Closes: #534
Closes: #614

## Description

- [x] Finish implementation
- [x] Mock: put context implementations in separate files
- [x] Implement derive macros to reduce boilerplate for users
  - [x] `ClientState`
  - [x] `ConsensusState`
- [ ] Integrate with basecoin-rs
- [x] Document why we need `ClientExecutionContext`
- [x] Tendermint LC: Add `CommonContext` as a supertrait to `Validation/ExecutionContext`
  - with method `consensus_state()`. All other `ValidationContext` methods are not needed in execution.
- [x] Write ADR
- [x] Document all new types and methods

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/context.rs
```diff
@@ -114,34 +93,46 @@ impl std::error::Error for RouterError {
 ///
 /// Trait used for the top-level [`validate`](crate::core::validate)
 pub trait ValidationContext: Router {
+    type ClientValidationContext;
+    type E: ClientExecutionContext;
```
Farhad-Shabani, line 97: Consider a scenario where a host handles multiple light clients, each wanting to introduce their own validation and execution context methods. Here we are limited to assigning just one struct/enum context that should encompass the needed methods by all the clients. Wouldn’t this lead to the unnecessary exposure of methods from one client (say client A) to another (client B)? 
Also, e.g. wherever there is an argument of type `self::AnyClientState` for a method of Client A, it potentially can accepts any other client variants defined under the `AnyClientState`.

crates/ibc/src/core/ics02_client/client_state.rs
```diff
@@ -51,15 +52,71 @@ pub trait ClientState:
...
+///     ClientValidationContext: MyValidationContext,
+/// {
+///   // `MyValidationContext` methods available
+/// }
+///
+/// trait MyValidationContext {
+///   // My Context methods
+/// }
+/// ```
+pub trait ClientStateValidation<ClientValidationContext> {
```
Farhad-Shabani, line 120: Compared to the previous implementation where `ctx` was bound by [`ValidationContext`](https://github.com/cosmos/ibc-rs/blob/5d1855b2593b1235060b231f2abb4d9407a850cd/crates/ibc/src/core/ics02_client/client_state.rs#L72), the current approach gives developers more flexibility. They can now pass any struct that implements `ValidationContext` or other types as the client validation context. Not sure if we could effectively restrain the light client capabilities here (I mean the [3rd goal](https://github.com/cosmos/ibc-rs/pull/708/files#diff-f0fc94f4a7703342248727cb915b38b67295163eaf33e0696401cde405c8ec23R12-R14), stated in ADR07). 
Now it is the responsibility of light client developers to ensure proper restriction of generics, similar to what is done for `TmClientState` e.g. [here](https://github.com/cosmos/ibc-rs/blob/617f642aec339e077f6584fc4be99b19784cda11/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L437) with `TmValidationContext` bound or [here](https://github.com/cosmos/ibc-rs/blob/617f642aec339e077f6584fc4be99b19784cda11/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L492-L493) for `Any*` types. Ideally, this should have been enforced at the `ClientState` trait level, though I'm not sure if it is possible with all the constraints we face.

crates/ibc/src/core/context.rs
```diff
@@ -114,34 +93,46 @@ impl std::error::Error for RouterError {
...
+    type AnyConsensusState: ConsensusState;
+    type AnyClientState: ClientState<Self::ClientValidationContext, Self::E>;
+
+    fn get_client_validation_context(&self) -> &Self::ClientValidationContext;
+    fn get_client_execution_context(&mut self) -> &mut Self::E;
```
Farhad-Shabani: Reminder to move this to the `ExecutionContext`.

crates/ibc/src/core/ics02_client/consensus_state.rs
```diff
@@ -1,87 +1,32 @@
...
-        fn eq_consensus_state(&self, other: &dyn ConsensusState) -> bool {
-            other
-                .as_any()
-                .downcast_ref::<CS>()
-                .map_or(false, |h| self == h)
-        }
-    }
+    /// Serializes the `ConsensusState`. This is expected to be implemented as
+    /// first converting to the raw type (i.e. the protobuf definition), and then
+    /// serializing that.
```
Farhad-Shabani, line 27: It was nice if we could enforce this requirement through the trait bounds, so any implementation of the `ConsensusState` would observe it. Particularly given that there is only a single valid way of encoding to the vec.

crates/ibc/src/core/ics02_client/consensus_state.rs
```diff
@@ -1,87 +1,32 @@
...
+    ///
+    /// Note that the `Protobuf` trait in `tendermint-proto` provides convenience methods
+    /// to do this automatically.
+    fn encode_vec(&self) -> Result<Vec<u8>, tendermint_proto::Error>;
```
Farhad-Shabani, line 31: Couldn't we go with `Protobuf` trait from `ibc-proto-rs` instead? by which `encode_vec` can return `Vec<u8>` instead of `Result<Vec<u8, tendermint_proto::Error>`.

Farhad-Shabani (commented):
Further to our discussions on this PR, I went over the changes once again and left a few questions to ensure some other parts of the design.

crates/ibc/src/core/context.rs
plafer: > Wouldn’t this lead to the unnecessary exposure of methods from one client (say client A) to another (client B)?

No; client A can only see the methods coming from its own trait, because that is the only trait bound that it will use in its implementation. For example, from the ADR,

```rust
impl<ClientValidationContext> ClientStateValidation<ClientValidationContext> for MyClientState
where
    ClientValidationContext: MyClientValidationContext,
{
    fn verify_client_message(
        &self,
        ctx: &ClientValidationContext,
        // ...
    ) -> Result<(), ClientError> { 
        // `get_resource_Y()` accessible through `ctx`
    }

    // ...
}
```
In this client's impl block, it can only use methods defined in `MyClientValidationContext`.

> Also, e.g. wherever there is an argument of type self::AnyClientState for a method of Client A, it potentially can accepts any other client variants defined under the AnyClientState.

By default, clients cannot do anything with `AnyClientState` because it is completely opaque. We solve that in the tendermint light client by adding [this trait bound](https://github.com/cosmos/ibc-rs/blob/746035633ff79003803cb0370553c8b9f0e4c4ed/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L492), which allows us to convert `AnyClientState` into a tendermint `ClientState`. It doesn't matter that the enum has other variants (or that it's an enum at all); from the perspective of the light client, the only thing it can do with it is convert it into its own `ClientState`.

crates/ibc/src/core/ics02_client/client_state.rs
plafer: > Not sure if we could effectively restrain the light client capabilities here (I mean the [3rd goal](https://github.com/cosmos/ibc-rs/pull/708/files#diff-f0fc94f4a7703342248727cb915b38b67295163eaf33e0696401cde405c8ec23R12-R14), stated in ADR07).
Now it is the responsibility of light client developers to ensure proper restriction of generics,

Our threat model is not "light client developers can potentially be malicious", as we clearly state that host chain developers should audit all the light client code they use. Then, in our design, we can assume non-malicious light clients (as malicious ones will not be included in any real chain). Hence, we don't need to try to restrict in any way the shape of traits that light clients use. Note that if a light client is going to be malicious, it can very well do it in a "normal looking API". The only small advantage that the current design has over the previous one in that respect is that we don't force light client developers to use a fixed set of methods, which they might misunderstand and misuse (hence leading to incorrect code).

crates/ibc/src/core/context.rs
plafer: [ce64959](https://github.com/cosmos/ibc-rs/pull/683/commits/ce64959e501fecaf0c142d869f4bd42b4504e453)

crates/ibc/src/core/ics02_client/consensus_state.rs
plafer: I tried using `Protobuf` as a supertrait, but all its trait bounds made the whole thing very messy. We should still investigate making this cleaner though.

New issue: https://github.com/cosmos/ibc-rs/issues/724

crates/ibc/src/core/ics02_client/consensus_state.rs
plafer: The `Protobuf` trait in ibc-proto-rs is trait object safe, which it trades off with being more complex, such as having `Self: erased::TryFrom<Raw>` instead of just `Self: TryFrom<Raw>`. Now that we no longer need trait object safety, we no longer need this additional complexity.

We can just bring the same changes and make `tendermint_proto::Protobuf::encode_vec()` return `Vec<u8>`?

crates/ibc/src/core/context.rs
Farhad-Shabani: For ppl implementing light clients it works so, though for the "light client developers", it is up to them whether to impose these restrictions. As in both the “impl block” and the `From<TmClientState>` bound you mentioned, they are done in the client-specific level (ics07) and not required by the bounds we have for the `ClientState` in ics02. 
However, I'm uncertain if there is a more efficient way to abstract and ensure that impls of each light client remain completely independent from one another.

Sounds like, there is an inevitable trade off between
“enabling light clients to introduce their own context methods” and “light client development complexity”

crates/ibc/src/core/ics02_client/client_state.rs
Farhad-Shabani: No, I didn’t mean restraining b/c of malicious behaviour. was thinking the same way… in terms of better development experience, reducing method misusage and being more specific of what a context argument can offer.

crates/ibc/src/core/ics02_client/consensus_state.rs
Farhad-Shabani: Hmm. We can do so but also was thinking of not being dependent on tendermint-specific things in the "core" module.

crates/ibc/src/hosts/tendermint/upgrade_proposal/context.rs
```diff
@@ -23,13 +21,13 @@ pub trait UpgradeValidationContext {
     fn upgraded_client_state(
         &self,
         upgrade_path: &UpgradeClientPath,
-    ) -> Result<Box<dyn ClientState>, UpgradeClientError>;
```
Farhad-Shabani, line 26: Regarding this upgrade context, I forgot to mention:

Although we currently maintain upgrade client interface under the Tendermint section, the intention behind using `Box<dyn ClientState>` instead of a concrete `TmClientState` type was to keep the trait as generic as possible, allowing anyone to use this abstraction until we can eventually place it under a more generalized host section. (On a side note, only `Plan` type in it, is a bit of Cosmos. Though looking at its fields, it's quite usable by other client types). 
Do you think, we can leave with `AnyClientState` and `AnyConsensusState` types here? similarly to the other contexts.

plafer: [fbd3ccc](https://github.com/cosmos/ibc-rs/pull/683/commits/fbd3ccc86b67e7c855920f45676188be3402b9d6) & [e044556](https://github.com/cosmos/ibc-rs/pull/683/commits/e044556201b9335d6d5a13854c89e44b3616cb68)

crates/ibc/src/core/ics02_client/consensus_state.rs
plafer: Indeed that would be desirable, but we should do this in another PR

Farhad-Shabani:
Reminder to add unclogs :)
 Also, I think #614 is resolved through this PR.

Farhad-Shabani (approved):
From my end, nothing else remains. Looks solid. Excellent work on this PR 👌🏻

PR #685, opened 2023-05-18
Fix ICS-20 `Amount` type

### LINKED ISSUE #684: Make ICS-20's `Amount` type transparent
We received feedback from @keppel that the `Amount` type's API is needlessly restrictive. We should simply define it as

```rust
pub struct Amount(pub U256);
```

to let users access the underlying value however they want

plafer:
Closes: #684

cc @keppel


DaviRain-Su:
I think add a new function, this is best way.

plafer:
@DaviRain-Su agreed, switched to implementing `AsRef<U256>` instead.

@Farhad-Shabani please merge if everything looks okay. We also got out-of-band approval from @keppel.

Farhad-Shabani (approved):
👌🏻

PR #693, opened 2023-05-28
fix TypeInfo implement for Timestamp

### LINKED ISSUE #687: Timestamp ser and der failed on borsh feature
## Bug Summary

Current Timestamp ser and der have failed,
```rust
use borsh::{BorshDeserialize, BorshSerialize};
let timestamp = Timestamp::now();
let encode_timestamp = timestamp.try_to_vec().unwrap();
let _ = Timestamp::try_from_slice(&encode_timestamp).unwrap();
```

report this error:
```bash
Custom { kind: InvalidData, error: "Not all bytes read" }'
```

DaviRain-Su:
Closes: #687 

## Description

When implementing the `parity scale codec` of `Timestamp`, the internal `Time` has been converted to `u64` storage, so here the type `i128` should also be changed to `u64` when implementing `TypeInfo`.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/timestamp.rs
+++ b/crates/ibc/src/core/timestamp.rs
@@ -77,7 +77,7 @@ impl scale_info::TypeInfo for Timestamp {
             .path(scale_info::Path::new("Timestamp", module_path!()))
             .composite(
                 scale_info::build::Fields::named()
-                    .field(|f| f.ty::<u6>().name("time").type_name("u64")),
+                    .field(|f| f.ty::<u64>().name("time").type_name("u64")),
             )
     }
 }
--- a/crates/ibc/src/core/timestamp.rs
+++ b/crates/ibc/src/core/timestamp.rs
@@ -75,12 +75,10 @@ impl scale_info::TypeInfo for Timestamp {
     fn type_info() -> scale_info::Type {
         scale_info::Type::builder()
             .path(scale_info::Path::new("Timestamp", module_path!()))
-            // i128 is chosen before we represent the timestamp is nanoseconds, which is represented as a i128 by Time
-            .composite(scale_info::build::Fields::named().field(|f| {
-                f.ty::<Option<i128>>()
-                    .name("time")
-                    .type_name("Option<i128>")
-            }))
+            .composite(
+                scale_info::build::Fields::named()
+                    .field(|f| f.ty::<u6>().name("time").type_name("u64")),
+            )
     }
 }
 


crates/ibc/src/core/timestamp.rs
```diff
@@ -75,12 +75,10 @@ impl scale_info::TypeInfo for Timestamp {
...
             .path(scale_info::Path::new("Timestamp", module_path!()))
-            // i128 is chosen before we represent the timestamp is nanoseconds, which is represented as a i128 by Time
-            .composite(scale_info::build::Fields::named().field(|f| {
-                f.ty::<Option<i128>>()
-                    .name("time")
-                    .type_name("Option<i128>")
-            }))
+            .composite(
+                scale_info::build::Fields::named()
+                    .field(|f| f.ty::<u64>().name("time").type_name("u64")),
```
Farhad-Shabani, line 80: From what I see, `Timestamp` exceeds 8 bytes, and the `scale_info::Type` with a `time: u64`, will not faithfully represent the `Timestamp`. Wouldn't be an issue with overflow somewhere later?

DaviRain-Su: Maybe force conversion to u128？

Farhad-Shabani: I believe, what we already have best fits the `Time` and it's ok. 
- The encode/decode has to deal with `u64` because the conversions are performed by `nanoseconds` of `u64` type.
- while the `Time` type within `Timestamp` on its own should have a size of 12 bytes (Considering the inner fields), and the `Option<i128>` represents it correctly.

Hopefully, I am not missing something for your particular use case.  if so, let me know. Happy to learn and be of assistance.

DaviRain-Su: I believe using `u64` here is appropriate because I am using the [`nanoseconds()`](https://github.com/cosmos/ibc-rs/blob/d59b303e09b3fbb415d89466dee5e9a119779aa8/crates/ibc/src/core/timestamp.rs#L56) method of the `Timestamp` struct, which is encoded as a `u64` in the `parity-scale-codec crate`. This replaces the original `Option<Time>` field in the `Timestamp` struct with a `u64`. Therefore, I think it is reasonable to declare `Timestamp` as a `u64` here, and when restoring it, I am using the `Timestamp::from_nanoseconds` method to restore it. I believe using `u64` to represent the encoding of `Timestamp` after being encoded by `parity-scale-codec` is appropriate.

Farhad-Shabani: > This replaces the original `Option<Time>` field in the Timestamp struct with a u64

:+1:

Farhad-Shabani (approved):

PR #698, opened 2023-06-01
Improve ChainId implementation; this is API breaking change

mina86:
ChainId::new allocates a new String so there’s no point in passing
ownership of name argument.  The constructor can accept it as a str
slice without loss of functionality.  This reduces allocations when
creating ids since caller no longer has to have an owned string.

Replace From implementations with TryFrom implementations.  Conversion
now fails if string is not in an epoch format.  There’s still a way to
construct an identifier without epoch format if it comes from
tendermint::chain::Id (or by using ChainId::new with zero version;
behaviour of that might change).  And since there’s From
implementation available, also get rid of ChainId::from_string.

Optimise epoch format parsing.  from_string used to check if argument
was in epoch format and then call chain_version which did it again.
To avoid duplicating work, introduce split_chain_id method which does
the parsing and returns result which includes all the information
callers might want.

Similarly, calling split and collecting values into a vector is
wasteful if all we need is the last token.  Furthermore, regexes are
quite a heavy machinery for the task of checking if string ends with
a number.  To address that, use split_last and parse the number to
check if it’s valid.

### PR author checklist:
- [n/a] Added tests.
- [n/a] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/ics24_host/identifier.rs
+++ b/crates/ibc/src/core/ics24_host/identifier.rs
@@ -3,8 +3,8 @@
 pub(crate) mod validate;
 use validate::*;
 
-use core::convert::{From, Infallible};
-use core::fmt::{Debug, Display, Error as FmtError, Formatter};
+use core::convert::From;
+use core::fmt::{Error as FmtError, Formatter};
 use core::str::FromStr;
 
 use derive_more::Into;
@@ -59,25 +59,12 @@ impl ChainId {
     /// use ibc::core::ics24_host::identifier::ChainId;
     ///
     /// let epoch_number = 10;
-    /// let id = ChainId::new("chainA".to_string(), epoch_number);
+    /// let id = ChainId::new("chainA", epoch_number);
     /// assert_eq!(id.version(), epoch_number);
     /// ```
-    pub fn new(name: String, version: u64) -> Self {
+    pub fn new(name: &str, version: u64) -> Self {
         Self {
             id: format!("{name}-{version}"),
-            version,
-        }
-    }
-
-    pub fn from_string(id: &str) -> Self {
-        let version = if Self::is_epoch_format(id) {
-            Self::chain_version(id)
-        } else {
-            0
-        };
-
-        Self {
-            id: id.to_string(),
             version,
         }
     }
@@ -103,16 +90,7 @@ impl ChainId {
     /// assert_eq!(ChainId::chain_version("testnet-helloworld-2"), 2);
     /// ```
     pub fn chain_version(chain_id: &str) -> u64 {
-        if !ChainId::is_epoch_format(chain_id) {
-            return 0;
-        }
-
-        let split: Vec<_> = chain_id.split('-').collect();
-        split
-            .last()
-            .expect("get revision number from chain_id")
-            .parse()
-            .unwrap_or(0)
+        Self::split_chain_id(chain_id).1.unwrap_or(0)
     }
 
     /// is_epoch_format() checks if a chain_id is in the format required for parsing epochs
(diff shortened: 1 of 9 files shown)


crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -123,46 +101,89 @@ impl ChainId {
...
-        Ok(Self::from_string(id))
+        Ok(id.into())
     }
 }
 
-impl Display for ChainId {
+impl core::fmt::Display for ChainId {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
-        write!(f, "{}", self.id)
+        self.id.fmt(f)
```
DaviRain-Su, line 222: Why do this?

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -123,46 +101,89 @@ impl ChainId {
...
+        match split(chain_id) {
+            Some((name, version)) => (name, Some(version)),
+            None => (chain_id, None)
+        }
+    }
+}
+
+impl Default for ChainId {
+    fn default() -> Self {
+        Self { id: String::from(DEFAULT_CHAIN_ID), version: 0 }
```
DaviRain-Su: This is invalid id, should be `Self { id: format!("{}-{}", DEFAULT_CHAIN_ID, 0), version: 0 }`

crates/ibc/src/core/ics24_host/identifier.rs
mina86: Honestly, I don’t trust Rust’s ability to optimise `write!(f, "{}", self.id)`.  I haven’t looked closely at this, but for example `foo.to_string()` is faster than `format!("{}", foo)` and I suspect the same is true in this case.

crates/ibc/src/core/ics24_host/identifier.rs
mina86: It is what the old code was doing:

```rust
    pub fn from_string(id: &str) -> Self {
        let version = if Self::is_epoch_format(id) {
            Self::chain_version(id)
        } else {
            0
        };

        Self {
            id: id.to_string(),
//              ^^ -- id is taken as given
            version,
        }
    }

// ...

impl Default for ChainId {
    fn default() -> Self {
        Self::from_string(DEFAULT_CHAIN_ID)
    }
}
```

DaviRain-Su: I know,but is not correct Init. Maybe old code have problem.

mina86: I’d argue that such change calls for a separate PR especially since it causes test failures.

DaviRain-Su: enn，agree to you. Here I think use chainid new method to implement default function.

mina86: Like I’ve mentioned it breaks existing tests and I won’t have time to work on fixing them.  If you’re not happy with how this PR looks now (and mind you it provides improvements and doesn’t introduce this bug) you’ll have to wait till I find more time to work on this.

plafer: I don't think we need to worry about the `Default` impl too much, since it is only used in tests. We are also planning on phasing out the use of defaults in tests.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -123,46 +100,95 @@ impl ChainId {
...
-            self.id = {
-                let mut split: Vec<&str> = self.id.split('-').collect();
-                let version = version.to_string();
-                if let Some(last_elem) = split.last_mut() {
-                    *last_elem = &version;
-                }
-                split.join("-")
-            };
-            self.version = version;
+        if self.version != 0 {
```
plafer, line 133: This is a slightly different check than previously. The previous implementation effectively checks if there's a `-` in the id, whereas this checks if version is `0` (which can occur with e.g. chain id `"mockChainNotEpochFormat"`).

I suggest we:
1. store an `is_epoch_format` boolean in the type itself, so that we only need to parse the chain ID upon construction.
  + Use this `is_epoch_format` instead of `self.version != 0`
2. Change this signature to `with_version(self, version: u64) -> Result<Self, IdentifierError>`
  + we needlessly require the caller to have a mutable reference to the object
  + we should fail if the identifier is not in epoch format, and let the caller decide what to do then. Note that this function is not used anywhere, so this change of signature should be straightforward

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -123,46 +100,95 @@ impl ChainId {
...
+            version,
+        }
+    }
+}
+
+impl From<String> for ChainId {
+    fn from(value: String) -> Self {
+        let version = Self::chain_version(&value);
+        Self { id: value, version }
+    }
```
plafer: Panicking is not acceptable; then one could easily crash a chain by sending a `ClientState` object with a very long version number ([here](https://github.com/cosmos/ibc-rs/blob/cef9b9285cff60e029aeef5e46f26ffdc31942ab/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L640)).

However, our previous solution of storing `0` in that case is also not that great. I suggest we simply implement `TryFrom<String>` and `TryFrom<&'a str>` instead, and bubble the errors up instead of hiding them. In tests, we can just `unwrap()`. Note that returning an error is consistent with [ibc-go's implementation](https://github.com/cosmos/ibc-go/blob/96808ebb72ec8db1edf71f4bff7e8ee68440237d/modules/core/02-client/types/height.go#L177-L181) (where a panic is equivalent to failing to transaction).

Good catch though, I never noticed this behavior before.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -123,46 +100,95 @@ impl ChainId {
...
+    /// `u64`.
+    fn split_chain_id(chain_id: &str) -> (&str, Option<u64>) {
+        fn split(chain_id: &str) -> Option<(&str, u64)> {
+            let (name, version) = chain_id.rsplit_once('-')?;
+            if name.is_empty() {
+                return None;
+            }
+            let (car, cdr) = version.as_bytes().split_first()?;
+            if b'1' <= *car && *car <= b'9' && cdr.iter().all(u8::is_ascii_digit) {
+                Some((name, version.parse().unwrap()))
```
plafer: As mentioned in other comments, we'd need to return an `IdentifierError` when the version parsing here fails

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -123,46 +100,95 @@ impl ChainId {
...
+
+    /// Splits chain_id into name and version if the id includes epoch number.
+    ///
+    /// Chain id with epoch number has format `{name}-{version}` where version
+    /// is a positive integer starting with non-zero digit.  If `chain_id` is in
+    /// that format, returns `(name, Some(version))`; otherwise returns
+    /// `(chain_id, None)`.
+    ///
+    /// **Panics** if chain id is in epoch format but the version overflows
+    /// `u64`.
```
plafer: Reminder to remove this after removing the panicking behavior

plafer (changes requested):
Thank you for this @mina86, this is a much needed cleanup of the type.

crates/ibc/src/core/ics24_host/identifier.rs
mina86: `self.version == 0` implies that the id is not in epoch format so
checking it is just an optimisation¹.  As far as I can tell the old
and new code behave the same way (module panic if version doesn’t fit
`u64`).  Neither the old nor the new code will change identifier such
as `foo-bar` for example.

I can add `is_epoch_format` if you prefer but it will increase the
size of the `ChainId` structure which I feel like it could end up
hurting rather than helping.

Returning `Result` certainly makes sense though note that changing
`mut self` to `self` doesn’t actually matter for the caller.  `self`
is moved into the method.  What possibly would make most sense is
`with_version(self, u64) -> Result<Self, Self>` (or changing the
semantics such that it adds version if not already in epoch format).

crates/ibc/src/core/ics24_host/identifier.rs
mina86: So to clarify, the desired behaviour is:
- `foo` → valid not in epoch format,
- `foo-bar` → valid not in epoch format,
- `foo-0` → valid not in epoch format (?),
- `foo-1` → valid in epoch format and
- `foo-18446744073709551616` → invalid.

PS. Regarding panicking, I was under the impression that this code is
executed within the VM where panics result in transaction failure.

crates/ibc/src/core/ics24_host/identifier.rs
plafer: > `self.version == 0` implies that the id is not in epoch format so
checking it is just an optimisation¹.

Ah right, I didn't realize `"foo-0"` is not considered epoch format. So basically we had it wrong before, and this is correct.

> I can add is_epoch_format if you prefer

Nope, this is no longer necessary (see previous comment).

> What possibly would make most sense is `with_version(self, u64) -> Result<Self, Self>` (or changing the semantics such that it adds version if not already in epoch format).

The reason I still prefer my suggestion is that when not in epoch format, the concept of `version` in a chain ID doesn't exist. In practice, all chains now use the epoch format. I do not see a use case to "convert your chain ID to epoch format"; most likely, if someone called `with_version()` on a chain ID that is not in epoch format, it's a bug and they should fix it, as opposed to intended behavior.

Note that this whole "encode the version number in the chain id" evolved as a hack; it wasn't in the original IBC standard, and is pretty messy. Non-epoch format chain IDs are pretty much non-existent.

crates/ibc/src/core/ics24_host/identifier.rs
plafer: > So to clarify, the desired behaviour is:

That's correct. I use [this golang playground](https://play.golang.com/p/guIY1ZxSyeU) (extracted from ibc-go's implementation) to figure out which strings should be considered epoch format.

Note that `foo-18446744073709551616` is considered valid given this regexp, but they later fail when trying to parse the version number into a `u64`.

> PS. Regarding panicking, I was under the impression that this code is executed within the VM where panics result in transaction failure.

We make no such assumption. And as it turns out, most (if not all) of our users don't use a VM at all; everything is compiled directly in the binary. So a panic would make the whole node crash (unless of course they guard against it, but we don't assume they do).

mina86: What’s the output of `ChainId::chain_version("foo-18446744073709551616")`?

What’s the output of `ChainId::from(tendermint::chain::Id("foo-18446744073709551616"))`? That conversion is used for serde deserialisation.

plafer: > What’s the output of ChainId::chain_version("foo-18446744073709551616")?

That should be an error; so the return type should be changed to `Result<u64, IdentifierError>`

> What’s the output of ChainId::from(tendermint::chain::Id("foo-18446744073709551616"))?

We should implement `TryFrom<tendermint::chain::Id> for `ChainId` instead, and return an `IdentifierError`.

plafer:
Is #721 meant to replace this PR?

mina86:
> Is #721 meant to replace this PR?

Not quite.  Since this PR is growing I’ve moved some of the changes to separate PR.

Farhad-Shabani:
Given that #762 is being merged, this PR gets outdated. Feel free to open an issue, if still any of the above has not been addressed.

PR #699, opened 2023-06-01
fix: match Order

### LINKED ISSUE #394: Remove `ChannelEnd::order_matches()`
We should use a `match` statement instead. This pattern encourages not looking at all possible variants, and shouldn't be used.

At the same time, we should understand what the `Order::None` represents, and if we keep it, what to do in cases where the order is `None`.

amirHdev:
Closes: #394

### DIFF
--- a/crates/ibc/src/core/ics04_channel/channel.rs
+++ b/crates/ibc/src/core/ics04_channel/channel.rs
@@ -282,7 +282,12 @@ impl ChannelEnd {
 
     /// Helper function to compare the order of this end with another order.
     pub fn order_matches(&self, other: &Order) -> bool {
-        self.ordering.eq(other)
+        match (self.ordering, other) {
+            (Order::None, Order::None) => true,
+            (Order::Unordered, Order::Unordered) => true,
+            (Order::Ordered, Order::Ordered) => true,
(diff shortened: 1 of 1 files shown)


plafer:
Thank you for submitting your first contribution!

Although @DaviRain-Su suggested doing this in the issue, this is not really what I had in mind. The goal is to *remove* the `order_matches` method.

I looked into it again, and I believe we should change all calls of the form

```rust
if chan_end.order_matches(&Order::Ordered) {
  /* if branch contents */
} else {
  /* else branch contents */
}
```

to 

```rust
match chan_end_on_b.ordering {
    Order::Ordered => {
        /* if branch contents */
    },
    Order::Unordered => {
        /* else branch contents */
    }
    Order::None => {
        /* do nothing */
    }
}
```

, and then remove `ChannelEnd::order_matches`.

This matches the [ibc-go behavior](https://github.com/cosmos/ibc-go/blob/96808ebb72ec8db1edf71f4bff7e8ee68440237d/modules/core/04-channel/keeper/packet.go#L207-L248).

plafer:
Closing as stale; feel free to open a new PR with the abovementionned suggestions applied!

PR #720, opened 2023-06-16
Suggestion for simplifying client contexts in PR 296

Farhad-Shabani:
______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/context.rs
+++ b/crates/ibc/src/core/context.rs
@@ -9,6 +9,8 @@ use ibc_proto::google::protobuf::Any;
 
 use crate::core::events::IbcEvent;
 use crate::core::ics02_client::error::ClientError;
+use crate::core::ics02_client::ClientExecutionContext;
+use crate::core::ics02_client::ClientValidationContext;
 use crate::core::ics03_connection::connection::ConnectionEnd;
 use crate::core::ics03_connection::error::ConnectionError;
 use crate::core::ics03_connection::version::{
@@ -24,16 +26,12 @@ use crate::core::ics23_commitment::commitment::CommitmentPrefix;
 use crate::core::ics24_host::identifier::ClientId;
 use crate::core::ics24_host::identifier::ConnectionId;
 use crate::core::ics24_host::path::{
-    AckPath, ChannelEndPath, ClientConnectionPath, ClientConsensusStatePath, CommitmentPath,
-    ConnectionPath, ReceiptPath, SeqAckPath, SeqRecvPath, SeqSendPath,
+    AckPath, ChannelEndPath, ClientConnectionPath, CommitmentPath, ConnectionPath, ReceiptPath,
+    SeqAckPath, SeqRecvPath, SeqSendPath,
 };
 use crate::core::router::Router;
 use crate::core::timestamp::Timestamp;
 use crate::Height;
-
-use super::ics02_client::client_state::ClientState;
-use super::ics02_client::consensus_state::ConsensusState;
-use super::ics02_client::ClientExecutionContext;
 
 /// Top-level error
 #[derive(Debug, Display, From)]
@@ -92,33 +90,11 @@ impl std::error::Error for RouterError {
 /// Context to be implemented by the host that provides all "read-only" methods.
 ///
 /// Trait used for the top-level [`validate`](crate::core::validate)
-pub trait ValidationContext: Router {
-    type ClientValidationContext;
-    type E: ClientExecutionContext;
-    type AnyConsensusState: ConsensusState;
-    type AnyClientState: ClientState<Self::ClientValidationContext, Self::E>;
-
-    fn get_client_validation_context(&self) -> &Self::ClientValidationContext;
-    fn get_client_execution_context(&mut self) -> &mut Self::E;
-
-    /// Returns the ClientState for the given identifier `client_id`.
-    ///
-    /// Note: Clients have the responsibility to store client states on client creation and update.
-    fn client_state(&self, client_id: &ClientId) -> Result<Self::AnyClientState, ContextError>;
+pub trait ValidationContext: Router + ClientValidationContext {
+    fn get_client_validation_context(&self) -> &Self::V;
 
     /// Tries to decode the given `client_state` into a concrete light client state.
     fn decode_client_state(&self, client_state: Any) -> Result<Self::AnyClientState, ContextError>;
-
-    /// Retrieve the consensus state for the given client ID at the specified
-    /// height.
-    ///
-    /// Returns an error if no such state exists.
-    ///
-    /// Note: Clients have the responsibility to store consensus states on client creation and update.
-    fn consensus_state(
-        &self,
-        client_cons_state_path: &ClientConsensusStatePath,
-    ) -> Result<Self::AnyConsensusState, ContextError>;
 
     /// Returns the time when the client state for the given [`ClientId`] was updated with a header for the given [`Height`]
     fn client_update_time(
@@ -243,7 +219,9 @@ pub trait ValidationContext: Router {
 /// Context to be implemented by the host that provides all "write-only" methods.
 ///
 /// Trait used for the top-level [`execute`](crate::core::execute) and [`dispatch`](crate::core::dispatch)
-pub trait ExecutionContext: ValidationContext {
+pub trait ExecutionContext: ValidationContext + ClientExecutionContext {
+    fn get_client_execution_context(&mut self) -> &mut Self::E;
+
     /// Called upon client creation.
     /// Increases the counter which keeps track of how many clients have been created.
     /// Should never fail.
--- a/crates/ibc/src/core/ics02_client/context.rs
+++ b/crates/ibc/src/core/ics02_client/context.rs
@@ -5,6 +5,7 @@ use crate::core::ics24_host::path::ClientConsensusStatePath;
 use crate::core::ics24_host::path::ClientStatePath;
 use crate::core::ContextError;
 
+/// Defines the types that all clients must implement.
 pub trait ClientTypes {
     type V: ClientValidationContext;
     type E: ClientExecutionContext;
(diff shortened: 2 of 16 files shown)


crates/ibc/src/core/ics02_client/context.rs
```diff
@@ -1,9 +1,33 @@
...
+    type E: ClientExecutionContext;
+    type AnyClientState: ClientState<Self::V, Self::E>;
+    type AnyConsensusState: ConsensusState;
+}
+
+/// Defines the methods that all client `ValidationContext`s (precisely the
+/// generic parameter of
+/// [`crate::core::ics02_client::client_state::ClientStateValidation`] ) must
+/// implement.
+pub trait ClientValidationContext: ClientTypes + Sized {
```
plafer: With the `ClientTypes` supertrait, we claim that every client's `ValidationContext` will need:
+ access to `consensus_state()` and `client_state()` methods
+ types `AnyClientState`, `AnyConsensusState`, `ClientValidationContext` and `ClientExecutionContext`

We can already see that the Tendermint client doesn't need the `client_state()` method, nor the type `AnyClientState`, `V` and `E`. 

Basically, this design works well when the clients' `ValidationContext`s are implemented on the same struct as the core `ValidationContext` (as is the case for `MockContext`). It falls apart when this is not the case: to implement `TmValidationContext` on a different type, you would end up putting bogus values for `V`, `E`, `AnyConsensusState`, and have `client_state()` be `unimplemented!()` (since it is never called). I consider this a code smell.

The reason I added a `ClientExecutionContext` trait is really because it is the clients' responsibility to store client/consensus states, so we *know* that they will need these methods. Those "special" clients who don't have consensus states (such as solomachine) will use `unimplemented!()` for the `consensus_state()` (the one downside of the design), but they're the exception rather than the rule.

crates/ibc/src/core/ics02_client/context.rs
```diff
@@ -13,11 +37,7 @@ use super::{client_state::ClientState, consensus_state::ConsensusState};
 /// Specifically, clients have the responsibility to store their client state
 /// and consensus states. This trait defines a uniform interface to do that for
 /// all clients.
-pub trait ClientExecutionContext: Sized {
-    type ClientValidationContext;
-    type AnyClientState: ClientState<Self::ClientValidationContext, Self>;
-    type AnyConsensusState: ConsensusState;
-
+pub trait ClientExecutionContext: ClientValidationContext + Sized {
```
plafer: Same here: we are forcing the clients' `ExecutionContext` to define a bunch of things we don't know for sure that they'll need (`V`, `E`, `client_state()`, etc), which works OK when all the traits are implemented on the same struct, but falls apart when it isn't (which would be the case for Namada).

The Tendermint client's `ExecutionContext` doesn't need `client_state()`, `V`, `E`, `AnyClientState`, and so shouldn't have to define them.

crates/ibc/src/core/context.rs
```diff
@@ -92,34 +90,12 @@ impl std::error::Error for RouterError {
...
-    type AnyClientState: ClientState<Self::ClientValidationContext, Self::E>;
-
-    fn get_client_validation_context(&self) -> &Self::ClientValidationContext;
-    fn get_client_execution_context(&mut self) -> &mut Self::E;
-
-    /// Returns the ClientState for the given identifier `client_id`.
-    ///
-    /// Note: Clients have the responsibility to store client states on client creation and update.
-    fn client_state(&self, client_id: &ClientId) -> Result<Self::AnyClientState, ContextError>;
+pub trait ValidationContext: Router + ClientValidationContext {
```
plafer, line 93: Given my other comments, having `ClientValidationContext` as a supertrait of the core `ValidationContext` no longer works - ultimately batching all the associated types in a `ClientTypes` trait would work if all these traits (core and client's `ValidationContext`/`ExecutionContext`) needed all the types, but they don't

crates/ibc/src/core/ics04_channel/context.rs
```diff
@@ -1,46 +1,28 @@
...
+use crate::core::{ContextError, ExecutionContext, ValidationContext};
 
 use super::packet::Sequence;
 
 /// Methods required in send packet validation, to be implemented by the host
-pub trait SendPacketValidationContext {
-    type ClientValidationContext;
-    type E: ClientExecutionContext;
-    type AnyConsensusState: ConsensusState;
-    type AnyClientState: ClientState<Self::ClientValidationContext, Self::E>;
```
plafer, line 27: In this case, `SendPacketValidationContext` really is a subset of the core `ValidationContext` - which is why we have the blanket implementation if a type already implements `ValidationContext`. I'm not sure if we need `SendPacketValidationContext` really... the `send_packet()` function is not fundamentally different from the rest of the handlers, I think it's safe to assume that everyone will use the same type for `ValidationContext` and `ExecutionContext`.

crates/ibc/src/core/ics02_client/context.rs
Farhad-Shabani: > With the ClientTypes supertrait, we claim that every client's ValidationContext will need:
access to consensus_state() and client_state() methods
types AnyClientState, AnyConsensusState, ClientValidationContext and ClientExecutionContext
We can already see that the Tendermint client doesn't need the client_state() method, nor the type AnyClientState, V and E.

Tendermint itself doesn't use it, but isn’t that much tailored specifically for the Tendermint clients?
I'm referring to the implicit design choice we made. We've granted the mentioned types to the execution trait, meaning clients have the write access. Based on this premise, shouldn't clients also have the read access? be aware of their own states? If someone wants to use it for validation, they can, even though it might not be utilized by all, like in the case of Tendermint.

Plus, my view was a bit beyond the requirements of light clients alone. The `ClientValidationContext` interface provides added flexibility in terms of granting client validation access to any components if needed. This can include a light client or even an application (e.g. in case of `SendPacketValidationContext`)

Farhad-Shabani: > Basically, this design works well when the clients' ValidationContexts are implemented on the same struct as the core ValidationContext (as is the case for MockContext). It falls apart when this is not the case: to implement TmValidationContext on a different type, you would end up putting bogus values for V, E, AnyConsensusState, and have client_state() be unimplemented!() (since it is never called). I consider this a code smell.

This is a crucial point came into the light and highlights two aspects:

(1) Depending on how hosts define context structs, a design can originate from a wrong impl, which imho necessitated a revision of `MockContext` to get it to the correct definition. Releasing with the current Mock implementation can be a source of significant confusion. The same can be said for basecoin. Can we carry out this before finalizing the ADR07?

(2) Given your explanation, ADR07 ideally shouldn't function with the current mock setup. This gives the impression that we might have adjusted parts of our design to accommodate the existing codes.

crates/ibc/src/core/ics02_client/context.rs
Farhad-Shabani: > The Tendermint client's ExecutionContext doesn't need client_state(), V, E, AnyClientState, and so shouldn't have to define them

From what I see the [`ClientExecutionContext`](https://github.com/cosmos/ibc-rs/blob/1d46026d9729622ca355b86e6bb52727c69c02e7/crates/ibc/src/core/ics02_client/context.rs#L16-L19) within PR296 already has the access to the most of them, e.g. `client_state()` `AnyClientState` and even part of `V` through `ClientValidationContext`. Am I missing something?

crates/ibc/src/core/ics04_channel/context.rs
Farhad-Shabani: My impression: `SendPacketValidationContext` was defined to provide more specificity and avoid granting excessive access to all `ValidationContext` methods. 
Btw, here is one the places we benefit from having `ClientValidtionContext`

crates/ibc/src/core/ics02_client/context.rs
plafer: Right, I redact to:

> The Tendermint client's ExecutionContext doesn't need `client_state()` and so shouldn't have to define it.

crates/ibc/src/core/ics02_client/context.rs
plafer: > Tendermint itself doesn't use it, but isn’t that much tailored specifically for the Tendermint clients?

It's the other way around: the Tendermint client works as a counter example here. Basically, if I can show you any client where its `ValidationContext` doesn't need one of the methods you're forcing it to have with the `ClientValidationContext` trait, then this shows a weakness in the design.

I think the `ClientExecutionContext` confused you here. Only until the very end did I add the `ClientExecutionContext`; the whole thing works without it. I only found that since our API explicitly gives the responsibility to clients to store their client & consensus states, then perhaps it would be clearer for everyone if we had a trait (`ClientExecutionContext`) which gave a uniform interface for all the clients to use to perform these 2 tasks. Note that we don't *need* it; every client *could* define their own `store_client_state()` and `store_consensus_state()`. Except now, that would make it quite tedious for the host, as each would have a slightly different interface, but semantically would achieve the same outcome (i.e. storing a client state and consensus state).

> Based on this premise, shouldn't clients also have the read access? be aware of their own states?

They absolutely can. Every client defines its own `ValidationContext`; if they need to read the client or consensus state, then all they need to do is add a method `client_state()` or `consensus_state()`. My point is: we don't know explicitly know if that will be necessary, so why enforce it with a `ClientValidation` trait bound? In other words, adding a `ClientValidation` trait bound is *limiting* because it enforces a structure on their contexts that we don't know if they will need, whereas not putting any trait bound lets them define their context however they see fit (including but not limited to reading their own `ClientState` and `ConsensusState`).

> Plus, my view was a bit beyond the requirements of light clients alone. The ClientValidationContext interface provides added flexibility in terms of granting client validation access to any components if needed. This can include a light client or even an application (e.g. in case of SendPacketValidationContext)

I don't follow here. `send_packet()` is part of `core` (i.e. we control the implementation), so there's no uncertainty as to which `Context` method will be needed.

Similarly for applications (i.e. implementing the `Module` trait), if we were to start using this pattern, imposing `ClientValidation` as a trait bound is *more limiting* than not putting any trait bound at all.

plafer: > (1) Depending on how hosts define context structs, a design can originate from a wrong impl, which imho necessitated a revision of MockContext to get it to the correct definition.

What would be a "correct" `MockContext`? IMO it's fine as is, exactly because our design should not prevent the user from implementing all the traits on the same struct - but neither should it *force* the user to. AFAIK, the current ADR7 design achieves this goal.

> (2) Given your explanation, ADR07 ideally shouldn't function with the current mock setup.

Why is that?

Farhad-Shabani:
Per our discussion, this is closed.

PR #725, opened 2023-06-21
Optimise ChainId epoch format parsing

mina86:
from_string checks if argument is in epoch format and then calls
chain_version which does it again.  chain_version and with_version use
regex to check if string is in epoch format and then not only parse
the string again but unnecessarily collect dash-separated parts into
a new vector.

Optimise it by providing split_chain_id method which checks the format
and returns individual name and version parts of the id if it is in
epoch format.  Furthermore, don’t use regex which is a heavy machinery
for simply checking if string ends in a positive integer.

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/ics24_host/identifier.rs
+++ b/crates/ibc/src/core/ics24_host/identifier.rs
@@ -65,16 +65,9 @@ impl ChainId {
     }
 
     pub fn from_string(id: &str) -> Self {
-        let version = if Self::is_epoch_format(id) {
-            Self::chain_version(id)
-        } else {
-            0
-        };
-
-        Self {
-            id: id.to_string(),
-            version,
-        }
+        let version = Self::chain_version(id);
+        let id = String::from(id);
+        Self { id, version }
     }
 
     /// Get a reference to the underlying string.
@@ -98,15 +91,8 @@ impl ChainId {
     /// assert_eq!(ChainId::chain_version("testnet-helloworld-2"), 2);
     /// ```
     pub fn chain_version(chain_id: &str) -> u64 {
-        if !ChainId::is_epoch_format(chain_id) {
-            return 0;
-        }
-
-        let split: Vec<_> = chain_id.split('-').collect();
-        split
-            .last()
-            .expect("get revision number from chain_id")
-            .parse()
+        split_chain_id(chain_id)
+            .and_then(|(_, version)| u64::from_str(version).ok())
             .unwrap_or(0)
     }
 
@@ -120,8 +106,7 @@ impl ChainId {
     /// assert_eq!(ChainId::is_epoch_format("c-1"), true);
     /// ```
     pub fn is_epoch_format(chain_id: &str) -> bool {
-        let re = safe_regex::regex!(br".*[^-]-[1-9][0-9]*");
-        re.is_match(chain_id.as_bytes())
+        split_chain_id(chain_id).is_some()
     }
 
     /// with_version() checks if a chain_id is in the format required for parsing epochs, and if so
@@ -131,19 +116,33 @@ impl ChainId {
     /// assert_eq!(ChainId::new("chainA", 1).with_version(2), ChainId::new("chainA", 2));
     /// assert_eq!("chain1".parse::<ChainId>().unwrap().with_version(2), "chain1".parse::<ChainId>().unwrap());
     /// ```
-    pub fn with_version(mut self, version: u64) -> Self {
-        if Self::is_epoch_format(&self.id) {
-            self.id = {
-                let mut split: Vec<&str> = self.id.split('-').collect();
-                let version = version.to_string();
-                if let Some(last_elem) = split.last_mut() {
-                    *last_elem = &version;
-                }
-                split.join("-")
-            };
-            self.version = version;
+    pub fn with_version(self, version: u64) -> Self {
+        if self.version == 0 {
+            return self;
         }
-        self
+        // self.version ≠ 0 ⇒ self.id is in epoch format ⇒ split_chain_id can’t fail.
+        let (name, _) = split_chain_id(&self.id).expect("Never fails");
+        Self::new(name, version)
+    }
+}
+
+/// Splits chain_id into name and version if the id includes epoch number.
+///
+/// Chain id with epoch number has format `{name}-{version}` where name is
+/// non-empty and version is a positive integer starting with a non-zero digit.
+/// If `chain_id` is in that format, returns `(name, version)`; otherwise
+/// returns an `None`.
+fn split_chain_id(chain_id: &str) -> Option<(&str, &str)> {
+    let (name, version) = chain_id.rsplit_once('-')?;
+    // `-123` and `foo--123` are not in epoch format.
+    if name.is_empty() || name.as_bytes().last() == Some(&b'-') {
+        return None;
+    }
+    let (car, cdr) = version.as_bytes().split_first()?;
+    if matches!(*car, b'1'..=b'9') && cdr.iter().all(u8::is_ascii_digit) {
+        Some((name, version))
+    } else {
+        None
     }
 }
 


mina86:
@plafer, friendly ping.

Farhad-Shabani:
@mina86 apologies for the delay in getting back to you. We have been looking for an appropriate time to carefully re-examine the details. Specifically, Under `ics24_host/identifier/validate.rs` there are some functions that could be modified and used for `ChainId` as well. Wanted to ensure we're on the right track. Though, we still may proceed in a timely manner. Hope you won't be blocked :)

Also, I’ve summarized your points that haven’t been tackled yet, plus some others we came across in issue #761 for our reference and the `unclog`. Since the new changes would greatly differ from the existing PRs, I'll open a new one. I hope that's okay with you. Let us know if there's anything we may have overlooked. 
Again, much appreciated for bringing these up :pray:

mina86:
Want me to drop this and #698 PRs? I think this PR might still be valid as it addresses some of the issues and refactors code to form where further changes will be easier to make.

Farhad-Shabani:
> Want me to drop this and #698 PRs? I think this PR might still be valid as it addresses some of the issues and refactors code to form where further changes will be easier to make.

I've opened #762. Though still is a work in progress, you can take a look at changes there, which are not in line with the mentioned PRs. Btw, I suggest keep them open till we make sure everything is sorted in #762. If not, we can come back  later.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -131,19 +117,38 @@ impl ChainId {
...
+            return Err(());
+        }
+        let (car, cdr) = version.as_bytes().split_first().ok_or(())?;
+        if matches!(*car, b'1'..=b'9') && cdr.iter().all(u8::is_ascii_digit) {
+            Ok((name, version))
+        } else {
+            Err(())
         }
-        self
     }
```
Farhad-Shabani: This function works independent of any instance of `ChainId`. Good to promote its usability and having it as a free function, imo.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -131,19 +117,38 @@ impl ChainId {
...
+    pub fn with_version(self, version: u64) -> Self {
+        if self.version == 0 {
+            self
+        } else {
+            let (name, _) = Self::split_chain_id(&self.id).unwrap();
+            // TODO: We're wasting allocation here.  We own self.id, so we could
+            // `self.id.truncate(name.len() + 1); write!(&mut self.id, "{}",
+            // version)`. Unfortunately, we’re no_std so `write!` is out and
+            // `self.id.push_str(&version.to_string());` defeats the purpose of
+            // the optimisation.  `itoa` crate is possible option.
```
Farhad-Shabani: Would you drop this comment? 
I see the point, though right now, this amount of optimization is not really matter.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -131,19 +117,38 @@ impl ChainId {
...
+        }
+    }
+
+    /// Splits chain_id into name and version if the id includes epoch number.
+    ///
+    /// Chain id with epoch number has format `{name}-{version}` where name is
+    /// non-empty and version is a positive integer starting with a non-zero
+    /// digit.  If `chain_id` is in that format, returns `(name, version)`;
+    /// otherwise returns an error.
+    fn split_chain_id(chain_id: &str) -> Result<(&str, &str), ()> {
```
Farhad-Shabani: Here would be ideal to return of a `Result<(&str, u64), IdentifierError>` type, so then:

- Can utilize `parse()` method and avoid manually parsing versions.
- Access to a `u64` wherever `is_split_chain` called without additional version conversion, even though it adds a bit of overhead.
- Returns the `IdentifierError` as a known type to the system.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -131,19 +117,38 @@ impl ChainId {
...
-                    *last_elem = &version;
-                }
-                split.join("-")
-            };
-            self.version = version;
+    pub fn with_version(self, version: u64) -> Self {
+        if self.version == 0 {
+            self
+        } else {
+            let (name, _) = Self::split_chain_id(&self.id).unwrap();
```
Farhad-Shabani: Can here we leave with `expect` instead of `unwrap()`?

Farhad-Shabani (changes requested):
@mina86 Let’s get this PR done first :)

Left some comments. 
Also, can you add a changelog entry for this under .changelog/unreleased/improvements?

crates/ibc/src/core/ics24_host/identifier.rs
mina86: I’d rather not do this in this PR to be honest.  Having this function parse the version changes the behaviour of `is_epoch_format` which I’d rather do in separate commit so this one can go in without having to worry about change in behaviour.

crates/ibc/src/core/ics24_host/identifier.rs
mina86: Honestly I don’t see the point of `expect` method.  Presumably it gives more context about the failure except the method is never used that way.  In this repository, 72% uses of `expect` simply use `Never fails` as error message.  That’s just pointless noise and in those places `unwrap` would be better.

Instead I’ve added a comment explaining why the unwrap never fails.

mina86: Oh Christ, I see you guys enabled this in Clippy. Fine then.

crates/ibc/src/core/ics24_host/identifier.rs
Farhad-Shabani: **(1)** Regarding the error type, let's stick to using `IdentifierError` anywhere that would return an identifier error. So, handling/propagating errors would remain consistent across `ibc-rs`.

**(2)** About the version parsing thing, mainly meant to keep things simple which got me thinking more... Are we sure changing `is_epoch_format` is worth all this added complexity? 
Ran a quick local benchmark to compare performance between regex and non-regex approaches. This is on the order of ~1 **microsecond**! Honestly, we're not gaining much b/c of that. It's much better to prioritize readability/maintainability instead of optimization. Particularly, I noticed it still considers e.g. `.   -1` and `comsos   hub-1` (with whitespace) valid! Let's keep things simple and as is for `is_epoch_format` in this PR.

Btw, still through this PR, the redundant call to `is_epoch_format` in `from_string` would be removed

crates/ibc/src/core/ics24_host/identifier.rs
Farhad-Shabani: Indeed. We definitely should do refinements around our `expect`, and even a bunch of error messages.

crates/ibc/src/core/ics24_host/identifier.rs
mina86: To me 1µs to parse handful of bytes does sound like a lot of time especially as this code is going to be used in smart contracts when execution time and code size (I have another change which removes safe-regex dependency completely) are more expensive than on bare metal.

Regarding other error checking that can happen in follow up changes.  Your #762 seems like it going into that direction and if there are other existing functions checking if identifiers consist of valid characters those will be able to be reused.

Farhad-Shabani: > To me 1µs to parse handful of bytes does sound like a lot of time especially as this code is going to be used in smart contracts when execution time and code size (I have another change which removes safe-regex dependency completely) are more expensive than on bare metal.

Totally see your point! Though, given the costs the entire IBC messaging process already has, this one is not honestly a major one, particularly when it comes to **(1)correctness** and **(2)maintainability**, I still strongly value them above **(3)optimization** here. Btw, feel free to share more context if this really matters to your project.

> Regarding other error checking that can happen in follow up changes. Your #762 seems like it going into that direction and if there are other existing functions checking if identifiers consist of valid characters those will be able to be reused.

Hmm, note that #762 vs this PR has different views and even implementation. Here, we're dropping regex b/c of **optimization** vs there, a part of regex might be discarded b/c of fixing the two bugs noted in #761 (**correctness**) and delegating [these checks](https://github.com/cosmos/ibc-rs/pull/725/files#diff-0e33ca83458b3683d1a621c10a75653bdc80b52ceff5ea6093669329653a7e06R138-R140) to an [existing function](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/core/ics24_host/identifier/validate.rs#L9-L51) (**maintainability**) so all identifiers get checked consistently too.

Anyhow, to not mess with the mentioned checks right after this PR and see a desire to leave with `rsplit_once` :)

1. I suggest we port here part of [762](#762) modifications related to checks in `split_chain_id`:
Basically, the [`validate_identifier`](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/core/ics24_host/identifier/validate.rs#L15) should be broken down. One for checking the characters say `validate_identifier_chars`, and the other would handle length verification say `validate_identifier_length` (like [here](https://github.com/cosmos/ibc-rs/pull/762/files#diff-67d84b11126e7bc766da4993d52c80aa7fc7ddef2aa92adeafeeac97491b43caR12-R54)). Then, we can use `validate_identifier_chars` to ensure the chain name meets ICS-24 spec and maintain identifier checks consistent across `IBC-rs`. 
2. Though, for the version string check, still believe regex and a `parse()` afterward suffice.

3. And, much appreciated to include a `unclog` 😊

mina86: > 1. I suggest we port here part of #762 modifications related to checks in `split_chain_id`: Basically, the `validate_identifier`
> 2. Though, for the version string check, still believe regex and a `parse()` afterward suffice.

But notice that at that point regular expressions won’t simplify things.  rsplit_once can split identifier into name and version, validate_identifier can validate the name, parse can validate the version and we only need check that name doesn’t end with a dash¹ and version doesn’t start with a zero:

```rust
fn split_chain_id(chain_id: &str) -> Result<(&str, u64), _> {
    let (name, version) = chain_id.rsplit_once('-').ok_or(NotEpochFormat)?;
    validate_identifier(name, 6, 43)?;
    if name.last() == Some(&b'-') || version.first() == Some(&b'0') {
        return Err(NotEpochFormat);
    }
    let version = NonZeroU64::from_str(version).map_err(|_| NotEpochFormat)?.get();
    Ok((name, version))
}
```

And because safe-regex doesn’t work on strings, the same function using regex becomes longer due to additional `&[u8]` → `str` conversion:

```rust
fn split_chain_id(chain_id: &str) -> Result<(&str, u64), _> {
    let (name, version) = regex!(br"(.*[^-])-([1-9][0-9]*)")
        .match_slices(chain_id.as_bytes())
        .ok_or_else(|| NotEpochFormat)?;
    let name = unsafe { core::str::from_utf8_unchecked(name) };
    validate_identifier(name, 6, 43)?;
    let version = unsafe { core::str::from_utf8_unchecked(version) };
    let version = NonZeroU64::from_str(version).map_err(|_| NotEpochFormat)?;
    Ok((name, version.get()))
}
```

----

¹ To be honest, since validate_identifier allows identifiers ending with a dash, I’d even argue that `chain--1` should be considered valid chain identifier.

Farhad-Shabani: > And because safe-regex doesn’t work on strings, the same function using regex becomes longer due to additional &[u8] → str conversion:
>     let (name, version) = regex!(br"(.*[^-])-([1-9][0-9]*)")

I meant using regex only for checking the version after extracting its string by `rsplit_once`.


> ```rust
> fn split_chain_id(chain_id: &str) -> Result<(&str, u64), _> {
>     let (name, version) = chain_id.rsplit_once('-').ok_or(NotEpochFormat)?;
>     validate_identifier(name, 6, 43)?;
>     if name.last() == Some(&b'-') || version.first() == Some(&b'0') {
>         return Err(NotEpochFormat);
>     }
>     let version = NonZeroU64::from_str(version).map_err(|_| NotEpochFormat)?.get();
>     Ok((name, version))
> }
> ```

Nevertheless, I think the above function works with a bit of adjustment:

> ¹ To be honest, since validate_identifier allows identifiers ending with a dash, I’d even argue that chain--1 should be considered valid chain identifier.

On the same page with you. `name.last() == Some(&b'-')` not needed. 👍🏻 

> version doesn’t start with a zero

True. should have this check. Though, note that `version == 0` is valid but others starting with zero no. 

> Err(NotEpochFormat)

For the errors, the `IdentifierError` type should be used

> validate_identifier(name, 6, 43)?;

Min length == 3. Right?

mina86: > On the same page with you. `name.last() == Some(&b'-')` not needed.

Note though that the check is in the current code:

    pub fn is_epoch_format(chain_id: &str) -> bool {
        let re = safe_regex::regex!(br".*[^-]-[1-9][0-9]*");
    //                                   ^^^^
        re.is_match(chain_id.as_bytes())
    }

> > version doesn’t start with a zero

> True. should have this check. Though, note that `version == 0` is valid but others starting with zero no.

Current code rejects zero:

        let re = safe_regex::regex!(br".*[^-]-[1-9][0-9]*");
    //                                         ^^^^
        re.is_match(chain_id.as_bytes())

It is true that ChainId::new will happily create such identifiers though.

> > validate_identifier(name, 6, 43)?;
>
> Min length == 3. Right?

~~I’ve looked at ICS 24 and it suggests default range for channel identifier to be [8, 64].  Adjusting it for a dash and version which together take from 2 to 21 characters, I came up with [6, 43].~~  [Never mind, it’s channel identifier not chain identifier so perhaps it doesn’t apply?]  If we look at the current code it’s more like minimum = 1.  I don’t have an opinion.  [6, 43] was just for demonstration.

----

Anyway, if you’re ok moving away from regex, I would suggest getting this PR merged and than I can pick up #698 again with more validation.  There we can discuss all the rules that identifiers need to follow and how the API should behave when encountering invalid identifiers.  There is *a lot* of questions to consider (which is why I want to do this cleanup first before dealing validation).

Farhad-Shabani: > If we look at the current code it’s more like minimum = 1. I don’t have an opinion. [6, 43] was just for demonstration.

3 as the min acceptable length (a letter for the chain name + `-` + a number-letter) is ok.

> Note though that the check is in the current code:

> Current code rejects zero:

Exactly, These are the reasons If regex needs to be refactored, it should be done because of correctness, not optimization!

> Anyway, if you’re ok moving away from regex, I would suggest getting this PR merged and than I can pick up https://github.com/cosmos/ibc-rs/pull/698 again with more validation.

Sorry, but regex is preferred over the [current `split_chain_id`](https://github.com/cosmos/ibc-rs/pull/725/files#diff-0e33ca83458b3683d1a621c10a75653bdc80b52ceff5ea6093669329653a7e06R135-R147). It does not make sense to refactor things again right after this PR. 
If we really want to move away from regex here, I am ok with the discussed function above, which is primarily defined to handle the validation stuff.

Farhad-Shabani:
Closing this, as we ended up fixing the `ChainId` issues all at once in #762, given our current bandwidth and priorities. Really appreciate your contribution :pray:

PR #760, opened 2023-07-12
chore: update some dependencies

Farhad-Shabani:
Closes: #XXX

__

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ci/no-std-check/Cargo.lock
+++ b/ci/no-std-check/Cargo.lock
@@ -506,6 +506,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "darling"
+version = "0.20.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0558d22a7b463ed0241e993f76f09f30b126687447751a8638587b864e4b3944"
+dependencies = [
+ "darling_core",
+ "darling_macro",
+]
+
+[[package]]
+name = "darling_core"
+version = "0.20.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ab8bfa2e259f8ee1ce5e97824a3c55ec4404a0d772ca7fa96bf19f0752a046eb"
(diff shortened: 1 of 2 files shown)


rnbguy:
@Farhad-Shabani, I maybe out of context. So let me ask few questions.

0. Is this a standard `cargo upgrade` PR?
1. Can you please explain why the `time` crate dependency looks like this?

Farhad-Shabani:
> 0. Is this a standard `cargo upgrade` PR?

Yes, I just gathered several into one PR

> 1. Can you please explain why the `time` crate dependency looks like this?

My understanding, dependent projects might be on a different exact version of `time`. By using a version range, we allow updates without forcing projects to upgrade, especially, like here, when only the patch numbers have changed.

rnbguy:
~Can't we directly use `0.3` in that case?~

Ah please ignore my previous reply. I was under the impression that `0.3` is equivalent to `>=0.3.0, <0.4.0`.

But, according to [cargo doc](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html), it's actually equivalent to `>=0.3.0, <1.0.0`.

rnbguy:
Just realized, I was suggesting ~0.3 or 0.3.* (equivalent to >=0.3.0, <0.4.0).

Does it work here?

Farhad-Shabani:
> Just realized, I was suggesting ~0.3 or 0.3.* (equivalent to >=0.3.0, <0.4.0).
> 
> Does it work here?

Yes. Both works, though with a bit differences. 
IMO, the current one provides more information. At a glance, users can see the exact support range. Besides, makes us to review patches (though usually it's a quick one), allowing us to stay informed and support only up to the version we reviewed.

rnbguy (approved):
Explanations looks good. Thanks Farhad !

PR #772, opened 2023-07-17
Json schema for Coin and some parts of Transfer

dzmitry-lahoda:
Closes: #772

## Description

We do coin router kind of thing in CosmWasm. It needs some IBC things, and it requests JsonSchema on types.
So I did it.

Also, I made serde of Amount consistently decimal(decimal string). It was from str decimal, and in coin it was decimal, but on its own if was hex string. So made on its own also decimal. For some it can be breaking change. But other way schema (and CosmWasm messages) would be bad/inconsistent.
______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.
- [x] Tagged *one* reviewer who will be the one responsible for shepherding this PR. may be i guess @Farhad-Shabani 

### Reviewer checklist:


crates/ibc/src/signer.rs
```diff
@@ -2,6 +2,9 @@ use crate::prelude::*;
 
 use derive_more::Display;
 
+#[cfg(feature = "schema")]
+use crate::alloc::borrow::ToOwned;
```
Farhad-Shabani: Isn't this redundant?

.changelog/unreleased/improvement/772-json-schema-for-channels-and-coin.md
```diff
@@ -0,0 +1,2 @@
+- Json schema for Coin and some parts of Transfer
```
Farhad-Shabani: Given the changes to `serde` on `Amount`, let's move this to the `breaking-changes`.

Farhad-Shabani (changes requested):
> We do coin router kind of thing in CosmWasm. It needs some IBC things, and it requests JsonSchema on types.

Great seeing you're using ibc-rs! Thanks for featuring this :)

> Also, I made serde of Amount consistently decimal(decimal string). It was from str decimal, and in coin it was decimal, but on its own if was hex string. So made on its own also decimal. For some it can be breaking change. But other way schema (and CosmWasm messages) would be bad/inconsistent.

Just got a deserialization issue when tested locally. Otherwise, we are good.

- Before, serialization roundtrip (ser > deser) on both `Coin` and `Amount` was ending up with a genuine value. Now, on deserialization, it fails. I think we should add a tailored `deserialize` for `U256` instead of using [this](https://github.com/cosmos/ibc-rs/blob/98099399137b544e66d50af45ecfc8c8e508d0b2/crates/ibc/src/serializers.rs#L29-L38). 
- Also, It would be great if you include a test for it.

crates/ibc/src/applications/transfer/amount.rs
```diff
@@ -6,10 +6,39 @@ use derive_more::{Display, From, Into};
...
+);
+
+#[cfg(feature = "serde")]
+fn deserialize<'de, D>(deserializer: D) -> Result<U256, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    use serde::Deserialize;
+    U256::from_dec_str(<&str>::deserialize(deserializer)?).map_err(serde::de::Error::custom)
+}
```
Farhad-Shabani: Can you place this under the [`seriliazers.rs`](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/serializers.rs#L0-L1) module? Also, `#[cfg(feature = "serde")]` not needed as having it derived on `seriliazers`

crates/ibc/src/applications/transfer/amount.rs
```diff
@@ -6,10 +6,39 @@ use derive_more::{Display, From, Into};
...
+#[test]
+#[cfg(feature = "serde")]
+fn serde_amount() {
+    let value = Amount::from(42);
+    let string = serde_json::to_string(&value).expect("can serde string");
+    assert_eq!(string, "\"42\"");
+    let binary = serde_json::to_vec(&value).expect("can serde binary");
+    let de: Amount = serde_json::from_slice(binary.as_ref()).expect("can deserialize");
+    assert_eq!(de, value);
+}
```
Farhad-Shabani: And move this down to the bottom of the page under `mod tests {}?` like e.g. [here](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/clients/ics07_tendermint/consensus_state.rs#L162-L183)

Farhad-Shabani (changes requested):
Awesome!

A quick nit and then done!

crates/ibc/src/applications/transfer/amount.rs
dzmitry-lahoda: i am not sure about it. any use for U256 from dec string rather than amount?

Farhad-Shabani (approved):
:pray:

crates/ibc/src/applications/transfer/amount.rs
dzmitry-lahoda: until it public, not use for this https://github.com/cosmos/ibc-rs/pull/791#issuecomment-1658313262

PR #789, opened 2023-07-27
Decouple `Router` from `ValidationContext`

### LINKED ISSUE #788: Decouple `Router` from `{Validation,Execution}Context`
Having both the `Router` and `ExecutionContext` be forced to be implemented on the same struct makes us run into issues where we need a `&mut` for the `Router`, and ` &mut` for the `ExecutionContext`, which Rust doesn't allow. Separating them solves this problem.

Related: #786

plafer:
Closes: #788

integration tests: https://github.com/informalsystems/basecoin-rs/pull/118

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/ics04_channel/handler/chan_open_init.rs
```diff
@@ -10,11 +10,12 @@ use crate::core::ics04_channel::events::OpenInit;
...
 use crate::core::ics24_host::identifier::ChannelId;
 use crate::core::ics24_host::path::{ChannelEndPath, SeqAckPath, SeqRecvPath, SeqSendPath};
-use crate::core::router::ModuleId;
+use crate::core::router::{ModuleId, Router};
 use crate::core::{ContextError, ExecutionContext, ValidationContext};
 
 pub(crate) fn chan_open_init_validate<ValCtx>(
     ctx_a: &ValCtx,
+    router_a: &impl Router,
     module_id: ModuleId,
```
Farhad-Shabani: Given the ultimate goal of both the `router_a` and `module_id` here (getting a `Module` instance available for callbacks) and the steps we take in order to retrieve the module (`PortId > ModuleId > dyn Module`), ideally either passing only the router (since the message contains `PortId`) or the object trait `Module` to these handler functions should suffice.
Though, I see the latter way might not be as straightforward to implement because of the dynamic dispatch for modules.

crates/ibc/src/core/handler.rs
```diff
@@ -28,12 +28,17 @@ use super::ics04_channel::handler::timeout::{
...
 use super::{ContextError, ExecutionContext, ValidationContext};
 
 /// Entrypoint which performs both validation and message execution
-pub fn dispatch(ctx: &mut impl ExecutionContext, msg: MsgEnvelope) -> Result<(), RouterError> {
-    validate(ctx, msg.clone())?;
-    execute(ctx, msg)
+pub fn dispatch(
+    ctx: &mut impl ExecutionContext,
+    router: &mut impl Router,
+    msg: MsgEnvelope,
```
Farhad-Shabani, line 39: Here, imo, the signature suggests the entrypoint might accept any combination of `ctx` and `router` (at least it enables this possibility). Basically, each implementation comes only with a single pair of ctx and router.
I liked the separation of the router's role from the context and the way things ended up in Basecoin by defining [`Ibc`](https://github.com/informalsystems/basecoin-rs/blob/ece57a88f24b24f4971970dc9eee09d68e48c9e2/src/modules/ibc/impls.rs#L88-L94), but I am not sure of leaving with this form of entrypoints.
As also, I was thinking, it's likely we'll extract the logger from context as well and have it under that `Ibc`. Then, e.g. should there be a separate argument for the logger too?
Wdyt?

Farhad-Shabani (commented):
This was a needed change! 
Left two comments. They may need a sync call.

crates/ibc/src/core/ics04_channel/handler/chan_open_init.rs
Farhad-Shabani: > Though, I see the latter way might not be as straightforward to implement because of the dynamic dispatch for modules.

Opened this: https://github.com/cosmos/ibc-rs/issues/804

plafer: [7f4c4e3](https://github.com/cosmos/ibc-rs/pull/789/commits/7f4c4e3b7cbe9c872fa68bafa9edb0dd989ab4f9)

crates/ibc/src/core/handler.rs
plafer: As discussed, I don't believe adding yet another abstraction/name to our library will necessarily make it easier to understand. Let's leave it as is for now, and if users are indeed confused, we can consider adding a trait or something to group the router and context together.

Farhad-Shabani (approved):
Thanks Philippe 🙏🏻

PR #790, opened 2023-07-27
fix: restore `no_std` support for `serde` feature

### LINKED ISSUE #741: Compile errors of `v0.41.0` while only `serde` feature enabled
## Problem Summary

During the integration of `ibc-rs v0.41.0` into [near-ibc](https://github.com/octopus-network/near-ibc.git), I got compile errors in the following places:

https://github.com/cosmos/ibc-rs/blob/349d7f5b259e2a34f1f9cbe730c1487e7ecffa38/crates/ibc/src/applications/transfer/packet.rs#L13-L25

https://github.com/cosmos/ibc-rs/blob/349d7f5b259e2a34f1f9cbe730c1487e7ecffa38/crates/ibc/src/core/ics02_client/events.rs#L135-L151

In `near-ibc` project, we only need to enable `borsh` and `serde` features of `ibc-rs`. So I checked source code and found that this is because the definition of `ibc_proto::ibc::applications::transfer::v2::FungibleTokenPacketData` and `ibc_proto::google::protobuf::Any` in `ibc-proto-rs v0.32.0` were defined as:

```rust
#[cfg_attr(feature = "std", derive(::serde::Serialize, ::serde::Deserialize))]
```

I have removed the feature condition from the derive statement for now, which allowed the code to compile. But I'm not sure whether I should commit this issue to `ibc-proto-rs`.

Anyway, I believe it would be worth discussing the issue in the `ibc-rs` implementation. It is important to determine whether the current implementation of the two mentioned places is necessary and appropriate.

### LINKED ISSUE #753: "ics23/serde" add to serde feature, so ibc-rs serde feature can't use on `no-std` env
## Feature Summary

relate: 
- https://github.com/cosmos/ics23/pull/155
- https://github.com/influxdata/pbjson/pull/99

### LINKED ISSUE #838: Switch to domain event type for the `upgrade_client_proposal_handler` return
## Problem Statement

The `upgrade_client_proposal_handler` function currently returns an [event](https://github.com/cosmos/ibc-rs/blob/cd3a266cf4dda173a0c1797d84d984e3b36c3c19/crates/ibc/src/hosts/tendermint/upgrade_proposal/handler.rs#L22) type from the `tendermint_proto` which starting from `tendermint-rs` v0.33.0 will be only compatible with CometBFT v0.38.
The `tendermint-proto` now provides support for multiple versions of CometBFT. So to ensure the handler compatibility with different versions and avoid making assumptions about hosts, the return type should be modified to `tendermint::abci::Event`. This allows for [easy conversion to any version of proto events](https://github.com/informalsystems/tendermint-rs/blob/52f3d8d546331a443c045b8c699a9840eab1a11e/tendermint/src/abci/event.rs#L197-L199), allowing for compatibility with different hosts.

Farhad-Shabani:
Closes: #741
Closes: #753
Closes: #838

Integration tests: https://github.com/informalsystems/basecoin-rs/pull/124

### Remarks
- This PR upgrades `ibc-proto-rs` to v0.34.0 as well.
- There is a side effect (b/c of upgrading `ibc-proto-rs`) to deal with afterward: #835
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


dzmitry-lahoda:
this works well for me

romac (approved):

plafer (approved):

PR #791, opened 2023-07-28
serde as string pub mod

dzmitry-lahoda:
Closes: #791

## Description

Serde as string is useful, Osmosis has impl and CosmWasm has impl and I will copy paste too. 
Serde repo has not such tool impl btw. Mostly hex/bytes things.

So at least one place, which also consistently supports no_std and other serializers and modes would be awesome.  Osmosis and CW-STD do not support no_std.

______

### PR author checklist:
- [x] Added tests. No code changes, no tests.
- [x] Linked to GitHub issue.
- [x] Tagged *one* reviewer who will be the one responsible for shepherding this PR. @Farhad-Shabani as he is already on some serde decision things

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/improvements/791-serde-as-string-pub.md
+++ b/.changelog/unreleased/improvements/791-serde-as-string-pub.md
@@ -1,0 +1,2 @@
+- Made serde as string public
+  ([#777](https://github.com/cosmos/ibc-rs/issues/777))
--- a/crates/ibc/src/lib.rs
+++ b/crates/ibc/src/lib.rs
@@ -63,7 +63,7 @@ mod signer;
 mod utils;
 
 #[cfg(feature = "serde")]
-mod serializers;
+pub mod serializers;
 
 #[cfg(test)]
 mod test;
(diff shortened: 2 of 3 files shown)


plafer:
Hey thank you for submitting this!

However, I don't understand what this achieves. Can you give me an example of something you can do with this change that you couldn't before?

dzmitry-lahoda:
it achieves sharing serde which does not inflicts f32/f64 validation issues

plafer:
Can you give us an example of which functions in the `serializers` module you would use, and for which structs?

Currently, these functions are not documented and only used internally, and if we're going to make them public, we need to understand what problem they solve.

EDIT: If I understand correctly, you would like to use the functions under `serializers::serde_string` to use on your own non-IBC structs?

Farhad-Shabani:
As mentioned [here](https://github.com/cosmos/ibc-rs/pull/791#issuecomment-1665749572), it's best to keep these serializers for internal use only. We don't have sufficient info on potential uses, and can't really gauge the possible consequences outside of our own bubble. By the way, if you happen to use them with any of ibc types, feel free to reopen this!

PR #803, opened 2023-08-01
f32/f64/usize in cosmwasm(and other no f32/f64 envs) reproduction and fixes

### LINKED ISSUE #800: do not use f64/f32/usize as it fails some wasm runtimes (cosmwasm)
## Feature Summary

CosmWasm(wasmd), and hence IBC WASM-08 fail to upload wasm file if it has code with f32/f64.
Unfortunately some things have usize which sometimes handled with floats (crazy but fact https://medium.com/cosmwasm/debugging-floating-point-generation-in-rust-wasm-smart-contract-f47d833b5fba ).

Given these things and that ibc-rs to support many runtime (near, solana, aptos, ic) which also may dislike these floats (as constraint and blockchain enviroments in general) preventing usage of this with linter is good.

Usize, except for some reason some crates use floats to handle it (now sure why), is of variable size, which may lead to bugs, e.g. when target would be wasm64.

## Proposal

Remove all f64/f32/usize usage and lint that from repo.

# LEGACY

## Proposal

I used ibc-rs, and in latest version I get 
```
12:54AM DBG invalid tx code=2 log="failed to execute message; message index: 0: Error calling the VM: Error compiling Wasm: Could not compile: WebAssembly translation error: Error in middleware Gatekeeper: Float operator detected: F64Load { memarg: MemoryImmediate { align: 3, offset: 0, memory: 0 } }. The use of floats is not supported.: create wasm contract failed" module=state
12:54AM INF executed block height=159 module=state num_invalid_txs=1 num_valid_txs=0
```
when it is part of contract.

I see old Cargo.lock 

```
 "tendermint 0.31.1",
 "tendermint-light-client-verifier 0.31.1",
 "tendermint-proto 0.31.1",
 "dyn-clone",
 "erased-serde",
 "ibc-proto 0.30.0",
 "ics23 0.9.0",
 "ibc 0.41.0",
```
new
```
 "tendermint 0.32.2",
 "tendermint-light-client-verifier 0.32.2",
 "tendermint-proto 0.32.2",
 "ics23 0.10.1",
 "ibc-derive 0.1.0",
 "ibc-proto 0.32.0",
 "ics23 0.10.1",
informalsystems-pbjson
```

Also I found `f64` usage in this repo when not needed.

So I do not use any things changed directly, and yet get issues.

Run 
https://github.com/CosmWasm/cosmwasm/blob/main/packages/check/README.md
in CI.

I did not yet found root cause so.

Thing happend from 0.41 release inclusive and  before 0.43. I was able to upload contract.

I think not only cw wasmd will be such, but wasm-08 too.

dzmitry-lahoda:
Closes: #800

## Description

Requires https://github.com/cosmos/ibc-rs/pull/790

1. replaced f64 logic with equivalent u64
2. forbidden use of f64 (and usize)
3. usize was replaced by u64, assuming that wasm64 is nearest highest target for mem size. i am ok with safety of thing, so u32/u16 could be better per  case (mostly length on names and some history things - sure never u64, but u64 is correct map to usize), so no ideal but working.
4. replaced weird serde issues (both serde and serde_wasm compile into wasm f64 even if do just strings, had to rip them as could)
5. added test to test changes (with these there is no validation error)
6. side change added schema for ChainId (i use in CW) 

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


dzmitry-lahoda:
feel free to merge with https://github.com/cosmos/ibc-rs/pull/790 and run insuction in shell.nix header. try to rollback any change and see effects. serde works exactly only this way i found, other compile with floats

dzmitry-lahoda:
for usize correctness parity uses static assets 

substrate/primitives/runtime-interface/src/impls.rs:45:1

so can asset some usize if needed too

dzmitry-lahoda:
pr has issue

 "serde-json-wasm 0.5.1 (git+https://github.com/dzmitry-lahoda-forks/serde-json-wasm?rev=8a7e522c0e4de36a6dfb535766f26a9941017d81)",

need to use no-std serde-json-wasm

dzmitry-lahoda:
https://github.com/CosmWasm/serde-json-wasm/pull/58

https://github.com/CosmWasm/serde-json-wasm/pull/56

we trying to get no_std merged for 1 year+ already so

Farhad-Shabani (changes requested):
Hi @dzmitry-lahoda. This is awesome and much appreciated! 👌🏻 
Let’s get it over the finish line. Would you plz apply the following changes:
1. Merge changes from the main to the PR's branch
2. Set `serde-json-wasm` package to use v1.0.0
3. Add a CI job that does `cw_check` by running the `shell.nix`
4. Add unclog

Then, I will run an integration test against gaia chains in [`basecoin-rs`](https://github.com/informalsystems/basecoin-rs) to make sure using `serde-json-wasm` we still keep compatibility with both `Hermes` and `basecoin_app` (Though, from what I see most probably should be ok)

Let me know if for any of above, I can assist

Farhad-Shabani:
Hey @dzmitry-lahoda, Any chance you could wrap up this PR? 
I also can assist with this if you select "_Allow edits from maintainers_".

crates/ibc/src/applications/transfer/amount.rs
```diff
@@ -15,7 +15,7 @@ use crate::alloc::{borrow::ToOwned, string::String};
 #[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Display, From, Into)]
 pub struct Amount(
     #[cfg_attr(feature = "schema", schemars(with = "String"))]
-    #[serde(serialize_with = "crate::serializers::serde_string::serialize")]
+    #[serde(serialize_with = "serialize")]
```
Farhad-Shabani, line 18: Taking another peek, the "serialize" function in "amount.rs" seems redundant. Why not use "crate::serializers::serde_string::serialize"?

dzmitry-lahoda:
hello, some things 
1. I cannot allow edit because i have -forks org I do all forks (to avoid mix of my projects), and there is bug https://github.com/orgs/community/discussions/5634 . I cannot set the value. So I am ok with copy paste or else.
2. https://github.com/CosmWasm/serde-json-wasm was released version 1.0. they made it no_std afaid and fixed some issue - so this one should be used.
3. and final one, latest version of CosmWasm wasmd emulated floats (software implementation I guess). But I am not sure it works in no_std or without using their wasmd wasmer host or without using special assembly rewrite hook. So float less wasm still a thing, so for some people may be less prio. 

but in general things I tried still holds, for next 1 month i am on ibc for solidity, so will not catch up this minimum this period, i guess company will continue more efforts to no_std versions of ibc in 2-3 month (after evm).

crates/ibc/src/applications/transfer/amount.rs
dzmitry-lahoda: afaik for some reason it triggered `float` to be compiled into wasm (seems like just patter matching, but validation tools do not care).

Farhad-Shabani:
Thanks a bunch for the tips and updates. :pray: No worries, I’ll fork yours and open up a new PR.
Just, as for the 3rd point, given another issue we've got [here](https://github.com/cosmos/ibc-rs/issues/850), still we should move away from floats.
 
Looking forward to hearing about the progress on the no_std versions of IBC.

Farhad-Shabani:
Superseded by #894

crates/ibc/src/applications/transfer/amount.rs
Farhad-Shabani: Would be OK if the `ci/cw-check` passes with the `crate::serializers::serde_string::serialize` or still might be the case?

dzmitry-lahoda: would be OK

dzmitry-lahoda: i used only reproducible check to fix issues

PR #806, opened 2023-08-01
imp(Router)!: simplify `Module` lookup

### LINKED ISSUE #802: Simplify Module lookup in the `Router` trait
## Summary

There is only one way to implement [`lookup_module_channel`](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/core/router.rs#L31-L46) and [`lookup_module_packet`](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/core/router.rs#L48-L61) methods within the `Router` trait. Their main purpose is to extract `PortId` from `ChannelMsg` and `PacketMsg`.
We should consider removing them and, instead, introduce a free mapping function for each of the `ChannelMsg` and `PacketMsg` that extracts `PortId`. Following this, the existing method [`lookup_module_by_port`](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/core/router.rs#L29) can be used everywhere for subsequent module lookup operation.

Farhad-Shabani:
Closes: #802
integration tests: https://github.com/informalsystems/basecoin-rs/pull/121
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/core/router.rs
+++ b/crates/ibc/src/core/router.rs
@@ -17,10 +17,13 @@ use crate::signer::Signer;
 /// Router as defined in ICS-26, which binds modules to ports.
 pub trait Router {
     /// Returns a reference to a `Module` registered against the specified `ModuleId`
-    fn get_route(&self, port_id: &PortId) -> Option<&dyn Module>;
+    fn get_route(&self, module_id: &ModuleId) -> Option<&dyn Module>;
 
     /// Returns a mutable reference to a `Module` registered against the specified `ModuleId`
-    fn get_route_mut(&mut self, port_id: &PortId) -> Option<&mut dyn Module>;
+    fn get_route_mut(&mut self, module_id: &ModuleId) -> Option<&mut dyn Module>;
+
+    /// Return the module_id associated with a given port_id
+    fn lookup_module(&self, port_id: &PortId) -> Option<ModuleId>;
 }
 
 /// Module name, internal to the chain.
--- a/crates/ibc/src/core/router.rs
+++ b/crates/ibc/src/core/router.rs
@@ -7,10 +7,7 @@ use core::fmt::{Debug, Display, Error as FmtError, Formatter};
 use crate::core::events::ModuleEvent;
 use crate::core::ics04_channel::acknowledgement::Acknowledgement;
 use crate::core::ics04_channel::channel::{Counterparty, Order};
-use crate::core::ics04_channel::error::PortError::UnknownPort;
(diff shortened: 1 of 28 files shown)


crates/ibc/src/core/router.rs
```diff
@@ -26,39 +23,7 @@ pub trait Router {
...
-            PacketMsg::TimeoutOnClose(msg) => &msg.packet.port_id_on_a,
-        };
-        let module_id = self
-            .lookup_module_by_port(port_id)
-            .ok_or(ChannelError::Port(UnknownPort {
-                port_id: port_id.clone(),
-            }))?;
-        Ok(module_id)
-    }
+    fn lookup_module(&self, port_id: &PortId) -> Option<ModuleId>;
```
plafer, line 26: While we're at it, do we even need the `ModuleId`? Does it provide any value? We could simply have

```rust
pub trait Router {
    fn get_route(&self, port_id: &PortId) -> Option<&dyn Module>;
    fn get_route_mut(&mut self, port_id: &PortId) -> Option<&mut dyn Module>;
}
```

plafer (commented):
Love the simplification; suggested even more simplification

crates/ibc/src/core/router.rs
Farhad-Shabani: Good Point!
The ~only noteworthy benefit rn~ is more efficiency in searching for modules compared to directly using `port_id`. 
Ran a local test on `MockContext`... With `ModuleId` it can be up to ~10 times faster at times. HW should say we are talking of ~1 microsecond.
~Anyway, for now, I dropped `ModuleId` ([9227ef7](https://github.com/cosmos/ibc-rs/pull/806/commits/9227ef7557aa8437b4eec5dc6fd1311fa1af0333)). It can be added at any time if it would really matter.~

Farhad-Shabani: Ahhh, I just recalled something. Reverted the above commit!
This `ModuleId` thing is essential for apps like ics27. Upon registering each interchain account on the controller chain a new `PortId` will be allocated at a certain height. As a result, there could be a bunch of `PortId`s instantiated. This `module_lookup` efficiently maps them to a single module instance. Otherwise, for each of those `PortId`s, a separate redundant module instance should be stored.

plafer (approved):

crates/ibc/src/core/router.rs
plafer: Great, let's not do that then!

PR #845, opened 2023-08-30
Add borsh derive for `MsgTransfer`

plafer:


crates/ibc/src/applications/transfer/amount.rs
```diff
@@ -28,6 +29,54 @@ impl parity_scale_codec::WrapperTypeDecode for Amount {
...
+        const NUM_WORDS_IN_U256: usize = 4;
+
+        let mut buf = [0; 32];
+        let bytes_read = reader.read(&mut buf)?;
+        if bytes_read != 32 {
+            return Err(borsh::maybestd::io::Error::new(
+                borsh::maybestd::io::ErrorKind::InvalidInput,
+                format!("Expected to read 32 bytes, read {bytes_read}"),
+            ));
+        }
```
Farhad-Shabani, line 60: Since the `buf`` is of fixed 32 size, would this error ever occur?

plafer: Yes, if somehow we read less than 32 bytes

Farhad-Shabani (approved):

PR #876, opened 2023-09-22
Remove regex requirement

### LINKED ISSUE #875: docs.rs build is failing
## Bug Summary

Failed docs.rs build: https://docs.rs/crate/ibc/0.45.0/builds/917198

```
[INFO] [stderr] error[E0422]: cannot find struct, variant or union type `LineColumn` in crate `proc_macro`
[INFO] [stderr]    --> /opt/rustwide/cargo-home/registry/src/index.crates.io-6f17d22bba15001f/safe-proc-macro2-1.0.36/src/wrapper.rs:485:33
[INFO] [stderr]     |
[INFO] [stderr] 485 |                 let proc_macro::LineColumn { line, column } = s.start();
[INFO] [stderr]     |                                 ^^^^^^^^^^ not found in `proc_macro`
[INFO] [stderr]     |
[INFO] [stderr] help: consider importing one of these items
[INFO] [stderr]     |
[INFO] [stderr] 1   + use crate::LineColumn;
[INFO] [stderr]     |
[INFO] [stderr] 1   + use crate::fallback::LineColumn;
[INFO] [stderr]     |
[INFO] [stderr] help: if you import `LineColumn`, refer to it directly
[INFO] [stderr]     |
[INFO] [stderr] 485 -                 let proc_macro::LineColumn { line, column } = s.start();
[INFO] [stderr] 485 +                 let LineColumn { line, column } = s.start();
[INFO] [stderr]     |
[INFO] [stderr] 
```

## Details

Rust [removed](https://github.com/rust-lang/rust/pull/111571) `proc_macro::LineColumn` which is being used in `safe-proc-macro2` (dependency to `safe-regex`) under `span-locations` feature.

Probably docsrs is building the dependencies with `--all-features` which breaks the build in recent nightlies.

The fix requires an update from `safe-proc-macro2` > `safe-regex`.

## Version

`0.42.0`-`0.45.0`

rnbguy:
Closes: #875

## Description

- Removes any _regex_ requirement
- Removes `safe_regex` which is breaking our docsrs builds

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/src/applications/transfer/coin.rs
+++ b/crates/ibc/src/applications/transfer/coin.rs
@@ -176,6 +176,7 @@ mod tests {
     #[case::tricky("123stake,1a1-999den0m", &[RawCoin::new(123, "stake"), RawCoin::new(1, "a1-999den0m")])]
     #[case::colon_delimiter("123stake:1a1:999den0m", &[RawCoin::new(123, "stake:1a1:999den0m")])]
     #[case::dash_delimiter("123stake-1a1-999den0m", &[RawCoin::new(123, "stake-1a1-999den0m")])]
+    #[case::slash_delimiter("123stake/1a1/999den0m", &[RawCoin::new(123, "stake/1a1/999den0m")])]
     fn test_parse_raw_coin_list(
         #[case] coins_str: &str,
         #[case] coins: &[RawCoin],
--- a/crates/ibc/src/applications/transfer/coin.rs
+++ b/crates/ibc/src/applications/transfer/coin.rs
@@ -80,7 +80,7 @@ where
             .filter(|(amount, denom)| !amount.is_empty() && !denom.is_empty())
             .filter(|(_, denom)| {
                 !denom.contains(|x| {
-                    !matches!(x, 'a'..='z' | 'A'..='Z' | '0'..='9' | '/' | ':' | '\\' | '.' | '_' | '-')
+                    !matches!(x, 'a'..='z' | 'A'..='Z' | '0'..='9' | '/' | ':' | '.' | '_' | '-')
                 })
             })
             .ok_or_else(|| TokenTransferError::InvalidCoin {
--- a/crates/ibc/src/applications/transfer/coin.rs
+++ b/crates/ibc/src/applications/transfer/coin.rs
@@ -59,23 +59,25 @@ where
         // a letter, a number or a separator ('/', ':', '.', '_' or '-').
         // Loosely copy the regex from here:
         // https://github.com/cosmos/cosmos-sdk/blob/v0.45.5/types/coin.go#L760-L762
-        let re = Regex::new(r"^([0-9]+)([a-zA-Z0-9/:\\._-]+)$")
-            .map_err(TokenTransferError::InvalidRegex)?;
+        // old regex: ^(?<amount>[0-9]+)(?<denom>[a-zA-Z0-9/:\\._-]+)$
 
-        let (amount_str, denom_str) = re
-            .captures(coin_str)
-            .and_then(|cap| {
-                let amount_str = cap.get(1)?;
-                let denom_str = cap.get(2)?;
-                Some((amount_str.as_str(), denom_str.as_str()))
+        let (amount, denom) = coin_str
+            .chars()
+            .position(|x| !x.is_numeric())
+            .map(|index| coin_str.split_at(index))
+            .filter(|(amount, denom)| amount.is_empty() || denom.is_empty())
+            .filter(|(_, denom)| {
+                denom.contains(|x| {
+                    !matches!(x, 'a'..='z' | 'A'..='Z' | '0'..='9' | '/' | ':' | '\\' | '.' | '_' | '-')
+                })
             })
             .ok_or_else(|| TokenTransferError::InvalidCoin {
                 coin: coin_str.to_string(),
             })?;
 
         Ok(Coin {
-            amount: amount_str.parse()?,
-            denom: denom_str.parse().map_err(Into::into)?,
+            amount: amount.parse()?,
+            denom: denom.parse().map_err(Into::into)?,
         })
     }
 }
--- a/crates/ibc/src/applications/transfer/coin.rs
+++ b/crates/ibc/src/applications/transfer/coin.rs
@@ -64,9 +64,9 @@ where
             .chars()
             .position(|x| !x.is_numeric())
             .map(|index| coin_str.split_at(index))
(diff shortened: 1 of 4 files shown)


crates/ibc/src/applications/transfer/coin.rs
```diff
@@ -59,24 +58,26 @@ where
...
-        let (m1, m2) = matcher.match_slices(coin_str.as_bytes()).ok_or_else(|| {
-            TokenTransferError::InvalidCoin {
+        // old regex: ^(?<amount>[0-9]+)(?<denom>[a-zA-Z0-9/:\\._-]+)$
+
+        let (amount, denom) = coin_str
+            .chars()
+            .position(|x| !x.is_numeric())
+            .map(|index| coin_str.split_at(index))
+            .filter(|(amount, denom)| !amount.is_empty() && !denom.is_empty())
+            .filter(|(_, denom)| {
```
soareschen, line 75: I think filter means that you will exclude the invalid characters, but the result would be valid. In the original case, invalid characters will cause the method to fail.

crates/ibc/src/applications/transfer/coin.rs
```diff
@@ -59,24 +58,26 @@ where
...
+                !denom.contains(|x| {
+                    !matches!(x, 'a'..='z' | 'A'..='Z' | '0'..='9' | '/' | ':' | '\\' | '.' | '_' | '-')
```
soareschen: I think `\\` was used as an escape for `\`, which is an escape for `.` in the original regex.

soareschen (changes requested):
We need to write tests to ensure that the new implementation works as expected.

rnbguy:
> We need to write tests to ensure that the new implementation works as expected.

Yea. Tests are coming :slightly_smiling_face: I wanted to confirm the parsing logic before that.

crates/ibc/src/applications/transfer/coin.rs
```diff
@@ -38,6 +37,19 @@ pub struct Coin<D> {
     pub amount: Amount,
 }
 
+impl<D> Coin<D> {
+    pub fn new<T, DT>(amount: T, denom: DT) -> Self
+    where
+        T: Into<Amount>,
+        DT: Into<D>,
```
Farhad-Shabani: A note regarding these bounds, as this method would get available to our users too (not only for the tests)

Although they've made improvements in cleaning up test arguments, and at first glance, it seems to give users more flexibility, it's important to note that there are situations where both `amount` and `denom` might come from either parsing a string or using `TryFrom`. This means not all conversions can simply turn into `Amount` and `D` so users may obtain them from conversions that might fail.

In fact, we even shouldn’t use `TryFrom` or similar restrictions in this specific situation. As particularly `Coin` is generic over denoms and We can't foresee how hosts will obtain this data. So any restrictions will assume a certain approach, which may not always work. Also, e.g. in cases like [here](https://github.com/cosmos/ibc-rs/blob/1aa853bb646d37aab3848096b4221450f3a2e760/crates/ibc/src/applications/transfer/coin.rs#L94-L97) and [here](https://github.com/cosmos/ibc-rs/blob/dc8e4161dd5e00a2031cf3feb58dad1ed7e26bdb/crates/ibc/src/applications/transfer/packet.rs#L46), it is easier to work with the own types.

Btw, if still want to keep our tests args as is, Perhaps we should move this method (as a helper) under the `tests` module and only have it for the `RawCoin` type. Wdyt?

crates/ibc/src/applications/transfer/coin.rs
```diff
@@ -59,24 +71,30 @@ where
         // a letter, a number or a separator ('/', ':', '.', '_' or '-').
         // Loosely copy the regex from here:
         // https://github.com/cosmos/cosmos-sdk/blob/v0.45.5/types/coin.go#L760-L762
```
Farhad-Shabani:
```suggestion
        // https://github.com/cosmos/cosmos-sdk/blob/v0.47.5/types/coin.go#L838-L840
```

crates/ibc/src/applications/transfer/coin.rs
```diff
@@ -119,53 +137,59 @@ impl<D: Display> Display for Coin<D> {
...
-            assert_eq!(coin.amount, 0u64.into());
-        }
+    use super::*;
 
-        {
-            // `!` is not allowed
-            let res = RawCoin::from_str("0x!");
-            assert!(res.is_err());
-        }
+    #[rstest]
```
Farhad-Shabani, line 135: Good to include a test for checking lower/uppers case sensitivity.

crates/ibc/src/applications/transfer/coin.rs
```diff
@@ -59,24 +71,30 @@ where
...
+        let (amount, denom) = coin_str
+            .chars()
+            .position(|x| !x.is_numeric())
+            .map(|index| coin_str.split_at(index))
+            .filter(|(amount, denom)| !amount.is_empty() && !denom.is_empty())
+            .filter(|(_, denom)| {
+                denom.chars().all(|x| {
+                    matches!(x, 'a'..='z' | 'A'..='Z' | '0'..='9' | '/' | ':' | '.' | '_' | '-')
+                })
+            })
```
Farhad-Shabani: Here we can simplify the implementation and use the same approach as [here](https://github.com/cosmos/ibc-rs/blob/dc8e4161dd5e00a2031cf3feb58dad1ed7e26bdb/crates/ibc/src/core/ics24_host/identifier/validate.rs#L22-L31) by defnining e.g. `VALID_DENOM_CHARACTERS` , after which the `!denom.is_empty()` check won’t be needed anymore too. Even we might be able to decouple checks and filter of the amount from the denom. Appreciate a further look on this.

Something like this:

```rust
const VALID_DENOM_CHARACTERS: &str = "/:._-";

let (amount, denom) = coin_str
            .chars()
            .position(|x| !x.is_numeric())
            .map(|index| coin_str.split_at(index))
            .filter(|(amount, _)| !amount.is_empty())
            .filter(|(_, denom)| {
                denom
                    .chars()
                    .all(|x| x.is_alphanumeric() || VALID_DENOM_CHARACTERS.contains(x))
            })
            .ok_or_else(|| TokenTransferError::InvalidCoin {
                coin: coin_str.to_string(),
            })?;
```

Farhad-Shabani (changes requested):

Farhad-Shabani (approved):
Good job @rnbguy. Thanks 👌🏻

soareschen (approved):

PR #916, opened 2023-10-12
Release v0.46.0

Farhad-Shabani:
Closes: #XXX

______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc-query/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name         = "ibc-query"
-version      = "0.45.0"
+version      = "0.46.0"
```
rnbguy, line 3: Question. How are we distributing `ibc-query` package? If it is a separate package, do we still need to keep the same version number as `ibc`? (e.g. `ibc-derive` has different version number)

rnbguy (commented):
Looks good. Just a nit request. Is it okay to make the `.toml` files ending with newlines?

crates/ibc-query/Cargo.toml
Farhad-Shabani: Good question!

More of a common practice, and in a simplistic view `ibc-query` is dependent on `ibc` and this approach ensures that all dependencies are kept in sync. Also, for an easier map version for users.

As for the `ibc-derive` things are a bit different. That’s not directly dependent on `ibc` crate. we re-export everything in `ibc-derive`, so projects won’t need to explicitly have it as a dependency. Also, `ibc-derive` only needs to get updated if there are only changes to `ClientState` APIs.

rnbguy: Sounds good to me. Just wanted to make sure you're aware of this. Thanks :raised_hands:

rnbguy (approved):
Thanks @Farhad-Shabani :tada:

PR #921, opened 2023-10-13
test: tendermint client update when client expires or validator set has changed

### LINKED ISSUE #538: [ICS02] Cover client frozen/expired conditions with unit tests
## Problem Statement

Make sure that we covered failure scenarios everywhere we check a status of a client (like [here](https://github.com/cosmos/ibc-rs/blob/45d5f0d25f323eb423d5511e2488be8104e1c39f/crates/ibc/src/core/ics02_client/handler/misbehaviour.rs#L27-L29)) by introducing frozen and expired conditions. 
Adding required fields to the `MockCotext` and writing unit tests should accomplish this.

Should come after fixing https://github.com/cosmos/ibc-rs/issues/536

rnbguy:
Closes: #538 

## Description

This PR adds and modifies a few test scenarios.

1. Client expiry while client state update.
2. Client state update with validator set changes (happy and sad).

______

### PR author checklist:
- [ ] Create a new issue for parameterized mock Header, Client, and Context configuration.
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc/Cargo.toml
+++ b/crates/ibc/Cargo.toml
@@ -73,7 +73,7 @@ scale-info = { version = "2.1.2", default-features = false, features = ["derive"
 ## for borsh encode or decode
 borsh = {version = "0.10", default-features = false, optional = true }
 parking_lot = { version = "0.12.1", default-features = false, optional = true }
-typed-builder = { version = "0.16.2", optional = true }
+typed-builder = { version = "0.17.0", optional = true }
 
 ibc-derive = { version ="0.3.0", path = "../ibc-derive" }
 
@@ -105,4 +105,4 @@ test-log = { version = "0.2.10", features = ["trace"] }
 tendermint-rpc = { version = "0.34", features = ["http-client", "websocket-client"] }
 tendermint-testgen = { version = "0.34" } # Needed for generating (synthetic) light blocks.
 parking_lot = { version = "0.12.1" }
-typed-builder = { version = "0.16.2" }
+typed-builder = { version = "0.17.0" }
--- a/crates/ibc/Cargo.toml
+++ b/crates/ibc/Cargo.toml
@@ -103,6 +103,4 @@ rstest = "0.18.1"
 tracing-subscriber = { version = "0.3.14", features = ["fmt", "env-filter", "json"]}
 test-log = { version = "0.2.10", features = ["trace"] }
 tendermint-rpc = { version = "0.34", features = ["http-client", "websocket-client"] }
-tendermint-testgen = { version = "0.34" } # Needed for generating (synthetic) light blocks.
-parking_lot = { version = "0.12.1" }
-typed-builder = { version = "0.17.0" }
+ibc = { path = ".", features = ["mocks"] }
--- a/crates/ibc/Cargo.toml
+++ b/crates/ibc/Cargo.toml
@@ -46,7 +46,7 @@ schema = ["dep:schemars", "serde", "std"]
 
 # This feature grants access to development-time mocking libraries, such as `MockContext` or `MockHeader`.
 # Depends on the `testgen` suite for generating Tendermint light blocks.
-mocks = ["tendermint-testgen", "tendermint/clock", "parking_lot"]
+mocks = ["tendermint-testgen", "tendermint/clock", "parking_lot", "typed-builder"]
 
 [dependencies]
 # Proto definitions for all IBC-related interfaces, e.g., connections or channels.
@@ -73,6 +73,7 @@ scale-info = { version = "2.1.2", default-features = false, features = ["derive"
 ## for borsh encode or decode
 borsh = {version = "0.10", default-features = false, optional = true }
 parking_lot = { version = "0.12.1", default-features = false, optional = true }
+typed-builder = { version = "0.16.2", optional = true }
 
 ibc-derive = { version ="0.3.0", path = "../ibc-derive" }
 
@@ -104,3 +105,4 @@ test-log = { version = "0.2.10", features = ["trace"] }
 tendermint-rpc = { version = "0.34", features = ["http-client", "websocket-client"] }
 tendermint-testgen = { version = "0.34" } # Needed for generating (synthetic) light blocks.
 parking_lot = { version = "0.12.1" }
+typed-builder = { version = "0.16.2" }
--- a/crates/ibc/src/core/ics02_client/handler/update_client.rs
+++ b/crates/ibc/src/core/ics02_client/handler/update_client.rs
@@ -869,62 +869,60 @@ mod tests {
 
     #[test]
     fn test_expired_client() {
-        let remote_chain_id = ChainId::new("mockgaiaB", 1).unwrap();
-        let host_chain_id = ChainId::new("mockgaiaA", 1).unwrap();
+        let chain_id_a = ChainId::new("mockgaiaA", 1).unwrap();
+        let chain_id_b = ChainId::new("mockgaiaB", 1).unwrap();
 
-        let remote_height = Height::new(1, 21).unwrap();
-        let remote_client_height_on_host = remote_height.sub(3).unwrap();
+        let update_height = Height::new(1, 21).unwrap();
+        let client_height = update_height.sub(3).unwrap();
 
-        let remote_client_id_on_host = ClientId::new(tm_client_type(), 0).unwrap();
+        let client_id_b = ClientId::new(tm_client_type(), 0).unwrap();
 
         let timestamp = Timestamp::now();
 
-        let mut host_ctx = MockContextConfig::builder()
-            .host_id(host_chain_id.clone())
+        let mut ctx_a = MockContextConfig::builder()
+            .host_id(chain_id_a.clone())
             .latest_height(Height::new(1, 1).unwrap())
             .latest_timestamp(timestamp)
             .build()
             .with_client_config(
                 MockClientConfig::builder()
-                    .client_chain_id(remote_chain_id.clone())
-                    .client_id(remote_client_id_on_host.clone())
-                    .client_state_height(remote_client_height_on_host)
+                    .client_chain_id(chain_id_b.clone())
+                    .client_id(client_id_b.clone())
+                    .client_state_height(client_height)
                     .client_type(tm_client_type())
                     .latest_timestamp(timestamp)
                     .build(),
             );
 
-        let mut remote_ctx = MockContextConfig::builder()
-            .host_id(remote_chain_id.clone())
+        let mut ctx_b = MockContextConfig::builder()
+            .host_id(chain_id_b.clone())
             .host_type(HostType::SyntheticTendermint)
-            .latest_height(remote_height)
+            .latest_height(update_height)
             .latest_timestamp(timestamp)
             .build();
 
         {
-            let remote_raw_client_state = host_ctx.client_state(&remote_client_id_on_host).unwrap();
-
-            let remote_client_state = match remote_raw_client_state {
+            let client_state = match ctx_a.client_state(&client_id_b).unwrap() {
                 AnyClientState::Tendermint(tm_client_state) => tm_client_state,
                 _ => panic!("never fails. not mock client"),
             };
 
-            let client_trusting_period = remote_client_state.trusting_period;
+            let client_trusting_period = client_state.trusting_period;
 
-            let future_timestamp = host_ctx
+            let future_timestamp = ctx_a
                 .host_timestamp()
                 .expect("never fails")
                 .add(client_trusting_period)
                 .expect("overflow");
 
-            host_ctx.advance_host_chain_height_with_timestamp(future_timestamp);
-            remote_ctx.advance_host_chain_height_with_timestamp(future_timestamp);
+            ctx_a.advance_host_chain_height_with_timestamp(future_timestamp);
+            ctx_b.advance_host_chain_height_with_timestamp(future_timestamp);
         }
 
-        let remote_client_state = host_ctx.client_state(&remote_client_id_on_host).unwrap();
+        let client_state = ctx_a.client_state(&client_id_b).unwrap();
 
-        assert!(remote_client_state
-            .status(&host_ctx, &remote_client_id_on_host)
+        assert!(client_state
+            .status(&ctx_a, &client_id_b)
             .unwrap()
             .is_expired());
     }
--- a/crates/ibc/src/core/ics02_client/handler/update_client.rs
+++ b/crates/ibc/src/core/ics02_client/handler/update_client.rs
@@ -254,23 +254,24 @@ mod tests {
         let client_height = Height::new(1, 20).unwrap();
         let chain_id_b = ChainId::new("mockgaiaB", 1).unwrap();
 
-        let mut ctx = MockContext::new(
-            ChainId::new("mockgaiaA", 1).unwrap(),
-            HostType::Mock,
-            5,
-            Height::new(1, 1).unwrap(),
-        )
-        .with_client_parametrized_with_chain_id(
-            // client state initialized with client_height, and
-            // [{id: 1, power: 50}, {id: 2, power: 50}] for validator set and next validator set.
-            chain_id_b.clone(),
-            &client_id,
-            client_height,
-            Some(tm_client_type()), // The target host chain (B) is synthetic TM.
-            Some(client_height),
-        );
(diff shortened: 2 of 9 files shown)


rnbguy:
I am not sure what is going on. The earlier commits failed on gh action, although they were successful on my machine (`cargo test --all-features --no-run` and `act`). I added 0f4503c as a fix - even though with it, the build fails on my machine - but it makes gh actions successful!

@Farhad-Shabani can you please take a look later?

rnbguy:
Did I hit a rare issue with Git?

This is from this PR
https://github.com/cosmos/ibc-rs/blob/0f4503ce85e042efaf89063989d81afe01d23706/crates/ibc/src/mock/context.rs#L240

This is from `main` branch
https://github.com/cosmos/ibc-rs/blob/245d2fc3d53d7b3c9eb6d2380580c798aa150fe9/crates/ibc/src/mock/context.rs#L203

Although it's shown when two commits are [compared](https://github.com/cosmos/ibc-rs/compare/245d2fc3d53d7b3c9eb6d2380580c798aa150fe9..0f4503ce85e042efaf89063989d81afe01d23706).

rnbguy:
~Please ignore my previous comments.~ I just realized the `main` branch progressed from what I had.

The new changes did not conflict with my PR syntactically but they [conflicted semantically](https://bors.tech/essay/2017/02/02/pitch). :sweat_smile:

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -249,47 +252,74 @@ mod tests {
...
-                vec![
-                    TestgenValidator::new("1").voting_power(60),
-                    TestgenValidator::new("2").voting_power(40),
-                ],
+        let ctx_b_val_history = vec![
+            // TODO(rano): the validator set params during setups.
+            // Here I picked the default validator set which is
+            // used at host side client creation.
+            //
+            // validator set of height-20
```
Farhad-Shabani: On a general note on such TODO’s annotation, let's keep the comments as clean as we can and instead if there is something we should really get to it later, let’s open an issue (not necessarily a detailed one). This way can better track our tech debts as well. Thanks!

crates/ibc/src/mock/context.rs
```diff
@@ -554,6 +676,81 @@ impl MockContext {
...
+                                client
+                                    .latest_timestamp
+                                    .sub(self.block_time * ((n_blocks - 1 - i) as u32))
+                                    .expect("never fails"),
+                            ),
+                        )
+                    })
+                    .collect();
+
+                // TODO(rano): parametrize the other params
```
Farhad-Shabani: Same comment on `TODO`s applies here as well.

crates/ibc/src/mock/host.rs
```diff
@@ -141,15 +157,27 @@ impl HostBlock {
...
         timestamp: Timestamp,
     ) -> SyntheticTmBlock {
-        let light_block = TestgenLightBlock::new_default_with_time_and_chain_id(
-            chain_id.to_string(),
-            timestamp.into_tm_time().expect("Never fails"),
-            height,
-        )
-        .generate()
-        .expect("Never fails");
+        // TODO(rano): use HeaderBuilder pattern
```
Farhad-Shabani: Same comment on `TODO`s applies here as well.

crates/ibc/src/mock/host.rs
```diff
@@ -141,15 +157,27 @@ impl HostBlock {
...
+            .chain_id(chain_id.as_str())
+            .next_validators(&validators)
+            .time(timestamp.into_tm_time().expect("Never fails"));
+
+        let light_block = TestgenLightBlock::new_default_with_header(header)
+            .generate()
+            .expect("Never fails");
+
         SyntheticTmBlock {
+            // TODO(rano): don't use constant for rev_height
```
Farhad-Shabani: Same comment on `TODO`s applies here as well.

crates/ibc/src/mock/host.rs
```diff
@@ -38,6 +39,7 @@ pub enum HostType {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct SyntheticTmBlock {
     pub trusted_height: Height,
+    pub trusted_next_validators: ValidatorSet,
```
Farhad-Shabani, line 42: Do we have to add this `trusted_next_validators` as a new field here? I mean, we can already get to the same stuff through the `light_block` field. Right?

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -249,47 +252,74 @@ mod tests {
...
-        )
-        .with_client_parametrized_with_chain_id(
-            chain_id_b.clone(),
-            &client_id,
-            client_height,
-            Some(tm_client_type()), // The target host chain (B) is synthetic TM.
-            Some(client_height),
-        );
+        let mut ctx = MockContextConfig::builder()
+            .host_id(ChainId::new("mockgaiaA", 1).unwrap())
```
Farhad-Shabani, line 257: `ctx_a` instead of `ctx` ?

crates/ibc/src/mock/context.rs
```diff
@@ -554,6 +676,81 @@ impl MockContext {
...
+                client_state.validate().expect("never fails");
+
+                let cs_states = blocks
+                    .into_iter()
+                    .map(|(height, block)| (height, block.into()))
+                    .collect();
+
+                (client_state.into(), cs_states)
+            }
+            _ => todo!(),
```
Farhad-Shabani: Should panic here instead of `todo!()` ?

crates/ibc/src/mock/context/clients.rs
```diff
@@ -1,19 +1,63 @@
...
+            config.trusting_period,
+            config.unbonding_period,
+            config.max_clock_drift,
+            config.latest_height,
+            config.proof_specs,
+            config.upgrade_path,
+            config.allow_update,
+        )
+    }
+}
```
Farhad-Shabani: Here, we should host the code related to the `mock` and implementing contexts on it. 
`TmClientStateConfig` should be placed under the `ics07_tendermint` and enabled using the `mocks` feature flag (along with others (https://github.com/cosmos/ibc-rs/blob/245d2fc3d53d7b3c9eb6d2380580c798aa150fe9/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L1122-L1183))

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -311,6 +341,89 @@ mod tests {
...
+    fn test_update_synthetic_tendermint_client_validator_change_fail() {
+        let client_id = ClientId::new(tm_client_type(), 0).unwrap();
+        let client_height = Height::new(1, 20).unwrap();
+        let chain_id_b = ChainId::new("mockgaiaB", 1).unwrap();
+
+        let ctx = MockContextConfig::builder()
+            .host_id(ChainId::new("mockgaiaA", 1).unwrap())
+            .host_type(HostType::Mock)
+            .latest_height(Height::new(1, 1).unwrap())
+            .max_history_size(5)
```
Farhad-Shabani: We can remove `.host_type()` and `.max_history_size()` given that we have access to these through default derivations.

crates/ibc/src/core/ics02_client/handler/update_client.rs
```diff
@@ -753,4 +866,66 @@ mod tests {
         assert!(res.is_ok());
         ensure_misbehaviour(&ctx_a, &client_id, &tm_client_type());
     }
+
+    #[test]
+    fn test_expired_client() {
+        let remote_chain_id = ChainId::new("mockgaiaB", 1).unwrap();
+        let host_chain_id = ChainId::new("mockgaiaA", 1).unwrap();
```
Farhad-Shabani: To be consistent with other places, could we keep using [our naming convention](https://github.com/cosmos/ibc-rs/blob/dc8e4161dd5e00a2031cf3feb58dad1ed7e26bdb/crates/ibc/src/core/mod.rs#L3-L23) instead of remote and host?

crates/ibc/Cargo.toml
```diff
@@ -73,6 +73,7 @@ scale-info = { version = "2.1.2", default-features = false, features = ["derive"
 ## for borsh encode or decode
 borsh = {version = "0.10", default-features = false, optional = true }
 parking_lot = { version = "0.12.1", default-features = false, optional = true }
+typed-builder = { version = "0.16.2", optional = true }
```
Farhad-Shabani: Should update this to 0.17.0?

crates/ibc/src/mock/context.rs
```diff
@@ -683,15 +880,17 @@ impl MockContext {
...
         let new_block = HostBlock::generate_block(
             self.host_chain_id.clone(),
             self.host_chain_type,
             latest_block.height().increment().revision_height(),
-            latest_block
-                .timestamp()
-                .add(self.block_time)
-                .expect("Never fails"),
+            timestamp,
         );
```
Farhad-Shabani: Here, we logically should have a single method to advance the host chain, managing both the height and timestamp. This method should not accept the timestamp as an argument to align with the `block_time` field of the `MockContext`. When necessary, advancing by a specific duration it either should be done by mulitple calls or setting the `block_time`` other that its default value.

crates/ibc/src/mock/context.rs
```diff
@@ -217,6 +219,131 @@ pub struct MockContext {
     pub logs: Vec<String>,
 }
 
+#[derive(Debug, TypedBuilder)]
```
Farhad-Shabani, line 222: As the entire MockContext is subject to change soon, I didn't delve deeply into abstraction using `TypedBuilder` in this PR. Seeing as it works for our tests, I think it's fine. Though a question I got: couldn’t we directly derive `TypedBuilder` on the own `MockContext`?

Farhad-Shabani (changes requested):

crates/ibc/src/mock/host.rs
rnbguy: Very good question. To answer your question, no. (Light-)Blocks do not store `trusted_height` and `trusted_next_validators`. They are computed by the relayer by querying the remote blockchain. Actually, I will suggest removing these two fields from `SyntheticTmBlock` as these two fields are not required for a host context. These fields are only needed when the relayer submits a client update message to the counterparty chain.

Previously, in our Mock version, `trusted_next_validators` was computed as the next validator of the current header - which is semantically wrong. But works for consecutive client height updates.

And the same was being done in the issue reported by Nomic. This is why my previously written test should not have failed. But I made a mistake :sweat_smile: The fix is in a2163f1cd2b4f848ae285ec26bb9f0ac0719d7b1. With this fix, the bug couldn't have been caught by tests.

Note, I can't set `light_block.header.next_validator` later - as this will break the header signature verification.

crates/ibc/src/mock/context/clients.rs
rnbguy: Ah yes. You are right :+1:

crates/ibc/src/mock/context.rs
rnbguy: Are we ok with maintaining the macro code for deriving Builders? I don't think we should manually write these setter methods.

crates/ibc/src/mock/context.rs
rnbguy: Hmm. Correct me if I am wrong. It can happen in real life that the chain was halted but later restarted again from a halted height. In that particular height, the block time can be arbitrarily long. There is no block production logic to avoid this. I tried to mimic that situation here.

On the other hand, I do agree that if we have a `block_time` parameter, we should follow it. If we must follow this, the other solution is to iterate a lot of blocks without updating the light-client to pass over the `trusting_period`. But in my experience, generating tm light clients is expensive, and I couldn't finish the test quickly with the default `trusting_period` - `64000` secs. Maybe we should reduce the `trusting_period` for this test?

crates/ibc/src/core/ics02_client/handler/update_client.rs
rnbguy: looks like `ctx` is the convention for most of the tests. Should I change each of them to `ctx_a`?

rnbguy: I updated the tests accordingly with `ctx_a`, where `ctx_b` is also used.

crates/ibc/Cargo.toml
```diff
@@ -104,3 +105,4 @@ test-log = { version = "0.2.10", features = ["trace"] }
 tendermint-rpc = { version = "0.34", features = ["http-client", "websocket-client"] }
 tendermint-testgen = { version = "0.34" } # Needed for generating (synthetic) light blocks.
 parking_lot = { version = "0.12.1" }
+typed-builder = { version = "0.17.0" }
```
rnbguy: @Farhad-Shabani should we use this to avoid redefining the dev dependencies?

https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481

With this, we can enable `ibc` crate with `mock` feature and drop the dev-deps for tendermint-testgen, typed-builder and parking-lot.

crates/ibc/src/core/ics02_client/handler/update_client.rs
Farhad-Shabani: For now, I suggest we just keep things right, for this PR. 
Later, we should review all similar misnamings.

crates/ibc/src/mock/context.rs
Farhad-Shabani: > It can happen in real life that the chain was halted but later restarted again from a halted height. In that particular height, the block time can be arbitrarily long. 

This subtle scenario can exist too afaik. Good point of view.
However, given most client expiry would occur during regular operations, I am more inclined towards keeping things consistent the `block_time`, and cutting down the trust period makes more sense to me.

crates/ibc/src/mock/context.rs
Farhad-Shabani: On board with this. Should preferably use existing solutions.
Let's discuss this in more detail later on to see how it would be best to derive Builders on different IBC types to keep things consistent and only for testing.

crates/ibc/src/mock/host.rs
Farhad-Shabani: > the same was being done in the issue reported by Nomic

I felt why this is added, but sounds like putting these under the `SyntheticTmBlock ` is causing confusion. 

> Actually, I will suggest removing these two fields from SyntheticTmBlock as these two fields are not required for a host context. These fields are only needed when the relayer submits a client update message to the counterparty chain.

Correct. I'll leave it to you. Would be good to come up with better struct definition here. But don't spend too much time on this. Again, as the mock abstraction will be changed in the end..

crates/ibc/Cargo.toml
Farhad-Shabani: Looks cool. Though not sure how much stable it is. I'll let you decide.

rnbguy: I think the workaround is stable and part of the dependency resolver version 2 - which we are using already. But the ideal solution should be more direct.

I will add this, as this only affects dev-dependencies.

rnbguy: Just one thing. This will enable `mocks` features by default in the test environment - which is equivalent to the current `cargo test --features mocks` command.

But IMO, this makes sense, as we want to run mock-based tests by default under `cargo test`.

Farhad-Shabani: That's fine then.
Just take this into consideration as well: we will soon split out the mock module into a new crate :)

crates/ibc/src/mock/host.rs
rnbguy: Let's tackle this in future PR or directly by mock abstraction - as this includes the existing `trusted_height` field.

rnbguy:
@Farhad-Shabani I have made the changes. I also added a new test to check `max_clock_drift`. I will create the issue about the TODO comments after this PR gets merged.

Do I need to add a changelog entry?

Farhad-Shabani:
> @Farhad-Shabani I have made the changes. I also added a new test to check `max_clock_drift`. I will create the issue about the TODO comments after this PR gets merged.

Thanks a lot @rnbguy.

> Do I need to add a changelog entry?

Yes please!

Farhad-Shabani (approved):
All set 🙏🏻

Issue #603, opened 2023-04-04
[ICS02] Allow zero `Height` creation, but keep the H>0 condition for CometBFT clients


Farhad-Shabani:
## Problem Statement

- **Client-specific condition**
As for the new `Height` creation ([here](https://github.com/cosmos/ibc-rs/blob/ef52941e29aa2759e2861cdae8a7398b114cc5f2/crates/ibc/src/core/ics02_client/height.rs#L37-L40)) it isn’t allowed to create a zero height and this height >0 is a client-specific condition. Looking into the links provided [here](https://github.com/cosmos/ibc-go/issues/1576#issuecomment-1166974455) It is required for CometBFT-based chains)

- **Not in conflict with ICS02**
This condition is not clearly outlined in ICS07, though should be added there. But ICS02 explicitly requires to have a zero element for the [Height](https://github.com/cosmos/ibc/tree/main/spec/core/ics-002-client-semantics#height) type which **is not in conflict** with having H>0 for the CometBFT clients:
    > There must also be a zero-element for a height type, referred to as 0, which is less than all non-zero heights.

- **Error handling side effect**
Upon converting a proto zero height, we have to deal with the zero-height error variant (`ClientError::InvalidHeight`) which is only needed for ICS07-related codes. And, in the places we have to observe H>0, this error is mapped to a different type and even left `unwrap()` like e.g. [here](https://github.com/cosmos/ibc-rs/blob/22fe103e119f1e8709962a3db2875893e9a64ec4/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L185-L187) & [here](https://github.com/cosmos/ibc-rs/blob/ef52941e29aa2759e2861cdae8a7398b114cc5f2/crates/ibc/src/clients/ics07_tendermint/header.rs#L51-L54) & [here](https://github.com/cosmos/ibc-rs/blob/ef52941e29aa2759e2861cdae8a7398b114cc5f2/crates/ibc/src/core/ics02_client/height.rs#L182) & [here](https://github.com/cosmos/ibc-rs/blob/22fe103e119f1e8709962a3db2875893e9a64ec4/crates/ibc/src/clients/ics07_tendermint/client_state.rs#L408-L410)

- **Benefit from `Option<Height>`**
Living in Rust, using `Option` and converting to `None` would be more efficient, where a proto zero height means the absence of height, while still allowing zero height creation. (As also suggested in https://github.com/cosmos/ibc/issues/776)

## Related Context
[...]

PR #932, opened 2023-10-21
ibc: prevent invalid Height from being deserialised

mina86:
Change revision_height from u64 to NonZeroU64 to make invalid heights
(ones where revision_height is zero) impossible to represent.  This
makes borsh and serde deserialisation fail when given invalid height
(while in the past they would construct an invalid object).

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


mina86:
@Farhad-Shabani

rnbguy:
Thanks for the PR @mina86.

We intend to change this behavior via #603. To summarize, `ics-02` allows zero for heights. It is the `ics-07` (_tmclient_) that doesn't allow zero for heights. Hence this validation should be client-specific.

But if this is blocking you, please open an issue and we will consider this PR.

PR #941, opened 2023-11-01
fix: support chain identifiers without `{chain_name}-{revision_number}` pattern

### LINKED ISSUE #940: `ChainId` parsing fails if revision number is absent
## Bug Summary

`ChainId::from_str("basecoin")` fails as revision number is not given. [_ibc-go_ docs](https://github.com/cosmos/ibc-go/blob/a11c558664df2d4c3a21dbad6118c54ebcd8e9af/docs/docs/01-ibc/01-overview.md?plain=1#L96-L111) discusses that the clients without revision requirement should have hardcoded zero as revision number.

## Details

Current `ChainId::from_str` uses `parse_chain_id_string`, which requires a revision number at the end of the string. 

https://github.com/cosmos/ibc-rs/blob/553bc79b31b31c586128e4e533105fe917df4c0e/crates/ibc/src/core/ics24_host/identifier.rs#L139-L146

But _ibc-go_ implementation [doesn't require](https://github.com/cosmos/ibc-go/blob/a11c558664df2d4c3a21dbad6118c54ebcd8e9af/modules/core/02-client/types/height.go#L172-L176) this and returns zero by default if the revision number is not specified.

## Version

v0.47.0

### LINKED ISSUE #943: `validate_prefix_length` can avoid creating new `String`s
## Bug Summary

`validate_prefix_length` unnecessarily creates a new `String` using `u64::MIN` and `u64::MAX` which can be avoided.

## Details

The current implementation looks as follows.

https://github.com/cosmos/ibc-rs/blob/ccca3ffeda2406af6c43558ebde84c2bc1d62ef9/crates/ibc/src/core/ics24_host/identifier/validate.rs#L59-L79

Instead of doing

```rs
validate_identifier_length( 
         &format!("{prefix}-{}", u64::MIN), 
         min_id_length, 
         max_id_length, 
     )
```

we can directly pass `prefix` with `{min,max}_id_length` with negative offset for `format!("-{}", u64::MIN)`.

```rs
validate_identifier_length( 
         prefix,
         min_id_length - 2,
         max_id_length - 2,
     )
```

And, similar for `u64::MAX` with offset `21`. But we need to be careful that these subtractions don't underflow.

## Version

v0.47.0

rnbguy:
Closes: #940
Closes: #943 

## Description

[ibc-go allows](https://github.com/cosmos/ibc-go/blob/a11c558664df2d4c3a21dbad6118c54ebcd8e9af/docs/docs/01-ibc/01-overview.md?plain=1#L96-L111) a hardcoded revision number for a chain-id without revision number.

Currently, `ibc-rs` requires all chain-ids to have a revision number. This PR allows `ChainId` struct to accept chain ID strings without revision numbers. 

This PR also adds an optimized implementation for `validate_length_prefix`.

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -43,7 +43,7 @@ const TRANSFER_PORT_ID: &str = "transfer";
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct ChainId {
     id: String,
-    revision_number: u64,
+    revision_number: Option<u64>,
```
Farhad-Shabani: From what I see, the primary focus here is on making `ChainId`'s specific format (`{chain_name}-{revision_number}`) optional, rather than the revision number. Thus, even if a `ChainId` doesn't adhere to the current format, it should still have a default revision number (0). Specifically the `Option` here serves the purpose of specifying whether the ChainId complies with the specified format. 
Note that since the `id` field fully represents the `ChainId`, it's possible to detect whether a `ChainId` includes a revision number through the 'id' itself. Therefore, I believe our solution should involve introducing a default revision number of zero when parsing the ChainId string, in cases where the `id` does not conform to the expected format and can keep the data structure as is.

Farhad-Shabani (changes requested):
Thanks @rnbguy for taking care of this 🙏🏻  
Since the ChainId's data structure has changed here and other changes are related, let's go over that first:

crates/ibc/src/core/ics24_host/identifier.rs
rnbguy: I wanted to make a distinction between _zero revision number_ and _no revision number_. I agree that `has_revision_number` can reuse `split_chain_id` - but I wanted to call it once when a chain ID is created/parsed and save the information as `Option<u64>`.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -72,12 +72,22 @@ impl ChainId {
...
 
+    pub fn new_without_revision_number(name: &str) -> Result<Self, IdentifierError> {
+        let prefix = name.trim();
+        validate_identifier_chars(prefix)?;
+        validate_identifier_length(prefix, 1, 43)?;
+        Ok(Self {
+            id: prefix.to_owned(),
+            revision_number: 0,
+        })
+    }
```
Farhad-Shabani: To enhance clarity, I recommend renaming the old `new` method to e.g. `new_with_revision_format` in order to explicitly convey its purpose of constructing with a specific format suitable for tendermint chains. Additionally, we can designate this newly added method as the default, general-purpose constructor, referred as `new` which is implemented as: `new_with_revision_format(name, 0)`

Also, can you plz add a docstring for it?

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -134,36 +164,28 @@ impl Display for ChainId {
...
-    let (name, rev_number_str) = match chain_id_str.rsplit_once('-') {
-        Some((name, rev_number_str)) => (name, rev_number_str),
-        None => {
-            return Err(IdentifierError::InvalidCharacter {
-                id: chain_id_str.to_string(),
-            })
-        }
-    };
+/// returns a tuple containing the chain name and (optional) revision number.
+fn parse_chain_id_string(chain_id_str: &str) -> Result<(&str, Option<u64>), IdentifierError> {
```
Farhad-Shabani: Here, we still can stick with the `(&str, u64)` type, and in case of parsing failure, it defaults to 0. The revision number is always mandatory.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -105,6 +120,21 @@ impl ChainId {
...
+    /// chain_id.unset_revision_number();
+    /// assert!(!chain_id.has_revision_number());
+    /// assert_eq!(chain_id.revision_number(), 0);
+    /// assert_eq!(chain_id.as_str(), "chainA");
+    /// ```
+    pub fn unset_revision_number(&mut self) {
+        let chain_name = self.chain_name();
+        self.id = chain_name.to_string();
+        self.revision_number = 0;
+    }
```
Farhad-Shabani: Since we have maintained the `ChainId` data structure in its original form, there is no longer a need for this method. The `ChainId` whether adheres to the specified format, where the concept of `unset` no longer applies or the `revision_number` consistently remains at zero for all the rest cases. right?

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -92,6 +102,11 @@ impl ChainId {
         self.revision_number
     }
 
+    /// Check if the revision number is set
+    pub fn has_revision_number(&self) -> bool {
+        self.split_chain_id().1.is_some()
+    }
```
Farhad-Shabani: Instead of `has_revision_number` method (which logically should always return true), there should be e.g. `is_revision_format` method that checks if the `id` is of the specific tendermint chainId format.
It also MUST check the extracted revision number from `id` matches with the `revision_number` field, so cases like `ChainId(chainA-2, 0)` should fail.
Otherwise it gets impossible for such cases to find out if the `chainA-2` with `revision_number = 0` is valid or not?!
From what I see, ibc-go checks ChainIds this way, meaning if the ChainId is in the format then 2 is definitely the `revision_number`

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -518,27 +540,81 @@ impl std::error::Error for IdentifierError {}
...
 
     use super::*;
 
-    #[test]
-    fn test_valid_chain_id() {
-        assert!(ChainId::from_str("chainA-0").is_ok());
-        assert!(ChainId::from_str("chainA-1").is_ok());
-        assert!(ChainId::from_str("chainA--1").is_ok());
-        assert!(ChainId::from_str("chainA-1-2").is_ok());
+    #[rstest]
```
Farhad-Shabani, line 549: Would you add a case to ensure that `ChainId::new_with_revision_format("chainA-2", 0)` fails?

.changelog/unreleased/bug-fixes/940-optional-revision-number-in-chain-id.md
```diff
@@ -0,0 +1,2 @@
+- Support `ChainId`s without revision numbers
```
Farhad-Shabani:
```suggestion
- Support `ChainId`s with non-conventional format, beyond the common `{chain_name}-{revision_number}` pattern used in Tendermint chains.
```

Farhad-Shabani (changes requested):

.changelog/unreleased/bug-fixes/940-optional-revision-number-in-chain-id.md
```diff
@@ -0,0 +1,2 @@
+- Support chain identifiers without the `{chain_name}-{revision_number}` pattern
```
Farhad-Shabani: The changelog file name: `940-optional-revision-number-in-chain-id.md` should be updated. The PR title as well.

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -134,36 +153,25 @@ impl Display for ChainId {
 }
 
 /// Parses a string intended to represent a `ChainId` and, if successful,
-/// returns a tuple containing the chain name and revision number.
+/// returns a tuple containing the chain name and (optional) revision number.
```
Farhad-Shabani:
```suggestion
/// returns a tuple containing the chain name and revision number.
```

Farhad-Shabani (commented):
Thanks @rnbguy for digging into this issue! Great job!
Just a few nits, then we are good to go!

crates/ibc/src/core/ics24_host/identifier.rs
```diff
@@ -95,21 +91,25 @@ impl ChainId {
     /// Swaps `ChainId`s revision number with the new specified revision number
```
Farhad-Shabani: This docstring should get updated too.

crates/ibc/src/core/ics24_host/identifier/validate.rs
```diff
@@ -190,42 +209,60 @@ mod tests {
...
     fn parse_invalid_lengthy_client_type() {
         let id = validate_client_type("InvalidClientTypeWithLengthOfClientId>65Char");
         assert!(id.is_err())
     }
+
+    #[rstest::rstest]
+    #[case::empty_prefix("", 1, 64, false)]
+    #[case::max_is_low("a", 1, 10, false)]
+    #[case::u64_max_is_too_big("a", 3, 21, false)]
+    #[case::u64_min_is_loo_small("a", 4, 22, false)]
```
Farhad-Shabani:
```suggestion
    #[case::u64_min_is_too_small("a", 4, 22, false)]
```

Farhad-Shabani (approved):
🎉

PR #942, opened 2023-11-03
feat: optimize `IdentifierError` variants

### LINKED ISSUE #978: `IdentifierError` variants are not mutually exclusive
Remove IdentifierError::Empty and ContainSeparator variants as they
are redundant.  The former is a special case of InvalidLength; the
latter is a special case of InvalidCharacter.

With that, remove length checking from validate_identifier_chars
(which didn’t belong there anyway) and rely on check in
validate_identifier_length instead.

mina86:
Remove IdentifierError::Empty and ContainSeparator variants as they
are redundant.  The former is a special case of InvalidLength; the
latter is a special case of InvalidCharacter.

With that, remove length checking from validate_identifier_chars
(which didn’t belong there anyway) and rely on check in
validate_identifier_length instead.

Closes: https://github.com/cosmos/ibc-rs/issues/978

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


Farhad-Shabani:
Thanks @mina86. Just kindly note that you are touching parts that are getting refactored thru this PR as well: https://github.com/cosmos/ibc-rs/pull/941.

Let me page @rnbguy to look into these points and apply the relevant changes there (particular first one).

Farhad-Shabani:
@mina86, also on a side request:
Really appreciate if you open an issue and put the description there before making any changes and link that to your PR. 
This way, it won't disrupt our project board and helps us prioritize things. We're trying to be efficient with our bandwidth. So, if something's lower on the priority list, we can give you the heads up, so you won't waste time on it.
Thanks!

rnbguy:
hey @mina86, #941 and this PR are in conflict. Please wait till we merge #941.

rnbguy:
The PR has been merged, you may continue with your PR :slightly_smiling_face: but it is now in conflict with `main` branch.

mina86:
> @mina86, also on a side request: Really appreciate if you open an issue and put the description there before making any changes and link that to your PR. This way, it won't disrupt our project board and helps us prioritize things. We're trying to be efficient with our bandwidth. So, if something's lower on the priority list, we can give you the heads up, so you won't waste time on it. Thanks!

From my point of view I’d rather write a PR and then be told that the review is low on priority list than create an issue.  I often don’t exactly know what I want and by the time I formulate an Issue I already have code written for it. If that’s the only concern than I’d prefer not deal with Issues. ;) But let me know if me sending PRs break your workflow.

I guess two features that I’m pretty sure I’d love to have for sure are efficient zero-copy `TryFrom<Vec<u8>>` and `TryFrom<String>>` implementations for the identifiers.

crates/ibc/src/core/ics24_host/identifier/validate.rs
```diff
@@ -74,7 +61,6 @@ pub fn validate_prefix_length(
     if max_id_length < 22 {
```
mina86: By the way, should this be panic similarly to `assert!(max >= min)` above?

rnbguy: This is a valid error. This is equivalent to the following error,
```rs
Error::InvalidLength {
  id: format!("{}-{}", prefix, u64::MAX),
  min: min_id_length,
  max: max_id_length,
}
```

rnbguy: Maybe we need an `InvalidPrefixLength` error variant to distinguish properly.

mina86: So why have that check? There already is check for `format!("{}-{}", prefix, u64::MAX)` being too long.

In fact I disagree that it’s equivalent; it’s a different kind of error. If max_id_length < 22 there exist no valid prefix. This is analogous to situation where max < min. As such, the `assert!(max >= min)` and this check cover the same kind of error (min and max arguments make no sense) and thus should result in the same kind of failure.

Or just get rid of both and let the rest of the code handle those cases.

(Of course this is separate to this PR).

rnbguy: > There already is check for `format!("{}-{}", prefix, u64::MAX)` being too long.

Not anymore. We use `validate_identifier_length(prefix, min_id_length.saturating_sub(21), max_id_length.saturating_sub(21))` directly.

`max_id_length < 22` makes sure two things. (1) the prefix is not empty (2) `max_id_length - 21` (used as `max_id_length.checked_sub(21)`) doesn't underflow.

mina86: > > There already is check for `format!("{}-{}", prefix, u64::MAX)` being too long.
>
> Not anymore. We use `validate_identifier_length(prefix, min_id_length.saturating_sub(21), max_id_length.saturating_sub(21))` directly.

Right, that’s what I meant.  If max_id_length < 22, that check will fail.

> `max_id_length < 22` makes sure two things. (1) the prefix is not empty

There’s explicit check for that at the beginning of the function:

    if prefix.is_empty() {
        return Err(Error::InvalidPrefix {
            prefix: prefix.into(),
        });
    }

> (2) `max_id_length - 21` (used as `max_id_length.checked_sub(21)`) doesn't underflow.

        max_id_length.saturating_sub(21),

handles underflow just fine.

The entire validate_identifier_length could be this:

```rust
pub fn validate_prefix_length(
    prefix: &str,
    min_id_length: u64,
    max_id_length: u64,
) -> Result<(), Error> {
    // Prefix must be at least `min_id_length - 2` characters long since the
    // shortest identifier we can construct is `{prefix}-0` which extends prefix
    // by two characters.
    let min = min_id_length.saturating_sub(2).max(1);
    // Prefix must be at most `max_id_length - 21` characters long since the
    // longest identifier we can construct is `{prefix}-{u64::MAX}` which
    // extends prefix by 21 characters.
    let max = max_id_length.saturating_sub(21);
    if (min..=max).contains(&(prefix.len() as u64)) {
        Ok(())
    } else {
        Err(Error::InvalidLength { /* ... */ })
    }
}
```

* If `max_id_length < min_id_length` than `max(max_id_length - 21, 0)
  < max(min_id_length - 2, 1)` and the `min..=max` range will be empty.
* If prefix is empty, than `prefix.len() < max(min_id_length - 2, 1)`.
* If `max_id_length < 22` than `max_id_length - 22 == 0` and since
  prefix cannot be empty (per previous point) `max_id_length - 22 <
  prefix.len()`.

mina86: Here’s a draft PR with that change: https://github.com/cosmos/ibc-rs/pull/960

rnbguy: This looks excellent :ok_hand:

You optimized one of my old commits https://github.com/cosmos/ibc-rs/commit/69301f8f8c30fd3d26cfb2ce1b042ce30a0cc652 that I reverted.

crates/ibc/src/core/ics24_host/identifier/validate.rs
```diff
@@ -1,19 +1,12 @@
 use super::IdentifierError as Error;
 use crate::prelude::*;
 
-/// Path separator (ie. forward slash '/')
-const PATH_SEPARATOR: char = '/';
```
rnbguy: Don't remove this. I noticed that the source code uses `'/'` in many places. We should use this instead.

.changelog/unreleased/breaking-changes/942-identifier-error.md
```diff
@@ -0,0 +1,6 @@
+- Remove IdentifierError::Empty and ContainSeparator variants as they
+  are special cases of other existing errors.
+  ([\#942](https://github.com/cosmos/ibc-rs/pull/942))
+- Remove IdentifierError::InvalidLength::length field since it can be
+  deduced from the id.
+  ([\#942](https://github.com/cosmos/ibc-rs/pull/942))
```
rnbguy: Please open an issue for this work and link to this PR :slightly_smiling_face:

rnbguy (changes requested):
Thanks :+1:

crates/ibc/src/core/ics24_host/identifier/validate.rs
mina86: It should be added (in another commit) to and used in ibc/src/core/ics24_host/path.rs then.  If I leave it here, it’s an unused identifier:

    error: constant `PATH_SEPARATOR` is never used

rnbguy: I see. It's okay for now. I will create a separate issue for this.

.changelog/unreleased/breaking-changes/978-identifier-error.md
```diff
@@ -0,0 +1,6 @@
+- Remove IdentifierError::Empty and ContainSeparator variants as they
+  are special cases of other existing errors.
+  ([\#978](https://github.com/cosmos/ibc-rs/issues/978))
+- Remove IdentifierError::InvalidLength::length field since it can be
+  deduced from the id.
+  ([\#978](https://github.com/cosmos/ibc-rs/issues/978))
```
rnbguy:
```suggestion
- Optimize `IdentifierError` variants and make them mutually exclusive.
  ([\#978](https://github.com/cosmos/ibc-rs/issues/978))
```

rnbguy (approved):
Thanks for the optimization ! :tada:

PR #968, opened 2023-11-16
feat: turn ClientConsensusStatePath::haight into Height

mina86:
Replace ClientConsensusStatePath’s epoch and height fields with
a single height field whose type is Height.  This stores the exact
same information but allows infallible way to get Height out of the
path.

Arguably it also makes more sense conceptually since the epoch and
height are really just a single Height field.  Even in formatted paths
those two are a single component so it makes sense to have them as
a single field.

Issue: https://github.com/cosmos/ibc-rs/issues/603

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/ibc-testkit/src/testapp/ibc/core/client_ctx.rs
+++ b/crates/ibc-testkit/src/testapp/ibc/core/client_ctx.rs
@@ -235,11 +235,9 @@ impl ClientExecutionContext for MockContext {
                 client_state: Default::default(),
             });
 
-        let height = Height::new(consensus_state_path.epoch, consensus_state_path.height)
-            .expect("Never fails");
         client_record
             .consensus_states
-            .insert(height, consensus_state);
+            .insert(consensus_state_path.height, consensus_state);
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


Farhad-Shabani:
The codes touched in this PR, have been tweaked to align with our `ibc` crate restructuring (in this [PR](https://github.com/cosmos/ibc-rs/pull/969)). Specifically, we've adjusted the `ClientConsensusStatePath` constructor to get the revision numbers and height separately for better decoupling of our ICS-02 and ICS-24 implementations there.
Let's hold off on this change until we wrap up the `ibc` restructuring PRs. I'll ping you after that to revisit and see if we can accommodate any of these changes.

Farhad-Shabani:
Given the new repo structure and to keep max compliance with IBC specs, the `Height` struct is now placed under the `ibc-core-client-types` crate and all the paths under the `ibc-core-host-types`, which two can't be dependent in the new setup (as it causes cyclic dependency issue).
Thus, the changes here can't be accommodated. Though, regarding the issue you mentioned [here](https://github.com/cosmos/ibc-rs/issues/603#issuecomment-1815023716), I believe it will be solved as soon as we allow zero height creation.

PR #972, opened 2023-11-19
feat(core-client)!: replace store_update_{time,height} by metadata update

### LINKED ISSUE #973: feat: replace store_update_{time,height} by metadata update
store_update_time and store_update_height methods are always called
together. In fact, they are also usually called together with
store_consensus_state.

Sadly, since they are separate calls, the implementations need to
treat them as separate pieces of data. For example, if an
implementation stores update times and heights in a map, it must
perform two separate lookups to update each value.

Change the interface such that store_consensus_state in addition takes
host timestamp and height. This makes it possible for implementations
to perform optimisations where timestamp and height are stored in
a single location. This reduces number of lookups and possibly also
improves memory footprint of the implementation.

In fact, they may be stared together with the consensus state making
the optimisation potential even stronger.

Alas, updating of the timestamp and height is at times called without
updating the consensus state. To accommodate that use case, introduce
store_consensus_metadata call which only takes the timestamp and
height arguments.

mina86:
Observe that update time and height are *always* read and manipulated
together.  Replace pairs of methods for reading, updating and deleting
time and height with a single combined method for each operation.

Specifically, replace:
a) client_update_{time,height} by client_update_meta,
b) store_update_{time,height} by store_update_meta and
c) delete_update_{time,height} by delete_update_meta.

This allows better optimised implementations which, for example, keep
both pieces of information in a single map and perform a single lookup
to read or update them.

Closes: https://github.com/cosmos/ibc-rs/issues/973

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


Farhad-Shabani:
Thank you @mina86  for pointing those details out. To progress the PR, here are some points to consider:

- Regarding the `store/delete-consensus-state` methods, note that the [client execution context](https://github.com/cosmos/ibc-rs/blob/d82246db39f620486de4a9ca5846f98ae0f57310/crates/ibc/src/core/ics02_client/context.rs#L37) (b/c of generic `V` type [here](https://github.com/cosmos/ibc-rs/blob/d82246db39f620486de4a9ca5846f98ae0f57310/crates/ibc/src/core/ics02_client/context.rs#L38)) can tap into other methods in our [main validation context](https://github.com/cosmos/ibc-rs/blob/527bb141383129db804b0b9d49793c4a900deede/crates/ibc/src/core/context.rs#L95) (And, actually Tendermint clients already have access to [host_height and host_timestamp](https://github.com/cosmos/ibc-rs/blob/d062728b9321c3f70aec514b38cf195b2f07eddd/crates/ibc/src/clients/ics07_tendermint/context.rs#L12-L21)). So, the design allows you to intoruce extra methods by defining additional customized validation context, like we did [here](https://github.com/cosmos/ibc-rs/blob/d82246db39f620486de4a9ca5846f98ae0f57310/crates/ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs#L134-L152) for our mock implementation. So then, you can go ahead, and call `self.host_timestamp()?` and `self.host_height()?` in your `store/delete-consensus-state` implementation as you require.
Therefore, no need to enforce all the ibc-rs downstream projects to pass host height and timestamp. Additionally, *_consensus_state, typically points out to `ConsensusState` structs in our codebase, and passing height and timestamp along with that can cause confusion among users imho. Let's not touch `store/delete-consensus-state` methods.

- Though, as for merging `store/delete_update_time` & `store/delete_update_client`, since it’s turning to an interface with less assumption about host chains, and vibing well with ibc-go APIs, makes sense.

Farhad-Shabani:
> Though, as for merging store/delete_update_time & store/delete_update_client, since it’s turning to an interface with less assumption about host chains, and vibing well with ibc-go APIs, makes sense

@mina86 If you feel up to finishing this PR as well, we can include the second point (above) in our coming release as well.

mina86:
The change to store/delete-consensus-state was based on the observation that storing or deleting consensus state is *always* followed by update to the time and height. You say

> So then, you can go ahead, and call self.host_timestamp()? and self.host_height()? in your store/delete-consensus-state implementation as you require.

but that means there’s no need for store_update_time and store_update_height at all?

Farhad-Shabani:
> but that means there’s no need for store_update_time and store_update_height at all?

Great question! The rationale behind having these methods is rooted in the need to avoid making assumptions about how light client developers handle client metadata with each `ConsensusState` update. A client could be designed to manage **multiple update clients in a single run**. In such a scenario, it can process and store multiple `ConsensusState` for each update client message but might only store one pair of update time and height at the end. Therefore, the approach allows for flexibility when handling various update scenarios without imposing strict requirements on metadata storage. Additionally, looking at it from the ibc-rs perspective, this gives us distinct access to those parts of the host's store, so can benefit from that in later applications or clients' development.

Btw, this got me thinking — we might be able to skip passing the `host_timestamp` and `host_height` to the `store_update_meta`. Tendermint clients already have internal access, and other clients potentially could follow a similar trait definition. 

Since this is an API-breaking change, let's give it more thought to ensure a robust API, so not imposing gradual adjustments on users.

mina86:
So if I understand correctly, this also means that implementations can
effectively ignore store_update_{time,height} calls and instead perform
the updates when handling store_consensus_state?

> Btw, this got me thinking — we might be able to skip passing the `host_timestamp` and `host_height` to the `store_update_meta`. Tendermint clients already have internal access, and other clients potentially could follow a similar trait definition. 

Makes sense to me.  I’ve whipped out an alternative PR at
<https://github.com/cosmos/ibc-rs/pull/981>.

I think in the end it also means that implementations could have empty
store_update_meta implementations provided that they updated the time
and height in store_consensus_state.

-- 
Best regards
ミハウ “𝓶𝓲𝓷𝓪86” ナザレヴイツ
«If at first you don’t succeed, give up skydiving»

Farhad-Shabani:
Hey @mina86. Sorry it took us a bit to circle back to this PR. Now that we've got #1069 merged, feel free to update the type of arguments with the newly defined paths, so can go ahead with this PR. Thanks!

mina86:
With this PR we actually want to replace the two new paths ClientUpdateTimePath and     ClientUpdateHeight with ClientUpdateMetadataPath, no?

Farhad-Shabani:
> With this PR we actually want to replace the two new paths ClientUpdateTimePath and ClientUpdateHeight with ClientUpdateMetadataPath, no?

You've got a good point. I thought in the wrong direction. Let's stick with the current paths and keep the method signatures as is. So that, inside the implementation of these methods, whoever's working on it can decide whether or not to use the path types.

ibc-clients/ics07-tendermint/types/src/error.rs
```diff
@@ -74,10 +74,8 @@ pub enum Error {
     NotEnoughTrustedValsSigned { reason: VotingPowerTally },
     /// verification failed: `{detail}`
     VerificationError { detail: LightClientErrorDetail },
-    /// Processed time for the client `{client_id}` at height `{height}` not found
+    /// Processed time or height for the client `{client_id}` at height `{height}` not found
     ProcessedTimeNotFound { client_id: ClientId, height: Height },
```
Farhad-Shabani: Rename this variant to `UpdateMetaDataNotFound` ?

.changelog/unreleased/breaking-changes/973-update-meta.md
```diff
@@ -0,0 +1,6 @@
+- Merge client update time and height modification method pairs into
```
Farhad-Shabani, line 1:
```suggestion
- [ibc-core-client] Merge client update time and height modification method pairs into
```

mina86:
In that case I think this PR is good as is.

Though I struggle to imagine why any implementation would choose use those paths. There’s no benefit in having separate paths for the the height and timestamp since those two pieces of information are always accessed together.

ibc-core/ics02-client/context/src/context.rs
```diff
@@ -12,19 +12,13 @@ use super::consensus_state::ConsensusState;
 /// [crate::client_state::ClientStateValidation] must
 /// inherit from this trait.
 pub trait ClientValidationContext {
-    /// Returns the time when the client state for the given [`ClientId`] was updated with a header for the given [`Height`]
-    fn client_update_time(
+    /// Returns the time and height when the client state for the given
+    /// [`ClientId`] was updated with a header for the given [`Height`]
+    fn client_update_meta(
```
Farhad-Shabani: to keep naming consistent with the other methods, how about we rename this to "update_meta"?

ibc-core/ics02-client/types/src/error.rs
```diff
@@ -26,10 +26,8 @@ pub enum ClientError {
     ClientStateAlreadyExists { client_id: ClientId },
     /// consensus state not found at: `{client_id}` at height `{height}`
     ConsensusStateNotFound { client_id: ClientId, height: Height },
-    /// Processed time for the client `{client_id}` at height `{height}` not found
+    /// Processed time or height for the client `{client_id}` at height `{height}` not found
     ProcessedTimeNotFound { client_id: ClientId, height: Height },
```
Farhad-Shabani: Rename this variant to `UpdateMetaDataNotFound` ?

Farhad-Shabani (commented):
Left a few comments, then all set :pray:

Farhad-Shabani:
> Though I struggle to imagine why any implementation would choose use those paths. There’s no benefit in having separate paths for the the height and timestamp since those two pieces of information are always accessed together.

Good call. I'll check out what's happening with projects around ibc-rs to get a clearer picture so can before the next release, decide if we should keep the current signatures or define e.g. "ClientUpdateMetadataPath".

.changelog/unreleased/breaking-changes/973-update-meta.md
```diff
@@ -0,0 +1,6 @@
...
+  one, that is replace
+  a) client_update_{time,height} by update_meta,
+  b) store_update_{time,height} by store_update_meta and
+  c) delete_update_{time,height} by delete_update_meta.
+  ([#973](https://github.com/cosmos/ibc-rs/issues))
```
Farhad-Shabani:
```suggestion
  ([\#972](https://github.com/cosmos/ibc-rs/pull/972))
```

Farhad-Shabani (approved):

PR #1004, opened 2023-12-12
fix: packet sequence in Timeout handlers should be encoded in big-endian

en:
## Description

refer to: 
[// Note: We expect the return to be a u64 encoded in big-endian. Refer to ibc-go:
        // https://github.com/cosmos/ibc-go/blob/25767f6bdb5bab2c2a116b41d92d753c93e18121/modules/core/04-channel/client/utils/utils.go#L191](https://github.com/informalsystems/hermes/blob/c945791f0fca43acf58648a78635ae21e774e483/crates/relayer/src/chain/cosmos.rs#L1989-L1990)

The `seq_on_a` of MsgTimeout packet should be encoded as a big-endian bytes.
Here's a code snippet to show this problem.
https://gist.github.com/en/e4f6befdef94b1e906ed0054709ccd4b

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/bug-fixes/1004-fix-packet-sequence-encoding.md
+++ b/.changelog/unreleased/bug-fixes/1004-fix-packet-sequence-encoding.md
@@ -1,0 +1,2 @@
+- Encode packet sequence into a big endian bytes.
+  ([\#1004](https://github.com/cosmos/ibc-rs/pull/1004))
--- a/ibc-core/ics04-channel/src/handler/timeout.rs
+++ b/ibc-core/ics04-channel/src/handler/timeout.rs
@@ -229,14 +229,13 @@ where
             }
             let seq_recv_path_on_b =
                 SeqRecvPath::new(&msg.packet.port_id_on_b, &msg.packet.chan_id_on_b);
-            let value = u64::from(msg.packet.seq_on_a).to_be_bytes().to_vec();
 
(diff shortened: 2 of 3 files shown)


Farhad-Shabani (changes requested):
Nice catch @en! Thanks for this complete report. 👌🏻 

Let’s apply the following changes:

1. This should be fixed [here](https://github.com/cosmos/ibc-rs/blob/2259108eda5fc5500e9234876f9482280eac44dc/ibc-core/ics04-channel/src/handler/timeout_on_close.rs#L142) as well for the `MsgTimeoutOnClose`.

2. Let’s revise this [`to_vec`](https://github.com/cosmos/ibc-rs/blob/2259108eda5fc5500e9234876f9482280eac44dc/ibc-core/ics24-host/types/src/identifiers/sequence.rs#L53-L57) method with `self.0.to_be_bytes.to_vec()` implementation and update it’s docstring, so we can afterwards call it on the `msg.packet.seq_on_a`

3. Dress up the PR with a [unclog](https://github.com/cosmos/ibc-rs/blob/main/CONTRIBUTING.md#changelog).

Then we are all set!

Farhad-Shabani (approved):
🙏🏻

PR #1007, opened 2023-12-13
Prioritize redundant packet checks

### LINKED ISSUE #817: Prioritize redundant packet checks for better performance in multi-relayer scenarios
## Background

Originally derived from a conversation with the author of [“Analyzing the Performance of IBC”](https://arxiv.org/abs/2303.10844)

## Concern
Having a single instance of relayer works better compared to multiple relayers, as the on-chain processing cost of redundant packets lowers the overall IBC performance.  

## Proposal

A resolution is to ensure checks for “already relayed packet” in places like [here](https://github.com/cosmos/ibc-rs/blob/7191d8839eb2ff8c9e95a455a5727e233ea2b28a/crates/ibc/src/core/ics04_channel/handler/recv_packet.rs#L49-L75), and [here](https://github.com/cosmos/ibc-rs/blob/7191d8839eb2ff8c9e95a455a5727e233ea2b28a/crates/ibc/src/core/ics04_channel/handler/acknowledgement.rs#L141-L150) get done before any other validation/execution processes as much as possible. This should lower the number of soft and, more importantly, hard checks against the store in redundant packet cases.

expertdicer:
Closes: #817

## Description

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-core/ics04-channel/src/handler/acknowledgement.rs
+++ b/ibc-core/ics04-channel/src/handler/acknowledgement.rs
@@ -113,9 +113,24 @@ fn validate<Ctx>(ctx_a: &Ctx, msg: &MsgAcknowledgement) -> Result<(), ContextErr
 where
     Ctx: ValidationContext,
 {
+    let packet = &msg.packet;
+
+    let commitment_path_on_a =
+        CommitmentPath::new(&packet.port_id_on_a, &packet.chan_id_on_a, packet.seq_on_a);
+
+    // Verify packet commitment
+    let commitment_on_a = match ctx_a.get_packet_commitment(&commitment_path_on_a) {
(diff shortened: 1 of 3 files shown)


Farhad-Shabani:
Thank you @expertdicer for bringing up this issue.
I took another look at our checks and current logic and had a lightbulb moment. So, these new moves could boost the parallel relaying setup's performance, but there's a catch. We're kind of assuming that all relayers are on the up-and-up, just forwarding legit packets. But the thing is these changes (as we are placing basic checks afterward) leaves our implementation wide open to potential DOS attacks leaving us w/o having clues of why we are hitting these early no-op returns.
Apologies for that. I'll close the issue as well.

PR #1010, opened 2023-12-17
fix: recursive call in connection `State` conversion to `i32`

### LINKED ISSUE #1009: Connection State conversion to i32 recursively calls itself causing stack overflow
## Bug Summary

The `from` implemented for `state` enum for conversion from `state` to `i32` in ics03-connection recursively calls itself causing an infinite loop and eventually overflowing the stack. 
For ref: https://github.com/cosmos/ibc-rs/blob/3894b50929e207eea3653e8bea3c70048d783d8c/ibc-core/ics03-connection/types/src/connection.rs#L560-L564

## Details

To reproduce the issue, try the following.
```rs
let state = ibc_core_connection_types::State::INIT;
let state_in_i32 = i32::from(state); // Would cause stack overflow here
```

## Version

0.48.1

dhruvja:
Closes: #1009 

## Description

Since the conversion from `state` to `i32` causes a recursive call, it can be solved by matching the enum variants and returning the i32 for the respective variant.

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


ibc-core/ics03-connection/types/src/connection.rs
```diff
@@ -559,6 +559,11 @@ impl TryFrom<i32> for State {
 
 impl From<State> for i32 {
     fn from(value: State) -> Self {
-        value.into()
+        match value {
+            State::Uninitialized => 0,
+            State::Init => 1,
+            State::TryOpen => 2,
+            State::Open => 3,
+        }
```
mina86:
```suggestion
        value as i32
```

mina86 (commented):
Also, while you’re doing this could you replace `TryFrom<i32>` implementation to call `Self::from_i32` (or the other way around)?

Farhad-Shabani (approved):
Nice catch @dhruvja, thank you for the fix.

dhruvja:
> Nice catch @dhruvja, thank you for the fix.

Spent quite a while figuring out the vague `stack overflow` bug lol.

PR #1032, opened 2024-01-08
fix: calculate trace hash from both class ID and token ID

yito88:
## Description
Fix for `feat/ics721-impl`

The class ID and the token ID specify the NFT.
The trace hash should be calculated from both IDs, and we need to submit a trace event per each token.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-apps/ics721-nft-transfer/src/context.rs
+++ b/ibc-apps/ics721-nft-transfer/src/context.rs
@@ -105,9 +105,13 @@ pub trait NftTransferValidationContext {
         memo: &Memo,
     ) -> Result<(), NftTransferError>;
 
-    /// Returns a hash of the prefixed class ID.
-    /// Implement only if the host chain supports hashed class ID.
-    fn class_hash_string(&self, _class_id: &PrefixedClassId) -> Option<String> {
+    /// Returns a hash of the prefixed class ID and the token ID.
+    /// Implement only if the host chain supports hashed class ID and token ID.
+    fn token_hash_string(
+        &self,
+        _class_id: &PrefixedClassId,
+        _token_id: &TokenId,
+    ) -> Option<String> {
         None
     }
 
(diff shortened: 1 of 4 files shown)


Farhad-Shabani (approved):
Does this PR mean that the [same event](https://github.com/bianjieai/nft-transfer/blob/1d33cd905efd41fe75ed9a400a130bccf57b5fe8/keeper/relay.go#L299-L304) within the [bianjieai/nft-transfer](https://github.com/bianjieai/nft-transfer) repo should get fixed as well?

Farhad-Shabani:
Just a nit. To pass the CI we need to place `context`, `handler`, and `module` [modules](https://github.com/cosmos/ibc-rs/blob/e5e8464972c04c21cba9ed8686c6d723c85aa71a/ibc-apps/ics721-nft-transfer/src/lib.rs#L20-L22) under the `#[cfg(feature = "serde")]` flag

yito88:
> Does this PR mean that the [same event](https://github.com/bianjieai/nft-transfer/blob/1d33cd905efd41fe75ed9a400a130bccf57b5fe8/keeper/relay.go#L299-L304) within the [bianjieai/nft-transfer](https://github.com/bianjieai/nft-transfer) repo should get fixed as well?

I think so. In `bianjieai/nft-transfer`, the prefixed class is handled like the prefixed denom in ICS-20. However, specifying the actual token is required with not only the class but also the token ID. In Cosmos, a non-NFT token received over IBC is shown as `ibc/{hash}`. I believe it's natural to express both non-NFT and NFT as `ibc/{hash}`.

Farhad-Shabani:
> > Does this PR mean that the [same event](https://github.com/bianjieai/nft-transfer/blob/1d33cd905efd41fe75ed9a400a130bccf57b5fe8/keeper/relay.go#L299-L304) within the [bianjieai/nft-transfer](https://github.com/bianjieai/nft-transfer) repo should get fixed as well?
> 
> I think so. In `bianjieai/nft-transfer`, the prefixed class is handled like the prefixed denom in ICS-20. However, specifying the actual token is required with not only the class but also the token ID. In Cosmos, a non-NFT token received over IBC is shown as `ibc/{hash}`. I believe it's natural to express both non-NFT and NFT as `ibc/{hash}`.

Makes sense. Thank you!

PR #1046, opened 2024-01-18
imp(ibc-testkit): remove field access of `MockContext`

### LINKED ISSUE #1043: imp(ibc-testkit): remove field access of `MockContext`
## Bug Summary

Some of the tests are written while accessing `MockContext` fields. Field access is discouraged over associated methods.

## Details

The following should be refactored to use an associated method, e.g., `MockContext::ibc_events(&self)`.

https://github.com/cosmos/ibc-rs/blob/d5e83145c177a2cd62135aebadaba5c9feb0fa2b/ibc-testkit/tests/core/ics04_channel/timeout.rs#L380-L385

## Version

till v0.49.1

rnbguy:
Closes: #1043

## Description

Removes `ctx.events` or `context.events` access from the tests.

This PR also moves `.events` and `.logs` fields under `MockIbcStore` to keep them under `Arc<Mutex<MockIbcStore>>`.

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/improvements/1043-remove-field-access-of-mockcontext.md
+++ b/.changelog/unreleased/improvements/1043-remove-field-access-of-mockcontext.md
@@ -1,0 +1,2 @@
+- [ibc-testkit] Remove field access of `MockContext`.
+  ([\#1043](https://github.com/cosmos/ibc-rs/issues/1043))
--- a/ibc-testkit/src/fixtures/core/context.rs
+++ b/ibc-testkit/src/fixtures/core/context.rs
@@ -120,8 +120,6 @@ impl From<MockContextConfig> for MockContext {
             history,
             block_time: params.block_time,
             ibc_store: Arc::new(Mutex::new(MockIbcStore::default())),
-            events: Vec::new(),
-            logs: Vec::new(),
         }
     }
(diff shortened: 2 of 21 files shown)


Farhad-Shabani (approved):
Thanks @rnbguy. Just a quick note about this change:

My understanding so far, an IBC module usually should not autonomously store any events and logs. I know it is so in the `basecoin-rs`, but rationally, the storage access in IBC is designed to be either provable or private, serving the purpose of maintaining a record of IBC states as outlined in the specs.

Hosts typically maintain a sort of pool that collects dispatched events and logs from various modules and different parts of the system. Then the aggregated arrays would be included in the response of a transaction. Though, to achieve this, we should decouple storage from keys, so fields in `IbcMockStore` would act as mappers instead of having instant access to the store. Nevertheless, for now, we are fine with this change.

rnbguy:
Totally see your point. But my view is that, since we are working in the context of the IBC _module_, I see them as effects of the IBC module.  So, to simplify things, I wanted to put everything that IBC affects - stores, logs, events; in one unified struct.

PR #1054, opened 2024-01-23
imp: re-export `Timestamp` & `Duration` from `ibc_proto::google::protobuf`

Farhad-Shabani:
## Description

This requirement emerged while implementing the `sov-celestia` light client, eliminating the need to import the same types from `tendermint_proto::google::protobuf`.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### DIFF
--- a/ibc-primitives/src/lib.rs
+++ b/ibc-primitives/src/lib.rs
@@ -25,8 +25,9 @@ pub use traits::*;
 mod types;
 pub use types::*;
 
-/// Re-export of common proto types from the `ibc-proto` crate.
+/// Re-exports google proto types and `Protobuf` trait from the `ibc-proto-rs`
+/// crate.
 pub mod proto {
-    pub use ibc_proto::google::protobuf::Any;
+    pub use ibc_proto::google::protobuf::*;
     pub use ibc_proto::Protobuf;
 }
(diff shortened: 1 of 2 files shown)


ibc-primitives/src/lib.rs
```diff
@@ -25,8 +25,9 @@ pub use traits::*;
 mod types;
 pub use types::*;
 
-/// Re-export of common proto types from the `ibc-proto` crate.
+/// Re-exports google proto types and `Protobuf` trait from the `ibc-proto-rs`
+/// crate.
 pub mod proto {
-    pub use ibc_proto::google::protobuf::Any;
+    pub use ibc_proto::google::protobuf::*;
```
rnbguy: the scope of `ibc_proto::google::protobuf::*` is too broad. The structs other than `Any`, `Timestamp`, and `Duration` are for parsing protobuf files. Do we really need them? 

```suggestion
    pub use ibc_proto::google::protobuf::{Any, Duration, Timestamp};
```

rnbguy (changes requested):
Asked a question to confirm if we need to re-export everything.

rnbguy (approved):
Great ! Thanks :pray:

ibc-primitives/src/lib.rs
Farhad-Shabani: I was under the impression that a few types are exposed by `ibc_proto::google::protobuf` not all. Thank you for pointing that out 👌🏻

Issue #1061, opened 2024-01-25
Ease frozen Height check in tendermint ClientState protobuf deserialization


Farhad-Shabani:
## Summary

The current implementation, with the frozen height check in [`TryFrom`](https://github.com/cosmos/ibc-rs/blob/main/ibc-clients/ics07-tendermint/types/src/client_state.rs#L244) causes issues for some hosts (see #1040), as it hampers the successful retrieval and deserialization of the `ClientState` during the standard process of converting raw bytes to proto type and then to domain type from the storage.
As within all `ibc-rs` handlers, we check the status of a client by performing the `verify_is_active()` on the retrieved `ClientState`, it’s safe to drop the frozen height check in this proto conversion.  

Props to @riversyang for bringing this up in the first place, and thanks to @rnbguy for spotting it while diving into the `ibc-testkit` refactoring!

## Version
=< 0.50.0

PR #1064, opened 2024-01-29
fix: revert `frozen_height` to 0 for active clients in conversions

Farhad-Shabani:
## Description

In #1061, when converting an active Tendermint ClientState to its corresponding proto type, we set the `frozen_height` as `None`  to ensure consistency when switching between domain and proto types. However, [basecoin-rs revealed](https://github.com/informalsystems/basecoin-rs/actions/runs/7661621925/job/20881356539#step:14:941) that the ibc-go driven chain can only properly decode the Protobuf encoded of an active `ClientState` value when the `frozen_height` is explicitly set to zero!

Integration test: https://github.com/informalsystems/basecoin-rs/pull/151

______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-clients/ics07-tendermint/types/src/client_state.rs
+++ b/ibc-clients/ics07-tendermint/types/src/client_state.rs
@@ -284,7 +284,11 @@ impl TryFrom<RawTmClientState> for ClientState {
             .try_into()
             .map_err(|_| Error::MissingLatestHeight)?;
 
-        let frozen_height = raw.frozen_height.and_then(|h| Height::try_from(h).ok());
+        // NOTE: In `RawClientState`, a `frozen_height` of `0` means "not
+        // frozen". See:
+        // https://github.com/cosmos/ibc-go/blob/8422d0c4c35ef970539466c5bdec1cd27369bab3/modules/light-clients/07-tendermint/types/client_state.go#L74
+        let frozen_height =
+            Height::try_from(raw.frozen_height.ok_or(Error::MissingFrozenHeight)?).ok();
 
         // We use set this deprecated field just so that we can properly convert
         // it back in its raw form
--- a/ibc-clients/ics07-tendermint/types/src/client_state.rs
+++ b/ibc-clients/ics07-tendermint/types/src/client_state.rs
@@ -6,6 +6,7 @@ use core::str::FromStr;
 use core::time::Duration;
 
 use ibc_core_client_types::error::ClientError;
+use ibc_core_client_types::proto::v1::Height as RawHeight;
 use ibc_core_client_types::Height;
 use ibc_core_commitment_types::specs::ProofSpecs;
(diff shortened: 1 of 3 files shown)


rnbguy:
Am I correct to conclude that `frozen_height` doesn't need to be `Option<Height>`? We can just use `Height` - as it is never set to`None`.

rnbguy:
I got my answer!

1. `RawClientState` never has `None` for `frozen_client` (ibc-go convention). The active case is denoted by `RawHeight { number: 0, height: 0 }`.
2. It's best to use `Option<Height>` and `None` will denote `RawHeight { number: 0, height: 0 }` which is invalid in our domain type `Height`.

ibc-clients/ics07-tendermint/types/src/client_state.rs
```diff
@@ -283,9 +284,6 @@ impl TryFrom<RawTmClientState> for ClientState {
             .try_into()
             .map_err(|_| Error::MissingLatestHeight)?;
 
-        // In `RawClientState`, a `frozen_height` of `0` means "not frozen".
-        // See:
-        // https://github.com/cosmos/ibc-go/blob/8422d0c4c35ef970539466c5bdec1cd27369bab3/modules/light-clients/07-tendermint/types/client_state.go#L74
         let frozen_height = raw.frozen_height.and_then(|h| Height::try_from(h).ok());
```
rnbguy: `None` and `RawHeight { number: 0, height: 0 }` - these two cases are treated identically here. But you commented later:
```
The protobuf encoded `frozen_height` of an active client must be set to `0`
```

In that case, this line should be as follows,

```rs
let frozen_height = Height::try_from(raw.frozen_height.ok_or(Error::EmptyFrozenHeight)?).ok();
```

Also, the removed comment above this is still relevant for the failure case of `Height::try_from`.

Farhad-Shabani: Per our discussion: [6d9416d](https://github.com/cosmos/ibc-rs/pull/1064/commits/6d9416dc2f4b13e40eaceae25dd032ab92a7d2bc)

ibc-clients/ics07-tendermint/types/src/error.rs
```diff
@@ -61,8 +61,8 @@ pub enum Error {
     HeaderTimestampTooLow { actual: String, min: String },
     /// header revision height = `{height}` is invalid
     InvalidHeaderHeight { height: u64 },
-    /// Disallowed to create a new client with a frozen height
-    FrozenHeightNotAllowed,
+    /// frozen height not found
```
rnbguy:
```suggestion
    /// frozen height is missing
```

rnbguy (approved):
One suggestion - otherwise, LGTM :sparkles:

PR #1081, opened 2024-02-08
fix: use header height for Tendermint consensus state storage

### LINKED ISSUE #1080: [ICS07] Incorrect `ConsensusState` storage for headers with heights lower than `ClientState` height
## Bug Summary

Reported by @vmarkushin 

When a header with a height lower than the current client height is submitted for the first time, we store the consensus state [at the height of the current client state](https://github.com/cosmos/ibc-rs/blob/0a275ff15b4392f33cf0e16ad880f67ddbcdeca4/ibc-clients/ics07-tendermint/src/client_state.rs#L386-L387) instead of the provided header height. This action overwrites the existing consensus state for the current height. Since the header is from the past, a `ConsensusState` with an older root is installed at that path. This disrupts commitment proof verification, as the provided proof will be for a different state root than the one we actually stored.

## Details

- [H] Installed headers
- (H) Unseen headers

---- [5] ---- [7] -- (8) ---- [10] -------------- (20)

Scenario: Submit a header with height 8 and the trusted height of 7

## Version

=< v0.50.0

Farhad-Shabani:
Closes: #1080

### Along with this PR

- Removed the check and test for stale mock headers as it was not sufficiently general and did not provide meaningful validation.
- Nit: placing unclog of issue 835 under the `breaking-changes`

______

### PR author checklist:
- [x] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


ibc-testkit/tests/core/ics02_client/update_client.rs
```diff
@@ -109,6 +109,81 @@ fn test_update_client_ok(fixture: Fixture) {
...
+
+    let mut ctx = MockContext::default().with_client_config(
+        MockClientConfig::builder()
+            .client_chain_id(chain_id_b.clone())
+            .client_type(tm_client_type())
+            .client_id(client_id.clone())
+            .latest_height(latest_height)
+            .build(),
+    );
+    let mut router = MockRouter::new_with_transfer();
```
seanchen1991, line 131: Small nit, but you can construct the `router` from the `fixture` function.

seanchen1991 (approved):
One small nit, otherwise looks good to me

ibc-testkit/tests/core/ics02_client/update_client.rs
Farhad-Shabani: The thing is: I need a `MockContext` with the Tendermint `ClientState` for this test, but the current fixture isn't configurable. Getting the router from fixture means constructing an unnecessary `MockContext`.

As an aside, this fixture should be overhauled in favor of a better approach like the one we have for connection tests.

ancazamfir (approved):
lgtm!

ibc-testkit/tests/core/ics02_client/update_client.rs
seanchen1991: > As an aside, this fixture should be overhauled in favor of a better approach like the one we have for connection tests.

I'm trying to address this by defining a `mock_context` function like so:
```
fn mock_context(
    client_id: ClientId,
    height: Height,
    client_chain_id: Option<ChainId>,
    client_type: Option<ClientType>,
) -> MockContext {
    let client_config_builder = MockClientConfig::builder();

    if let Some(client_chain_id) = client_chain_id {
        client_config_builder.client_chain_id(client_chain_id);
    }

    if let Some(client_type) = client_type {
        client_config_builder.client_type(client_type);
    }

    let client_config = client_config_builder
        .client_id(client_id)
        .latest_height(height)
        .build();

   MockContext::default().with_client_config(client_config)
}
```

However, the compiler complains and prescribes calling `clone` on every line that modifies the `MockClientConfigBuilder`, which is obviously not ideal. Is there a better way to go about defining a configurable `mock_context` function?

I'm not sure where the connection tests are located.

seanchen1991: Looks like fully refactoring the client update tests to follow the approach that the connection tests use is a much bigger undertaking that deserves its own issues and PR. I will open a separate issue for that.

Farhad-Shabani: Yes. Let's take care of this as a separate issue. Thank you Sean 🙏🏻

PR #1095, opened 2024-02-19
imp: remove order_matches from ChannelEnd impl

### LINKED ISSUE #394: Remove `ChannelEnd::order_matches()`

catconcat:
Closes:  #394

## Description

Remove the `order_matches` method, refactor the current usage of `order_matches` to use match statements instead.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/improvements/394-remove-order-match.md
+++ b/.changelog/unreleased/improvements/394-remove-order-match.md
@@ -1,2 +1,2 @@
-- Deprecate order_matches method from ChannelEnd impl
+- [ibc-core] Deprecate `ChannelEnd::order_matches` method
   ([\#394](https://github.com/cosmos/ibc-rs/issues/394))
--- a/.changelog/unreleased/improvements/394-remove-order-match.md
+++ b/.changelog/unreleased/improvements/394-remove-order-match.md
@@ -1,2 +1,2 @@
-- Remove order_matches method from ChannelEnd impl
+- Deprecate order_matches method from ChannelEnd impl
   ([\#394](https://github.com/cosmos/ibc-rs/issues/394))
--- a/.changelog/unreleased/improvements/394-remove-order-match.md
+++ b/.changelog/unreleased/improvements/394-remove-order-match.md
@@ -1,0 +1,2 @@
+- Remove order_matches method from ChannelEnd impl
+  ([\#394](https://github.com/cosmos/ibc-rs/issues/394))
--- a/ibc-core/ics04-channel/src/handler/recv_packet.rs
+++ b/ibc-core/ics04-channel/src/handler/recv_packet.rs
@@ -224,40 +224,46 @@ where
             .map_err(PacketError::Channel)?;
     }
 
-    if chan_end_on_b.order_matches(&Order::Ordered) {
-        let seq_recv_path_on_b =
-            SeqRecvPath::new(&msg.packet.port_id_on_b, &msg.packet.chan_id_on_b);
-        let next_seq_recv = ctx_b.get_next_sequence_recv(&seq_recv_path_on_b)?;
-        if msg.packet.seq_on_a > next_seq_recv {
-            return Err(PacketError::InvalidPacketSequence {
-                given_sequence: msg.packet.seq_on_a,
-                next_sequence: next_seq_recv,
+    match chan_end_on_b.ordering {
+        Order::Ordered => {
+            let seq_recv_path_on_b =
+                SeqRecvPath::new(&msg.packet.port_id_on_b, &msg.packet.chan_id_on_b);
+            let next_seq_recv = ctx_b.get_next_sequence_recv(&seq_recv_path_on_b)?;
+            if msg.packet.seq_on_a > next_seq_recv {
+                return Err(PacketError::InvalidPacketSequence {
+                    given_sequence: msg.packet.seq_on_a,
+                    next_sequence: next_seq_recv,
+                }
+                .into());
             }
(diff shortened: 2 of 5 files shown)


catconcat:
Hi @Farhad-Shabani, please take a look at this PR. Thank you.

ibc-core/ics04-channel/src/handler/recv_packet.rs
```diff
@@ -224,40 +224,46 @@ where
...
-            msg.packet.seq_on_a,
-        );
-        let packet_rec = ctx_b.get_packet_receipt(&receipt_path_on_b);
-        match packet_rec {
-            Ok(_receipt) => {}
-            Err(ContextError::PacketError(PacketError::PacketReceiptNotFound { sequence }))
-                if sequence == msg.packet.seq_on_a => {}
-            Err(e) => return Err(e),
+        Order::None => {
+            // Do nothing
```
rnbguy: In the current implementation, `Order::Unordered` and `Order::None` were treated the same. Can you explain why you chose to perform nothing on `Order::None`? Does the current behavior introduce any bugs?

rnbguy (changes requested):

ibc-core/ics04-channel/src/handler/recv_packet.rs
catconcat: as @plafer suggested, doing nothing on `Order::None` is the behavior in [ibc-go](https://github.com/cosmos/ibc-go/blob/96808ebb72ec8db1edf71f4bff7e8ee68440237d/modules/core/04-channel/keeper/packet.go#L207-L248). I'm not sure if there's difference between these two implementations though.

rnbguy: ah thanks ! 

ref: [plafer's comment](https://github.com/cosmos/ibc-rs/pull/699#issuecomment-1573936340) from another PR

ibc-core/ics04-channel/types/src/channel.rs
```diff
@@ -279,11 +279,6 @@ impl ChannelEnd {
         Ok(())
     }
```
rnbguy, line 281: Deprecate the method instead of deletion for now.

```suggestion

    /// Helper function to compare the order of this end with another order.
    #[deprecated(
        since = "0.50.1",
        note = "Use `Eq` or `match` directly on the `Order` enum instead"
    )]
    pub fn order_matches(&self, other: &Order) -> bool {
        self.ordering.eq(other)
    }
```

.changelog/unreleased/improvements/394-remove-order-match.md
```diff
@@ -0,0 +1,2 @@
+- Deprecate order_matches method from ChannelEnd impl
```
rnbguy:
```suggestion
- [ibc-core] Deprecate `ChannelEnd::order_matches` method
```

rnbguy (changes requested):
Thanks @tropicaldog for the PR :pray:

Last nit before I approve the PR !

.changelog/unreleased/improvements/394-remove-order-match.md
catconcat: Thank you for the suggestion. I'll be more mindful with the writing in the future.

rnbguy (approved):
Awesome ! Thanks :sparkles:

Farhad-Shabani (changes requested):
Thank you @tropicaldog for handling this PR and @rnbguy for the review. Just wanted to make a point:

We consider `Order::None` as an [invalid order type](https://github.com/cosmos/ibc-rs/blob/0d8fcd367f12d6a9369ca4a65bd85588d1e96192/ibc-core/ics04-channel/types/src/channel.rs#L250-L257), explicitly disallowing the creation of a channel with this order type. This validation aligns with the check conducted [in ibc-go](https://github.com/cosmos/ibc-go/blob/1504c46bf9a2d3ed8ba460f465496a1d0c714cdd/modules/core/04-channel/types/channel.go#L32-L34).

Therefore, if our IBC handlers encounter a channel end with an order of `None` during the retrieval process from a store, it signals a potential manipulation in the storage. In such cases, I believe it should be treated as an error.

ibc-core/ics04-channel/src/handler/timeout.rs
```diff
@@ -215,37 +215,46 @@ where
...
-                msg.packet.seq_on_a,
-            );
-
-            client_state_of_b_on_a.verify_non_membership(
-                conn_end_on_a.counterparty().prefix(),
-                &msg.proof_unreceived_on_b,
-                consensus_state_of_b_on_a.root(),
-                Path::Receipt(receipt_path_on_b),
-            )
         };
```
rnbguy, line 258:
```suggestion
        };

```

Just to stay consistent to your change at `timeout_on_close.rs`.

rnbguy (changes requested):
Thanks @tropicaldog ! Just a nit request. Then we are ready to merge :slightly_smiling_face:

rnbguy (approved):
Nice work, @tropicaldog :rocket:

Thanks for the first contribution :purple_heart:

Farhad-Shabani (approved):
Thank you for the fix! 👌🏻

PR #1097, opened 2024-02-22
feat: detach client state verifier enabling users to implement custom verifier

### LINKED ISSUE #979: [ICS07] Detach `verifier` field from Tendermint `ClientState` struct
## Overview

Relevant Context: https://github.com/cosmos/ibc-rs/pull/958

The current Tendermint [`ClientState`](https://github.com/cosmos/ibc-rs/blob/4421b0dcd6eab6c914b8c5a7d45817b1edb6b2ce/ibc-clients/ics07-tendermint/types/src/client_state.rs#L54-L70) struct includes a `verifier` field that serves as a handle for calling underlying verification functions. This design, while functional, poses challenges in terms of consistency with the proto-generated `ClientState` struct and hinders the support for customized header/misbehaviour verifications as requested [here](https://github.com/cosmos/ibc-rs/pull/958#issue-1985552103).

## Proposal
Ideally, Tendermint clients should implement a `Verifier` trait with a `verifier` method to give callers a handle and an associated type so they can determine underlying ed25519 verifier implementation. We can then provide a built-in/standard implementation of our interfaces on the Tendermint `ClientState` struct, utilizing the default `ProdVerifier`, which is intended for the majority of users who would typically import it. Though, still projects would have the flexibility to opt for a more cost-effective verifier implementation of their choice.

How to get our design there:

1. **Refactor ClientState methods**
    - Extract the existing implementation within ClientState methods into standalone, reusable, and publicly accessible functions.
2. **Define the Verifier trait**
    - Articulate a Verifier trait that encapsulates the verification logic and offers a handle to the verifier of the user's chosen type.
3. **Implement `ClientState` with default ProdVerifier**
    - Present a readily implemented ClientState with the default verifier (ProdVerifier), catering to the needs of the majority of users.
4. **Support customized verification**
    - The `ClientState` struct can be easily re-implemented by users requiring customized verification.

### LINKED ISSUE #1078: FR: Allow custom verifier in ibc::clients::tendermint::types::ClientState
## Feature Summary

At the moment `verifier` field in `ibc::clients::tendermint::types::ClientState` is hard coded to be tendermint’s `ProdVerifier`.  That type implement signature verification using Rust implementation of Ed25519.  However, this doesn’t work for Solana where signatures need to be verified using Solana-specific code.

As an aside, the actual implementation of verification on Solana is more complex than just replacing a stateless function (specifically to verify state an extra argument needs to be passed to the verifier) but I could live without addressing that.

## Proposal

Make `ibc::clients::tendermint::types::ClientState` generic on the verifier type.

seanchen1991:
Closes: #979
Closes: #1078

Integration-tests: https://github.com/informalsystems/basecoin-rs/pull/165

## Description

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:


Farhad-Shabani (changes requested):
We're almost there, just one more step to go! 🙌

The standalone functions currently insist on getting a `client_state` argument of the [`ClientState`](https://github.com/cosmos/ibc-rs/blob/81d8f4605814e854203291e5f4f66be6930d2131/ibc-clients/ics07-tendermint/src/client_state.rs#L58) type (like [here](https://github.com/cosmos/ibc-rs/blob/81d8f4605814e854203291e5f4f66be6930d2131/ibc-clients/ics07-tendermint/src/client_state.rs#L213)), making them practically unusable for developers who want to use a custom Tendermint Verifier. 
In the custom route, they'd need to define their own newtype wrapper and implement all the `ClientState` traits on that. Therefore, they should be able to pass their own `ClientState` type to these functions. Though, there is a simpler way around: If we switch the type of the `client_state` field to `ClientStateType`, the problem is solved!

seanchen1991:
> If we switch the type of the `client_state` field to `ClientStateType`, the problem is solved

That means we'll need to implement the `TmVerifier` trait on the `ClientState` type in the ics07-tendermint-types crate. Is that acceptable?

Farhad-Shabani:
> That means we'll need to implement the `TmVerifier` trait on the `ClientState` type in the ics07-tendermint-types crate. Is that acceptable?

Good point! Upon reflection, I realized that we don't necessarily need to implement `TmVerifier` on a `ClientState` object. Since it doesn't consume anything from `self`. Therefore what about defining a unit `DefaultVerifier` struct and adding an additional argument as `verifier: &impl TmVerifier` to the `verify_header` and `verify_misbehaviour` functions? (Then, we should be able to pass a client state with the type of `CientStateType` to the standalone functions)
It would look like something as follows:

```rust
pub struct DefaultVerifier;

impl TmVerifier for DefaultVerifier {
    type Verifier = ProdVerifier;

    fn verifier(&self) -> Self::Verifier {
        ProdVerifier::default()
    }
}


impl<V> ClientStateValidation<V> for ClientState
where
    V: ClientValidationContext + TmValidationContext,
    V::AnyConsensusState: TryInto<TmConsensusState>,
    ClientError: From<<V::AnyConsensusState as TryInto<TmConsensusState>>::Error>,
{
    fn verify_client_message(
        &self,
        ctx: &V,
        client_id: &ClientId,
        client_message: Any,
    ) -> Result<(), ClientError> {
        verify_client_message(self.inner(), ctx, client_id, client_message, &DefaultVerifier)
    }
}


pub fn verify_client_message<V>(
    client_state: &ClientStateType, // now we should be able to use `CientStateType` here
    ctx: &V,
    client_id: &ClientId,
    client_message: Any,
    verifier: &impl TmVerifier,
) -> Result<(), ClientError>
where
    V: ClientValidationContext + TmValidationContext,
    V::AnyConsensusState: TryInto<TmConsensusState>,
    ClientError: From<<V::AnyConsensusState as TryInto<TmConsensusState>>::Error>,
{
    match client_message.type_url.as_str() {
        TENDERMINT_HEADER_TYPE_URL => {
            let header = TmHeader::try_from(client_message)?;
            verify_header(client_state, ctx, client_id, &header, verifier)
        }
        TENDERMINT_MISBEHAVIOUR_TYPE_URL => {
            let misbehaviour = TmMisbehaviour::try_from(client_message)?;
            verify_misbehaviour(client_state, ctx, client_id, &misbehaviour, verifier)
        }
        _ => Err(ClientError::InvalidUpdateClientMessage),
    }
}

```

ibc-clients/ics07-tendermint/src/client_state/misbehaviour.rs
```diff
@@ -9,132 +9,142 @@ use tendermint_light_client_verifier::Verifier;
...
 use crate::consensus_state::ConsensusState as TmConsensusState;
+use crate::context::TmVerifier;
+
+// verify_misbehaviour determines whether or not two conflicting headers at
+// the same height would have convinced the light client.
+pub fn verify_misbehaviour<ClientValidationContext>(
+    client_state: &TmClientState,
+    ctx: &ClientValidationContext,
+    client_id: &ClientId,
+    misbehaviour: TmMisbehaviour,
```
Farhad-Shabani: a small nit: let's take the reference of `TmMisbehaviour` here since the ownership is not needed.

ibc-clients/ics07-tendermint/src/client_state/update_client.rs
```diff
@@ -10,219 +10,228 @@ use tendermint_light_client_verifier::Verifier;
...
-    ) -> Result<(), ClientError>
-    where
-        ClientValidationContext: TmValidationContext,
+use crate::context::{CommonContext, TmVerifier, ValidationContext as TmValidationContext};
+
+pub fn verify_header<ClientValidationContext>(
+    client_state: &ClientState,
+    ctx: &ClientValidationContext,
+    client_id: &ClientId,
+    header: TmHeader,
```
Farhad-Shabani: a small nit: let's take the reference of `TmHeader` here since the ownership is not needed.

Farhad-Shabani (changes requested):

ibc-clients/ics07-tendermint/src/client_state.rs
```diff
@@ -378,152 +520,269 @@ where
...
+    }
+}
+
+/// Seed the host store with initial client and consensus states.
+///
+/// Note that this function is typically implemented as part of the
+/// [`ClientStateExecution`] trait, but has been made a standalone function
+/// in order to make the ClientState APIs more flexible.
+pub fn initialise_client_state<E>(
+    client_state: &ClientState,
```
seanchen1991: This is the one place where the `ClientState` newtype is being passed. This is because the `ClientStateType` is missing `From<ibc_client_tendermint_types::ClientState>` for `AnyClientState`; I'm not sure that's a bound that we can satisfy, but correct me if I'm wrong.

Farhad-Shabani: We can simply replace the `<E as ClientExecutionContext>::AnyClientState: From<ClientState>` constraint everywhere with `<E as ClientExecutionContext>::AnyClientState: From<ClientStateType>`, and everything should work then. This adjustment also eliminates the need for the redundant `ClientState::from(client_state)` step in situations where we need to acquire `AnyClientState` for the storing purpose.

Farhad-Shabani: While at it, I think we can do so for the `AnyConsensusState` constraint too to get things consistent. 
I mean using:
    `<E as ClientExecutionContext>::AnyConsensusState: From<ConsensusStateType>`
instead of:
    `<E as ClientExecutionContext>::AnyConsensusState: From<TmConsensusState>`,

seanchen1991: Changing `From<TmConsensusState>` to `From<ConsensusStateType>` introduces this error:
```
error[E0277]: the trait bound `<E as ClientExecutionContext>::AnyConsensusState: core::convert::From<consensus_state::ConsensusState>` is not satisfied
   --> ibc-clients/ics07-tendermint/src/client_state.rs:767:53
    |
767 |         TmConsensusState::from(new_consensus_state).into(),
    |                                                     ^^^^ the trait `core::convert::From<consensus_state::ConsensusState>` is not implemented for `<E as ClientExecutionContext>::AnyConsensusState`
    |
    = note: required for `consensus_state::ConsensusState` to implement `core::convert::Into<<E as ClientExecutionContext>::AnyConsensusState>`
help: consider further restricting the associated type
    |
710 |     <E as ClientExecutionContext>::AnyConsensusState: From<ConsensusStateType>, <E as ClientExecutionContext>::AnyConsensusState: core::convert::From<consensus_state::ConsensusState>
    |                                                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
It's not clear to me how to resolve this.

Farhad-Shabani: In situations such as [here](https://github.com/cosmos/ibc-rs/blob/fa1db8c476d50e176b06f1f8a93150588ae87dc3/ibc-clients/ics07-tendermint/src/client_state.rs#L646), where we initially built the `TmConsensusState` and then converted it to the `AnyConsensusState`, the creation of `TmConsensusState` will no longer be necessary. 
Also, other places like [here](https://github.com/cosmos/ibc-rs/blob/fa1db8c476d50e176b06f1f8a93150588ae87dc3/ibc-clients/ics07-tendermint/src/client_state.rs#L571) we should obtain `tm_consensus_state` from `ConsensusStateType`.
Then, I believe you shouldn't encounter that error.

Farhad-Shabani (approved):
Big thanks for tackling this issue @seanchen1991!
I did some tidying up. Plus, the `client_state.rs` file was getting pretty massive, making it tricky to move around. While we were pulling out implementations into standalone functions, I figured it was a good opportunity to organize the code better. So, I split the trait implementations into their own modules. Hope that makes sense to you too!

ibc-clients/ics07-tendermint/src/client_state/common.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    let last_height = latest_height.revision_height();
+
+    // Verify the proof of the upgraded client state
+    verify_membership(
+        client_state,
+        &upgrade_path_prefix,
+        &proof_upgrade_client,
+        root,
+        Path::UpgradeClient(UpgradeClientPath::UpgradedClientState(last_height)),
```
rnbguy, line 174: according to ibc-go, this should be _plan_ height.

ref: [SetUpgradedClient](https://github.com/cosmos/ibc-go/blob/698cc742b06ab1fc84c93889edd669f4bb949119/modules/core/02-client/keeper/keeper.go#L507).

ibc-clients/ics07-tendermint/src/client_state/common.rs
```diff
@@ -0,0 +1,236 @@
...
+        upgraded_client_state.to_vec(),
+    )?;
+
+    // Verify the proof of the upgraded consensus state
+    verify_membership(
+        client_state,
+        &upgrade_path_prefix,
+        &proof_upgrade_consensus_state,
+        root,
+        Path::UpgradeClient(UpgradeClientPath::UpgradedClientConsensusState(last_height)),
```
rnbguy, line 184: according to ibc-go, this should be _plan_ height.

ref: [SetUpgradedConsensusState](https://github.com/cosmos/ibc-go/blob/698cc742b06ab1fc84c93889edd669f4bb949119/modules/core/02-client/abci.go#L30).

PR #1104, opened 2024-02-28
fix(ibc-clients/tendermint): disallow empty proof-specs and empty depth range in proof-specs

### LINKED ISSUE #1100: bug(ibc-clients/tendermint): sanity check for `proof_specs` are skipped
## Bug Summary

The `proof_specs` for tendermint client state should validate if they have `min_depth <= max_depth`.

## Details

This is the current `proof_specs` validation.

https://github.com/cosmos/ibc-rs/blob/e048e3400ce5b826b8c6e496fbc54b3a52a2a592/ibc-clients/ics07-tendermint/types/src/client_state.rs#L184-L188

There should be one more check that validates the above inequality. It is best to implement a `validate` method for [`ProofSpecs`](https://github.com/cosmos/ibc-rs/blob/e048e3400ce5b826b8c6e496fbc54b3a52a2a592/ibc-core/ics23-commitment/types/src/specs.rs#L13) that includes all these checks and use it here.

```rs
pub fn validate(&self) -> bool {
  !self.0.is_empty() &&
  self.0.iter().all(|proof_spec| proof_spec.min_depth <= proof_spec.max_depth) 
}
```

## Version

<= 0.50.0

gr4yha7:
Closes: #1100 

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

- [x] Manually tested (in case integration/unit/mock tests are absent).

### DIFF
--- a/ibc-clients/ics07-tendermint/types/src/client_state.rs
+++ b/ibc-clients/ics07-tendermint/types/src/client_state.rs
@@ -419,9 +419,7 @@ pub(crate) mod serde_tests {
 
 #[cfg(test)]
 mod tests {
-    use ibc_core_commitment_types::proto::ics23::{
-        HashOp, InnerSpec, LeafOp, LengthOp, ProofSpec as Ics23ProofSpec,
-    };
+    use ibc_core_commitment_types::proto::ics23::ProofSpec as Ics23ProofSpec;
 
     use super::*;
 
@@ -462,25 +460,9 @@ mod tests {
             want_pass: bool,
         }
 
-        let leaf = LeafOp {
-            hash: HashOp::Sha256.into(),
-            prehash_key: 0,
-            prehash_value: HashOp::Sha256.into(),
-            length: LengthOp::VarProto.into(),
-            prefix: vec![0_u8],
-        };
-        let inner = InnerSpec {
-            child_order: vec![0, 1],
-            min_prefix_length: 1,
-            max_prefix_length: 1,
-            child_size: 32,
-            empty_child: vec![],
-            hash: HashOp::Sha256.into(),
-        };
-
         let empty_depth_range_proof_specs: Vec<Ics23ProofSpec> = vec![Ics23ProofSpec {
-            leaf_spec: Some(leaf),
-            inner_spec: Some(inner),
+            leaf_spec: None,
+            inner_spec: None,
             min_depth: 2,
             max_depth: 1,
             prehash_key_before_comparison: false,
--- a/ibc-clients/ics07-tendermint/types/src/client_state.rs
+++ b/ibc-clients/ics07-tendermint/types/src/client_state.rs
@@ -177,9 +177,7 @@ impl ClientState {
         }
 
         // Sanity checks on client proof specs
-        if let Err(e) = self.proof_specs.validate() {
-            return Err(Error::InvalidProofSpec(e));
-        }
+        self.proof_specs.validate()?;
 
         // `upgrade_path` itself may be empty, but if not then each key must be non-empty
         for (idx, key) in self.upgrade_path.iter().enumerate() {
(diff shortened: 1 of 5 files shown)


gr4yha7:
@rnbguy

rnbguy:
Thanks @gr4yha7 ! Can you please add some unit tests to the new `validate()` method?

ibc-clients/ics07-tendermint/types/src/client_state.rs
```diff
@@ -568,7 +594,15 @@ mod tests {
...
                     proof_specs: ProofSpecs::from(Vec::<Ics23ProofSpec>::new()),
-                    ..default_params
+                    ..default_params.clone()
+                },
+                want_pass: false,
+            },
+            Test {
+                name: "Invalid (empty) proof specs depth range".to_string(),
+                params: ClientStateParams {
+                    proof_specs: ProofSpecs::from(empty_depth_range_proof_specs),
```
rnbguy: you can directly use

```rs
proof_specs: vec![Ics23ProofSpec {
    leaf_spec: None,
    inner_spec: None,
    min_depth: 2,
    max_depth: 1,
    prehash_key_before_comparison: false,
}].into()
```

as a mock value.

ibc-core/ics23-commitment/types/src/error.rs
```diff
@@ -13,6 +13,10 @@ pub enum CommitmentError {
     EmptyMerkleRoot,
     /// empty verified value
     EmptyVerifiedValue,
+    /// empty proof specs
+    EmptyProofSpecs,
+    /// empty depth range: [{0}, {1}]
+    EmptyDepthRange(i32, i32),
```
Farhad-Shabani: Just a point caught my eye, we shouldn’t allow negative values too. I opened up another issue (https://github.com/cosmos/ibc-rs/issues/1108) for that. Let’s use `u32` here.

Farhad-Shabani (changes requested):
You may have noticed, just highlighting that there's an implicit check as part of ICS23 [here](https://github.com/cosmos/ics23/blob/9f86b674ada0b03cf0540e2c9484699e6d6bc4f1/rust/src/verify.rs#L121-L133). Though, I'm still in favor of adding an explicit check in ibc-rs to prevent the creation of faulty clients.
 All good, thank you both. Just requested a small fix!

ibc-clients/ics07-tendermint/types/src/client_state.rs
```diff
@@ -176,11 +176,9 @@ impl ClientState {
             });
         }
 
-        // Disallow empty proof-specs
-        if self.proof_specs.is_empty() {
-            return Err(Error::Validation {
-                reason: "ClientState proof-specs cannot be empty".to_string(),
-            });
+        // Sanity checks on client proof specs
+        if let Err(e) = self.proof_specs.validate() {
```
Farhad-Shabani: Just another suggestion:
Given the added `InvalidProofSpec`, let's add the following in the `error.rs`

```rust
impl From<CommitmentError> for Error {
    fn from(e: CommitmentError) -> Self {
        Self::InvalidProofSpec(e)
    }
}
```

so that can leave with `self.proof_specs.validate()?`

rnbguy: 8a985086

ibc-core/ics23-commitment/types/src/error.rs
rnbguy: Let's use `i32` here, as this is the error report. I check for non-negative values in `.validate()`.

Let's resolve your issue in future PR.

67415960

Farhad-Shabani (approved):
:pray:

rnbguy:
Thanks @gr4yha7 :pray: nice work :sparkles:

gr4yha7:
> Thanks @gr4yha7 🙏 nice work ✨

Thanks for the assistance @rnbguy

PR #1115, opened 2024-03-05
imp(api)!: refactor client relevant APIs to allow independent ICS-02 integration

### LINKED ISSUE #1114: Refactor client-related APIs for modularity and allow independent ICS-02 integration
## Summary

Reorganizing and simplifying client-related APIs under the various `ValidationContext` and `ExecutionContext` traits at different levels in ICS-24, ICS-02, and ICS-07 crates, and consolidating them primarily under the ICS-02 client validation and execution contexts.

## Objective

Enhance API delineation to facilitate modular integration of specific client state implementation based on ibc-rs without mandating to implement all the IBC stack. (Namely not requiring to implementing all the methods under the top-level [`Validation/ExecutionContext`](https://github.com/cosmos/ibc-rs/blob/4421b0dcd6eab6c914b8c5a7d45817b1edb6b2ce/ibc-core/ics24-host/src/context.rs#L28-L264) traits. This change would result in:

1. **Streamlines CosmWasm light client integration**
    - Eases independent integration of ICS02-client traits for CosmWasm contract development with ibc-rs.
2. **Enables ICS-02 client integration with any implementations**
    - Allows any IBC-adjacent implementation to be equipped with implemented light clients from ibc-rs.

## Remark

Given the scope of this issue which will include a substantial number of breaking changes, it would be more efficient to communicate and execute this enhancement through an ADR.

Farhad-Shabani:
Closes: #1114

Integration tests: https://github.com/informalsystems/basecoin-rs/pull/166

______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


seanchen1991 (approved):

ibc-core/ics02-client/context/src/context.rs
```diff
@@ -29,23 +46,30 @@ pub trait ClientValidationContext {
...
 /// and consensus states. This trait defines a uniform interface to do that for
 /// all clients.
-pub trait ClientExecutionContext: Sized {
-    type V: ClientValidationContext;
-    type AnyClientState: ClientState<Self::V, Self>;
-    type AnyConsensusState: ConsensusState;
+pub trait ClientExecutionContext:
+    ClientValidationContext<ClientStateRef = Self::ClientStateMut>
+{
+    type ClientStateMut: ClientStateExecution<Self>;
```
rnbguy, line 54: I like this solution, but I would have done it in the following way

```rs
pub trait ClientExecutionContext: ClientValidationContext
where
    Self::ClientStateRef: ClientStateExecution<Self>
{
    fn client_state_mut(&self, client_id: &ClientId) -> 
    Result<Self::ClientStateRef, ContextError>
}
```

This way we can remove the new associated type `ClientStateMut` here. We can just rename `ClientStateRef` to `ClientState` in that case.

ibc-core/ics02-client/src/handler/create_client.rs
```diff
@@ -26,11 +24,13 @@ where
     // Construct this client's identifier
     let id_counter = ctx.client_counter()?;
 
-    let client_state = ctx.decode_client_state(client_state)?;
+    let client_val_ctx = ctx.get_client_validation_context();
```
rnbguy, line 27: I am thinking, do we really need `get_client_validation_context()`. I see `ValidationContext: ClientValidationContext`. Can we use `ctx` directly in place of `client_val_ctx`?

ibc-core/ics02-client/context/src/context.rs
```diff
@@ -4,17 +4,34 @@ use ibc_core_host_types::identifiers::ClientId;
...
+    ///
+    /// Returns an error if no such state exists.
+    ///
+    /// Note: Clients have the responsibility to store consensus states on client creation and update.
+    fn consensus_state(
+        &self,
+        client_cons_state_path: &ClientConsensusStatePath,
+    ) -> Result<Self::ConsensusStateRef, ContextError>;
+
+    fn client_update_meta(
```
rnbguy, line 36: I would like `client_processed_meta` or something like `client_processed_info`.

ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
```diff
@@ -298,8 +289,9 @@ where
 impl<E> ClientStateExecution<E> for MockClientState
 where
     E: ClientExecutionContext + MockClientContext,
-    <E as ClientExecutionContext>::AnyClientState: From<MockClientState>,
-    <E as ClientExecutionContext>::AnyConsensusState: From<MockConsensusState>,
+    <E as ClientExecutionContext>::ClientStateMut: From<MockClientState>,
+    <E as ClientValidationContext>::ConsensusStateRef: From<MockConsensusState>,
+    <E as ClientValidationContext>::ConsensusStateRef: TryInto<MockConsensusState>,
```
rnbguy: you can write these two constraints in one line `From<MockConsensusState> + TryInto<MockConsensusState>`

ibc-testkit/tests/core/ics02_client/create_client.rs
```diff
@@ -48,7 +49,7 @@ fn test_create_client_ok() {
 
     assert!(res.is_ok(), "execution happy path");
 
-    let expected_client_state = ctx.decode_client_state(msg.client_state).unwrap();
+    let expected_client_state = ClientStateRef::<MockContext>::try_from(msg.client_state).unwrap();
```
rnbguy, line 52: Looks like now we have to remember the type of the context. Can we achieve something like this,

```rs
fn try_from_any<T>(_: &T, any_msg: Any) -> Result<ClientStateRef::<T>, ...> {
    ClientStateRef::<T>::try_from(any_msg)
}

let expected_client_state = try_from_any(msg.client_state).unwrap()
```

rnbguy (changes requested):
Awesome work @Farhad-Shabani ! :rocket:

I added my first batch of reviews :slightly_smiling_face:

ibc-core/ics02-client/context/src/context.rs
Farhad-Shabani: Initially, I tried this approach when I started touching relevant parts but got stuck with the compiler kept asking for a variety of trait bounds in lots of places. So, I opted for having additional associated types as a workaround so not having to deal with the trait bounds everywhere. But now you brought it up, I gave it another try on top of the current codebase and actually it worked and got nicer particularly using type aliases. 👍🏻 [ab714ae](https://github.com/cosmos/ibc-rs/pull/1115/commits/ab714ae61b644bce4e6a0478b4f917869ceb9db1)

ibc-core/ics02-client/src/handler/create_client.rs
Farhad-Shabani: Insightful question!

Currently, the approach is modular, which is especially beneficial for hosts like the Sovereign SDK, allowing the organization of each IBC layer as an independent module, which enhances maintainability, encapsulation, and overall control over components, enabling the assembly of a variety of modules. In addition, this does not enforce the necessity to implement client contexts directly on the main context object too, whereas having the `ClientValidationContext` as a super trait does enforce this requirement. These two objects can live as separate objects, which is a better implementation method IMO. Of course, we can still implement all context traits on a single object, as it is the case with the `MockContext` in the `ibc-testkit`.

ibc-core/ics02-client/context/src/context.rs
Farhad-Shabani: I just renamed this particular method and added a `client` prefix since I found the alone `update_meta` a bit confusing. As this method is newly introduced (post v0.50.0) and has not been included in any release yet, I thought it would be ok to touch it here. But to move forward with your suggestion and consistently updating namings we should touch on other relevant methods and types too. If you agree (since it will impose another group of breaking changes) let’s address it in a later PR.

ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
Farhad-Shabani: [ab714ae](https://github.com/cosmos/ibc-rs/pull/1115/commits/ab714ae61b644bce4e6a0478b4f917869ceb9db1)

ibc-testkit/tests/core/ics02_client/create_client.rs
Farhad-Shabani: I may not fully grasp your point, but it appears to be working efficiently as is.

ibc-core/ics25-handler/src/entrypoint.rs
```diff
@@ -17,16 +18,20 @@ use ibc_core_connection::handler::{
...
-use ibc_core_host::{ExecutionContext, ValidationContext};
+use ibc_core_host::{ClientStateMut, ExecutionContext, ValidationContext};
 use ibc_core_router::router::Router;
 use ibc_core_router::types::error::RouterError;
 
 /// Entrypoint which performs both validation and message execution
-pub fn dispatch(
-    ctx: &mut impl ExecutionContext,
+pub fn dispatch<Ctx>(
+    ctx: &mut Ctx,
```
rnbguy, line 26: nice ! personally, I like static dispatch over dynamic dispatch :slightly_smiling_face:

ibc-core/ics04-channel/src/handler/acknowledgement.rs
```diff
@@ -37,6 +37,7 @@ pub fn acknowledgement_packet_execute<ExecCtx>(
 ) -> Result<(), ContextError>
 where
     ExecCtx: ExecutionContext,
+    ClientStateMut<ExecCtx>: ClientStateExecution<ExecCtx::E>,
```
rnbguy: Shouldn't this bound be part of `ExecutionContext` trait?

```rs
ClientStateMut<ExecCtx>: ClientStateExecution<ExecCtx::E>
```

rnbguy: I see that it's already present

https://github.com/cosmos/ibc-rs/blob/4c4b72375592887844f513ba8fbc549a162b6626/ibc-core/ics24-host/src/context.rs#L150

ibc-clients/ics07-tendermint/src/context.rs
```diff
@@ -43,9 +43,18 @@ pub trait ValidationContext:
 ///
 /// This trait is automatically implemented for all types that implement
 /// [`ValidationContext`] and [`ClientExecutionContext`]
-pub trait ExecutionContext: ValidationContext + ClientExecutionContext {}
+pub trait ExecutionContext: ValidationContext + ClientExecutionContext
```
rnbguy, line 57: I am a bit confused :sweat_smile: Is this the Tendermint specific `ExecutionContext`?

rnbguy (changes requested):

ibc-core/ics02-client/src/handler/create_client.rs
rnbguy: If we want to maintain `ValidationContext: ClientValidationContext,` it makes sense to merge these two interfaces. IMO this is adding confusion in the interface.

For modularity, let's say for cw-contracts we can implement `ClientExecutionContext` for `OurContext`. But if one wants an `ExecutionContext` on top of this, they can extend the existing `OurContext` instead of thinking of it as a separate object.

ibc-core/ics02-client/context/src/context.rs
rnbguy: Sounds ok with me :slightly_smiling_face:

ibc-testkit/tests/core/ics02_client/create_client.rs
rnbguy: Ah sorry I wasn't clear. Here, you had to be explicit about `<MockContext>`. I was wondering if we can avoid it.

Note that `ctx` already has this type info. Currently, Rust doesn't have a way to extract type statically from a variable directly. My above suggestion is a working hack of the same.

ibc-clients/ics07-tendermint/src/context.rs
Farhad-Shabani: Right, right!

ibc-core/ics04-channel/src/handler/acknowledgement.rs
Farhad-Shabani: Yeah, exactly.
BTW, this is a part that I didn’t like first. Since it made us introduce such bounds everywhere. Though overall I think it's now considerably clean.

ibc-core/ics02-client/src/handler/create_client.rs
Farhad-Shabani: Ahh, I think I've found the source of confusion. 
Firstly, we don't have  `ValidationContext: ClientValidationContext` at the ICS-24 level. A similar bound exists between ICS-02 and ICS-07 contexts, which makes sense there. As these contexts are not expected to be implemented on different `ClientState` objects. However, we should keep these traits separate due to the Tendermint-specific methods residing under the ICS-07 `ValidationContext`.

ibc-core/ics04-channel/src/handler/acknowledgement.rs
rnbguy: So, we can remove these bounds right?

```diff
    ExecCtx: ExecutionContext,
-   ClientStateMut<ExecCtx>: ClientStateExecution<ExecCtx::E>,
```

rnbguy: It looks like my suggestion requires implied bounds (rust RFC 2089)
Check this comment - https://github.com/rust-lang/rust/issues/52662#issuecomment-473678551

Interestingly, Rust nightly can already do this via `associated_type_bounds`.

```rs
#![feature(associated_type_bounds)]

trait Color {
    type Channel;
}

trait CloneColor: Color<Channel: Clone> {}

fn foo<T: CloneColor>() {
    let f = T::Channel::clone;
}
````

**Review by rnbguy — APPROVED** [review]
LGTM 🙏🏻

Great work 🚀

File: ibc-core/ics04-channel/src/handler/acknowledgement.rs (hunk 7 above)
Comment (Farhad-Shabani) [reply ibc-core/ics04-channel/src/handler/acknowledgement.rs]: Thank you for investigating this.
Per our discussion, we're going away with this PR's implementation.

File: ibc-core/ics02-client/context/src/context.rs (hunk 1 above)
Comment (Farhad-Shabani) [reply ibc-core/ics02-client/context/src/context.rs:54]: Regrettably, the second approach poses challenges when dealing with context objects that include generic types constrained by specific traits, such as `IbcContext<S> where S: ProvableStore + Default` in basecoin-rs.
When implementing the updates from this PR, the compiler raises an `overflow evaluating the requirement` as follows:
```
error[E0275]: overflow evaluating the requirement <ibc::impls::IbcContext<S> as ibc::core::host::ExecutionContext>::E == _
--> basecoin/modules/src/ibc/impls.rs:930:1
|
930 | / impl<S> ExecutionContext for IbcContext<S>
931 | | where
932 | |     S: Store + Debug,
933 | |     ClientStateMut<Self>: ClientStateExecutionSelf::E,
| |________________________________________________________^
|
note: required for ibc::impls::IbcContext<S> to implement ibc::core::host::ExecutionContext
--> basecoin/modules/src/ibc/impls.rs:930:9
|
930 | impl<S> ExecutionContext for IbcContext<S>
|         ^^^^^^^^^^^^^^^^     ^^^^^^^^^^^^^
...
933 |     ClientStateMut<Self>: ClientStateExecutionSelf::E,
|                           ----------------------------- unsatisfied trait bound introduced here
```

Despite the intricacies of the problem, I found debugging such issues to be quite demanding. This situation is not ideal, as it could potentially leave our users stuck, requiring them to navigate how to customize things on their end due to this complication. This lack of user-friendliness prompts me to revert to the original approach.

PR #1129, opened 2024-03-16
feat(ibc-query): define request/response domain types

### LINKED ISSUE #1128: Add domain definitions for request/response query types in `ibc-query`
## Feature Summary

1. Enable usage of richer, self-expressive types for (de)serialization, eliminating the need for secondary conversion to/from in hosts.
2. Include the `query_height` field in some requests, necessary for RPC methods, particularly for non-Cosmos chains.
    - Current gRPC request proto types only return commitment proofs at the latest state machine height, which may not match the state root being checked against.
    - In the specific case of Sovereign SDK rollups, they emit an aggregated ZK proof for every number of blocks containing the latest state root covered by that proof. E.g. if a rollup is currently at height 12 and we have ZK proof for every 5 heights, we can process IBC packets up to height 10. Consequently, we should be able to query commitment proofs at height 10. So, any requests that include proof in their response should provide this capability to specify the query height.

## Additional Notes

- At the moment the Protobuf definitions do not support this query height as in the Cosmos ecosystem:
    - CometBFT-based chains can answer queries at various heights via ABCI queries
    - Cosmos SDK chains (ie. chains using ibc-go) can answer queries at various heights via the `x-cosmos-block-height` header
- These domain types serve not only for interoperability with proto types but also for hosts using different query services like JSON RPC.

Farhad-Shabani:
Closes: #1128
______

### PR author checklist:
- [ ] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-clients/ics07-tendermint/types/src/lib.rs
+++ b/ibc-clients/ics07-tendermint/types/src/lib.rs
@@ -4,6 +4,7 @@
 #![forbid(unsafe_code)]
 #![cfg_attr(not(test), deny(clippy::unwrap_used))]
 #![cfg_attr(not(test), deny(clippy::disallowed_methods, clippy::disallowed_types,))]
+#![allow(clippy::result_large_err)]
 #![deny(
     warnings,
     trivial_casts,
--- a/ibc-clients/ics07-tendermint/types/src/lib.rs
+++ b/ibc-clients/ics07-tendermint/types/src/lib.rs
@@ -4,7 +4,6 @@
 #![forbid(unsafe_code)]
 #![cfg_attr(not(test), deny(clippy::unwrap_used))]
 #![cfg_attr(not(test), deny(clippy::disallowed_methods, clippy::disallowed_types,))]
-#![allow(clippy::result_large_err)]
 #![deny(
     warnings,
     trivial_casts,
--- a/ibc-query/README.md
+++ b/ibc-query/README.md
@@ -28,6 +28,6 @@ to the proto types for efficient integration.
   not support conversion into protos due to the absence of the `query_height`
   fields.
 
-- Currently `ibc-query` do not support pagination. If pagination is a
+- Currently `ibc-query` does not support pagination. If pagination is a
   requirement for your project, please open an issue and provide details about
   your usage.
--- a/ibc-query/README.md
+++ b/ibc-query/README.md
@@ -1,7 +1,20 @@
 # IBC Query
 
-This crate provides a set of utility traits and implementations for querying the
-state of an [`ibc-rs`](https://github.com/cosmos/ibc-rs) enabled chain.
+## Overview
 
-Implementations contain essential IBC query methods and gRPC query services for
-each of the IBC core client, connection, and channel layers.
+This crate offers a comprehensive set of utility types, traits, and functions
+designed for integrating either a gRPC query server or implementing RPC methods
+in hosts. It specifically facilitates querying the state of the IBC core client,
+connection, and channel layers of a chain enabled with `ibc-rs`.
+
+## Features
+
+- Provides essential utility request/response domain types and their conversions
+to the proto types for efficient integration.
+- Provides convenient query objects with pre-implemented gRPC query services.
+- Offers convenient objects on which query service has been implemented and
+- Includes convenient `QueryContext` and `ProvableContext` traits that extend
+  the capabilities of an implemented IBC module, enabling the retrieval of state
+  from the chain.
+- Derives `serde` for all the domain types to facilitate (de)serialization of
+  the domain types.
--- a/ibc-query/README.md
+++ b/ibc-query/README.md
@@ -16,5 +16,5 @@ to the proto types for efficient integration.
 - Includes convenient `QueryContext` and `ProvableContext` traits that extend
   the capabilities of an implemented IBC module, enabling the retrieval of state
   from the chain.
-- Derives `serde` for all the domain types to facilitate (de)serialization of
-  the domain types.
+- Derives `serde` and `schema` for all the domain types to facilitate
+  (de)serialization of the domain types.
--- a/ibc-query/README.md
+++ b/ibc-query/README.md
@@ -16,5 +16,18 @@ to the proto types for efficient integration.
 - Includes convenient `QueryContext` and `ProvableContext` traits that extend
   the capabilities of an implemented IBC module, enabling the retrieval of state
   from the chain.
-- Derives `serde` and `schema` for all the domain types to facilitate
-  (de)serialization of the domain types.
+- Derives `serde` and `schema` for all the domain types enabling easy
+  (de)serialization. This feature is particularly beneficial for JSON RPC
+  implementations.
+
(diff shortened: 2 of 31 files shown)


ibc-query/src/types.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// is set.
+    pub count_total: bool,
+    /// reverse is set to true if results are to be returned in the descending order.
+    pub reverse: bool,
+}
+
+impl PageRequest {
+    pub fn all() -> Self {
+        // Note: do not use u64::MAX as the limit, as it may have unintended consequences
+        // See https://github.com/informalsystems/hermes/pull/2950#issuecomment-1373733744
```
rnbguy: This seems a Cosmos-SDK issue. Not sure why we can't do this in `ibc-rs`?

Also, the bug is from [this line in cosmos-sdk](https://github.com/cosmos/cosmos-sdk/blob/26c1db6481f70e90ab44af83d24402c0d3d51865/types/query/pagination.go#L33).

```go
var maxUint uint64 = math.MaxUint64
maxUintAsInt := int(maxUint)
maxUintAsUint := uint(maxUint)
fmt.Println(maxUintAsInt, maxUintAsUint)
// outputs: -1 18446744073709551615
```

ibc-query/src/utils.rs
```diff
@@ -0,0 +1,40 @@
...
+impl<T, Raw> IntoDomain<T> for Request<Raw>
+where
+    T: From<Raw>,
+{
+    fn into_domain(self) -> T {
+        self.into_inner().into()
+    }
+}
+
+pub trait IntoProto<Raw>: Sized
```
rnbguy: `IntoProto` naming is unfitting here - as it builds the `tonic::Response`

ibc-query/src/utils.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl<T, Raw> IntoDomain<T> for Request<Raw>
+where
+    T: From<Raw>,
+{
+    fn into_domain(self) -> T {
+        self.into_inner().into()
+    }
+}
```
rnbguy, line 29: `IntoDomain` implies `TryIntoDomain` - where the error variant is _never_ type.

Since `.into_domain()` calls are always in the scope that returns `Result<_, Status>`, we can replace `.into_domain()` calls by `.try_into_domain()?` and remove this trait.

ibc-query/src/error.rs
```diff
@@ -10,21 +12,27 @@ use tonic::Status;
...
 }
 
 impl From<QueryError> for Status {
     fn from(e: QueryError) -> Self {
         match e {
             QueryError::ContextError(e) => Status::internal(e.to_string()),
-            QueryError::ClientError(e) => Status::internal(e.to_string()),
             QueryError::IdentifierError(e) => Status::internal(e.to_string()),
             QueryError::ProofNotFound { description } => Status::not_found(description),
+            QueryError::MissingField(description) => Status::not_found(description),
```
rnbguy: We can use [`InvalidArgument`](https://docs.rs/tonic/latest/tonic/struct.Status.html#method.invalid_argument) here.

ibc-query/src/error.rs
```diff
@@ -10,21 +12,27 @@ use tonic::Status;
...
     /// Proof not found: {description}
     ProofNotFound { description: String },
+    /// Missing field: {0}
+    MissingField(String),
+}
+
+impl QueryError {
+    pub fn missing_field(description: impl ToString) -> Self {
+        QueryError::MissingField(description.to_string())
+    }
```
rnbguy: To keep things consistent, we should also create a method for `ProofNotFound`.

ibc-query/README.md
```diff
@@ -1,7 +1,33 @@
...
+
+## Remarks
+
+- At present, the Protobuf representation of request types does not include
+  support for querying at a specific height. Consequently, the current state of
+  `ibc-query` allows conversion from protos as a compatible direction but does
+  not support conversion into protos due to the absence of the `query_height`
+  fields.
+
+- Currently `ibc-query` do not support pagination. If pagination is a
```
rnbguy:
```suggestion
- Currently `ibc-query` does not support pagination. If pagination is a
```

ibc-clients/ics07-tendermint/types/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 #![forbid(unsafe_code)]
 #![cfg_attr(not(test), deny(clippy::unwrap_used))]
 #![cfg_attr(not(test), deny(clippy::disallowed_methods, clippy::disallowed_types,))]
+#![allow(clippy::result_large_err)]
```
rnbguy: Not really sure, why we have this. Without this lint, `cargo clippy` passes on my side.

rnbguy (changes requested):
Overall LGTM :+1:

Added review comments for some changes.

ibc-query/src/types.rs
rnbguy: @romac since you made the original comment, can you please validate this?

ibc-query/src/error.rs
rnbguy: fdfca24

ibc-clients/ics07-tendermint/types/src/lib.rs
rnbguy: 81690fe

ibc-query/src/error.rs
rnbguy: fd318c5

ibc-query/src/types.rs
romac: > Also, the bug is from [this line in cosmos-sdk](https://github.com/cosmos/cosmos-sdk/blob/26c1db6481f70e90ab44af83d24402c0d3d51865/types/query/pagination.go#L33).

Ah nice find! 🙌

romac: > Not sure why we can't do this in ibc-rs?

No reason to not do this in ibc-rs, so the comment is moot.

rnbguy: d67e28e

ibc-query/src/utils.rs
rnbguy: I was wrong. Ignore this.

ibc-query/src/utils.rs
rnbguy: I renamed `IntoProto` and `into_proto()` to `IntoResponse` and `into_response`.

rnbguy: ad7ca4d9

rnbguy (approved):
Great work to handle the gRPC types smoothly.

I added a few changes as discussed in previous review comments. I also removed several clippy lints.

rnbguy:
`ibc-query` crate lacks tests. We will add them in future PRs.

ibc-query/src/types.rs
rnbguy: thanks romac :pray: for prompt reply !

Issue #1045, opened 2024-01-18
imp(testkit): adopt `basecoin-store` in `ibc-testkit`


rnbguy:
## Feature Summary

Currently, `ibc-testkit` uses a MockContext with a very minimalistic data structure. We want to improve this to capture more adversarial scenarios.

To achieve this, along with other refactors, we must introduce a blockchain-compatible store like Merkle Tree.

`basecoin` already has [a working Merkle tree with a helpful set of store data structures and types](https://github.com/informalsystems/basecoin-rs/tree/main/crates/store). ~We want to move the store crate from there under ibc-testkit.~ We may directly use it when it is published on crates.io.

## Proposal

`<= 0.49.1`

rnbguy:
closed by #1109

PR #1135, opened 2024-03-19
feat(ibc-testkit): revamp `MockContext` with chain capabilities

rnbguy:
Closes: #1045

## Description

- takes account of new `ClientStateValidation` and `ClientContextValidation`
- introduces block production using Merkle root
- `begin_block` and `end_block` methods for `MockContext`
- removes clippy lints in `ibc-testkit`

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-core/ics23-commitment/types/src/commitment.rs
+++ b/ibc-core/ics23-commitment/types/src/commitment.rs
@@ -148,12 +148,6 @@ pub struct CommitmentPrefix {
     bytes: Vec<u8>,
 }
 
-impl Default for CommitmentPrefix {
-    fn default() -> Self {
-        Self { bytes: vec![0x00] }
-    }
-}
-
 impl CommitmentPrefix {
     pub fn as_bytes(&self) -> &[u8] {
         &self.bytes
--- a/ibc-testkit/src/context.rs
+++ b/ibc-testkit/src/context.rs
@@ -1,0 +1,424 @@
+use core::fmt::Debug;
+use core::ops::Sub;
+use core::time::Duration;
+
+use basecoin_store::context::ProvableStore;
+use basecoin_store::impls::{GrowingStore, InMemoryStore, RevertibleStore};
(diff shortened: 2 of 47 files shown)


rnbguy:
I managed to write a minimal proof verification test with Tendermint light client. For now, it uses the `ProofSpecs` of `basecoin_store`.

To keep the proof retrieval minimal, I had to modify the `ProvableContext::get_proof` of `MockIbcStore`. Before it returned the encoded bytes of a single`CommitmentProof`. Now it returns the encoded bytes of `MerkleProofs`, which is essentially a list of `CommitmentProof`s corresponding to a `MerklePath`.

Two shortcomings of `basecoin_store`:
- No support for cosmos-sdk specs informalsystems/basecoin-rs#171
- No support for non-membership proofs informalsystems/basecoin-rs#158

---

Ignore this comment - as the changes related to this will be processed in a separated PR.

Farhad-Shabani (approved):
I've made a small simplification by adding a `history()` method to the `TestHost` trait. 
I hope this change works well for you! 
Regarding the `HostParams` and `BlockParams`, I think we can merge them into one to streamline things. Apart from this, everything else looks good. Let's proceed with the next PR!

rnbguy:
That was a good suggestion. While doing that, I realized that the chain ID (and revision number) are not that relevant for our test cases. We can actually totally remove the chain_ids and use fixed zero as a default revision number. It is only needed to test client upgrades.

I still kept it as some of the tests are using `Height::new(1, <height>)`. Let's refactor the following in future PR:
- Add `TestHost::revision_number` that is only necessary for producing blocks (for a host with no concept of chain ID - this would just return zero)
- Make `latest_height` in `MockContextConfig` to be just revision height i.e. `u64` (as the Context does not need to deal with `revision_number` - this should be dealt via `TestHost` implementation)

PR #1139, opened 2024-03-23
fix: make client response fields public + remove reduntant `QueryConnectionChannelsRequest`

Farhad-Shabani:
Closes: #XXX

## Description
As a follow-up to PR #1129, this PR fixes some small inconsistencies.

______

### PR author checklist:

### Reviewer checklist:

### DIFF
--- a/ibc-query/src/core/client/types/response.rs
+++ b/ibc-query/src/core/client/types/response.rs
@@ -165,9 +165,9 @@ impl From<IdentifiedClientState> for RawIdentifiedClientState {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
 pub struct QueryConsensusStateResponse {
-    consensus_state: Any,
-    proof: Proof,
-    proof_height: Height,
+    pub consensus_state: Any,
+    pub proof: Proof,
+    pub proof_height: Height,
 }
 
 impl QueryConsensusStateResponse {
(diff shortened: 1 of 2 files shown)


rnbguy:
Good catches! Additionally, I realized:
- `impl Protobuf<RawType> for DomainType` is not required.
- `request.rs` only requires `impl TryFrom<RawType> for DomainType`.
- `response.rs` only requires `impl From<DomainType> for RawType`.

Farhad-Shabani:
> Good catches! Additionally, I realized:
> 
> * `impl Protobuf<RawType> for DomainType` is not required.
> * `request.rs` only requires `impl TryFrom<RawType> for DomainType`.
> * `response.rs` only requires `impl From<DomainType> for RawType`.


From the host's perspective, we definitely only need the conversion direction you mentioned. However, it's crucial that these types also cater to relayers, so we'll need to ensure the other direction is well-supported too.

Regarding `Protobuf`, I'm still leaning towards implementing it because it offers integrators convenient access to some useful encode/decode methods.

rnbguy:
> it's crucial that these types also cater to relayers

> Regarding Protobuf, I'm still leaning towards implementing it because it offers integrators convenient access to some useful encode/decode methods.

Ah. I see. In that case, I am totally fine implementing these. But I mentioned this because this requirement is not maintained for all domain<>raw types.

Shouldn't we implement `impl Protobuf<RawType> for DomainType` for all?

Farhad-Shabani:
> Shouldn't we implement `impl Protobuf<RawType> for DomainType` for all?

Correct. Currently, we haven't been able to implement that for all cases. I've explained the reason in the ibc-query's [README](https://github.com/cosmos/ibc-rs/tree/main/ibc-query#remarks).
We are in discussions with the ibc-go team to update the request protobuf definitions, incorporating the query height in relevant types. Therefore, we'll need to wait until that change is applied.

rnbguy (approved):
LGTM :sparkles:

rnbguy:
@Farhad-Shabani Thanks for your responses. Now everything is good :slightly_smiling_face:

PR #1151, opened 2024-04-03
tests: add unit tests for client recovery

seanchen1991:
## Description

Adds unit tests for validating client recovery functionality in `ibc-testkit`.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
+++ b/ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
@@ -41,7 +41,7 @@ impl MockClientState {
     pub fn new(header: MockHeader) -> Self {
         Self {
             header,
-            trusting_period: Duration::from_nanos(0),
+            trusting_period: Duration::from_nanos(100),
             frozen: false,
         }
     }
@@ -79,8 +79,8 @@ impl MockClientState {
         self.frozen
     }
 
-    fn expired(&self, _elapsed: Duration) -> bool {
-        false
+    fn expired(&self, elapsed: Duration) -> bool {
+        elapsed > self.trusting_period
     }
 }
 
--- a/ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
+++ b/ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
@@ -38,6 +38,9 @@ pub struct MockClientState {
 }
 
 impl MockClientState {
+    /// Initializes a new `MockClientState` with the given `MockHeader` and a
+    /// trusting period of 100 nanoseconds as a default. If the trusting period
+    /// needs to be changed, use the `with_trusting_period` method to override it.
     pub fn new(header: MockHeader) -> Self {
         Self {
             header,
--- a/ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
+++ b/ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
@@ -39,12 +39,12 @@ pub struct MockClientState {
 
 impl MockClientState {
     /// Initializes a new `MockClientState` with the given `MockHeader` and a
-    /// trusting period of 100 nanoseconds as a default. If the trusting period
+    /// trusting period of 10 seconds as a default. If the trusting period
     /// needs to be changed, use the `with_trusting_period` method to override it.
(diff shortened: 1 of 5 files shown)


ibc-testkit/tests/core/ics02_client/recover_client.rs
```diff
@@ -0,0 +1,206 @@
...
+    execute(&mut ctx, &mut router, msg_envelope).expect("create subject client execution");
+
+    // Ensure that the subject client state is in the `Status::Expired` state
+    // by sleeping for the length of its trusting period
+    sleep(subject_trusting_period);
+
+    // Create the substitute client
+    let substitute_client_state =
+        MockClientState::new(MockHeader::new(substitute_height).with_current_timestamp())
+            .with_trusting_period(substitute_trusting_period);
```
rnbguy: you may create a header with a specific timestamp by `with_timestamp`. That way we don't need to call `sleep`.

ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
```diff
@@ -79,8 +79,8 @@ impl MockClientState {
         self.frozen
     }
 
-    fn expired(&self, _elapsed: Duration) -> bool {
-        false
+    fn expired(&self, elapsed: Duration) -> bool {
+        elapsed > self.trusting_period
```
rnbguy, line 86: Good catch! :pray:

ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
```diff
@@ -41,7 +41,7 @@ impl MockClientState {
     pub fn new(header: MockHeader) -> Self {
         Self {
             header,
-            trusting_period: Duration::from_nanos(0),
+            trusting_period: Duration::from_nanos(100),
```
rnbguy: we can keep this untouched and use a separate method like `with_timestamp`.

rnbguy (changes requested):
Thanks ! :+1: Added a few requests.

ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
seanchen1991: This was added after discussing with @Farhad-Shabani, who felt that it wasn't correct to initialize `MockClientState`s with trusting periods of 0. It makes sense to me that it's not realistic and that this field should be initialized with a positive value.

ibc-testkit/tests/core/ics02_client/recover_client.rs
seanchen1991: Are you thinking of initializing the subject client with a timestamp that is equivalent to `Duration::now() - subject_trusting_period` so that we can avoid the `sleep` call?

rnbguy: Yes. I think it's better to just use a single timestamp at the beginning and increment it to simulate time passing.

You should use `MockHeader::with_timestamp` in that case.

ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
rnbguy: I see. In that case, can you please add a doc string for this?

ibc-testkit/tests/core/ics02_client/recover_client.rs
seanchen1991: [62f2fc7](https://github.com/cosmos/ibc-rs/pull/1151/commits/62f2fc73ab10c284673cad918a0302e20db39d84)

ibc-testkit/src/testapp/ibc/clients/mock/client_state.rs
seanchen1991: [cf24230](https://github.com/cosmos/ibc-rs/pull/1151/commits/cf242303c3d74b2c7ffdbba0b7f2ff148f29003a)

rnbguy (approved):
LGTM :tada:

Farhad-Shabani (approved):
Just `clippy` throwing some warnings, otherwise looks good! 
Thank you @seanchen1991!

Issue #374, opened 2023-01-24
Verify the header's expected value of `ibc_to_abci_client_events` test with an event from ibc-go


Farhad-Shabani:
## Summary

Verify the header's expected string value in `ibc_to_abci_client_events()` test with an emitted event on the `ibc-go`

## Details
The testing chain (respectively generated header) structure & params on `ibc-go` does not correspond with the dummy header we create through `get_dummy_ics07_header()` function. we already run the test with the mock header. Thus, it tests the conversion internally, but does not check it against an emitted `UpdateClient` event on `ibc-go`

PR #1179, opened 2024-04-21
fix(ics04): packet receive event type identifier

### LINKED ISSUE #1180: ics04 doesn't follow go implementation
## Bug Summary

As seen in the [go implementation](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/events.go#L29) the event attribute should be named `recv_packet`

penso:
Closes: #1180

# Description

Followup speaking with @rnbguy about a wrong event name I noticed while looking at optionally using ibc-rs for my Cosmos indexer, so it matches the [go implementation](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/events.go#L29)

______

### PR author checklist:
- [ ] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-core/ics04-channel/types/src/events/mod.rs
+++ b/ibc-core/ics04-channel/types/src/events/mod.rs
@@ -32,13 +32,14 @@ const CHANNEL_OPEN_ACK_EVENT: &str = "channel_open_ack";
 const CHANNEL_OPEN_CONFIRM_EVENT: &str = "channel_open_confirm";
 const CHANNEL_CLOSE_INIT_EVENT: &str = "channel_close_init";
 const CHANNEL_CLOSE_CONFIRM_EVENT: &str = "channel_close_confirm";
+const CHANNEL_CLOSED_EVENT: &str = "channel_close";
+
 /// Packet event types
 const SEND_PACKET_EVENT: &str = "send_packet";
 const RECEIVE_PACKET_EVENT: &str = "recv_packet";
 const WRITE_ACK_EVENT: &str = "write_acknowledgement";
 const ACK_PACKET_EVENT: &str = "acknowledge_packet";
 const TIMEOUT_EVENT: &str = "timeout_packet";
-const CHANNEL_CLOSED_EVENT: &str = "channel_close";
 
 #[cfg_attr(
     feature = "parity-scale-codec",
--- a/ibc-core/ics04-channel/types/src/events/mod.rs
+++ b/ibc-core/ics04-channel/types/src/events/mod.rs
@@ -25,7 +25,8 @@ use super::Version;
 use crate::error::ChannelError;
 use crate::packet::Packet;
 
-/// Channel event types
+/// Channel event types corresponding to ibc-go's channel events:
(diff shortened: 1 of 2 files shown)


ibc-core/ics04-channel/types/src/events/mod.rs
```diff
@@ -32,13 +32,14 @@ const CHANNEL_OPEN_ACK_EVENT: &str = "channel_open_ack";
 const CHANNEL_OPEN_CONFIRM_EVENT: &str = "channel_open_confirm";
 const CHANNEL_CLOSE_INIT_EVENT: &str = "channel_close_init";
 const CHANNEL_CLOSE_CONFIRM_EVENT: &str = "channel_close_confirm";
+const CHANNEL_CLOSED_EVENT: &str = "channel_close";
```
seanchen1991, line 36: Moving the `CHANNEL_CLOSED_EVENT` constant under the channel event types section introduces some ambiguity regarding the semantics and the purpose of the constant. The string itself being just `"channel_close"` when all the other `"channel_*"` constants carry some additional meaning also contributes to the ambiguity. 

Can we at least add a docstring comment on the `CHANNEL_CLOSED_EVENT` constant?

rnbguy: you mean `"channel_close_*"`, right?

It should definitely be 
```diff
- "channel_close"
+ "channel_closed"
```

rnbguy: I would suggest just adding a reference link to `ibc-go`. Because we copy these values from there.

seanchen1991: [ibc-go](https://github.com/cosmos/ibc-go/blob/c4413c5877f9ef883494da1721cb18caaba7f7f5/modules/core/04-channel/types/events.go#L60) actually uses `channel_close`, not `channel_closed`, which is a bit confusing. But if parity is what we're aiming for, then I guess we go with `channel_close`.

rnbguy: Yea, that's what I meant that we don't have a choice here :sweat_smile:

seanchen1991 (approved):
Looks good to me. Thanks for the contribution @penso!

ibc-core/ics04-channel/types/src/events/mod.rs
```diff
@@ -32,13 +32,14 @@ const CHANNEL_OPEN_ACK_EVENT: &str = "channel_open_ack";
 const CHANNEL_OPEN_CONFIRM_EVENT: &str = "channel_open_confirm";
```
Jaycat66, line 33: [cbpay-buttons.zip](https://github.com/cosmos/ibc-rs/files/15077669/cbpay-buttons.zip)
[zenledger-20230929-163048.csv](https://github.com/cosmos/ibc-rs/files/15077672/zenledger-20230929-163048.csv)

rnbguy:
Related to #374

PR #1200, opened 2024-04-26
ci(release): enforce acyclic dep order during release

rnbguy:
Closes: #XXX

## Description

The latest release failed on CI because it tried to release `ibc-client-tendermint-cw` before `ibc-testkit`. But `ibc-client-tendermint-cw` has a dev-dependency to `ibc-testkit`.

It looks like it is [an issue at `cargo-release`](https://github.com/crate-ci/cargo-release/issues/605#issuecomment-1331359921), as it ignores dev-dependency edges.

> [!IMPORTANT]
> The workspace crates must not depend on `ibc-testkit` - as it depends on `ibc`.

Also, there is a question: should we release cosmwasm contracts on crates-io, e.g. `ibc-client-tendermint-cw`, on crates-io? If not, we should exclude them in `cargo-release` command. Maybe we can compile and add the wasm blobs to release artifacts.

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


seanchen1991:
> Maybe we can compile and add the wasm blobs to release artifacts.

I think this makes sense. I think it's fine to not release `ibc-client-tendermint-cw` on crates.io so long as there isn't any demand for it, especially if it will make things simpler on the release front. We can re-visit whether it should be released on crates.io in a future version.

seanchen1991 (approved):
I added some more documentation on how to resolve acyclic dev dependencies. It looks good to me now 👍

PR #1230, opened 2024-05-16
fix(ics-23): Auto-derive `serde::Serialize` for `CommitmentPrefix`

### LINKED ISSUE #1229: Fix inconsistencies when de/serializing `CommitmentPrefix`
## Feature Summary

Raised by @soareschen:

> It looks like there are some inconsistencies between the JSON serialization and deserialization of ConnectionEnd in ibc-rs. I tried to query for the connection end from sov-ibc, and when try to parse the response using QueryConnectionResponse, I got the error:
>
> `Error: ParseError(Error("invalid type: string \"ibc\", expected struct CommitmentPrefix", line: 1, column: 139))`
>
> So it seems like the commitment_prefix field is serialized into string during serialization, but is expected to be wrapped inside a CommitmentPrefix wrapper field during JSON deserialization.
>
> I think the culprit is probably here:
```rs
#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
pub struct CommitmentPrefix {
    bytes: Vec<u8>,
}

#[cfg(feature = "serde")]
impl serde::Serialize for CommitmentPrefix {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        format!("{self:?}").serialize(serializer)
    }
}
```
> The Deserialize instance is auto derived using serde, but the Serialize instance is manually implemented to serialize it as string

## Proposal

Either auto-derive `serde::Serialize` as well for `CommitmentPrefix`, or manually deserialize it so that parsing is consistent.

seanchen1991:
Closes: #1229 

## Description

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/breaking-changes/1229-auto-derive-ser-commitment-prefix.md
+++ b/.changelog/unreleased/breaking-changes/1229-auto-derive-ser-commitment-prefix.md
@@ -1,0 +1,2 @@
+- [ics23] Auto-derive `serde::Serialize` for `CommitmentPrefix`.
+  ([\#1229](https://github.com/cosmos/ibc-rs/issues/1229))
--- a/ibc-core/ics23-commitment/types/src/commitment.rs
+++ b/ibc-core/ics23-commitment/types/src/commitment.rs
@@ -141,7 +141,7 @@ impl<'a> TryFrom<&'a CommitmentProofBytes> for MerkleProof {
     feature = "borsh",
     derive(borsh::BorshSerialize, borsh::BorshDeserialize)
 )]
-#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
 #[derive(Clone, PartialEq, Eq, Hash)]
 pub struct CommitmentPrefix {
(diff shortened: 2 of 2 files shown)


soareschen:
I think this is a reasonable fix. However, we may want to consider whether it is fine for breaking compatibility in this case. Let's wait for @rnbguy and @romac to review and discuss the changes.

romac:
I think that's fine, better to break compatibility now and have consistent serialization than the reverse.

soareschen:
Sounds good. @seanchen1991 let's add a changelog to indicate the breaking change?

seanchen1991:
@soareschen, you mentioned that you opened a fork that manually implements deserialization from `String`. Would that route be preferable than opting to auto-derive `Serialize`?

seanchen1991:
Oh, I just saw [this](https://github.com/cosmos/ibc-rs/issues/1229#issuecomment-2117001922) comment, which makes me think we should just opt to auto-derive `Serialize` instead of manually implement `Deserialize`.

soareschen:
Yes, the custom fix is at https://github.com/cosmos/ibc-rs/tree/soares/fix-commitment-prefix-deserialize. But we should serialize and deserialize it as bytes, as the current behavior for serializing non-UTF8 commitment prefix is incorrect.

soareschen (approved):

PR #1281, opened 2024-07-15
refactor(08-wasm): queries refactored to match `08-wasm` API

### LINKED ISSUE #1277: FR: Refactor CosmWasm messages to use distinct response types for each query
## Feature Summary

Currently, the contract API messages defined in the CosmWasm context [return the same response type](https://github.com/cosmos/ibc-rs/blob/e36015c94122faddb2e6cacec464b6c6f9749c91/ibc-clients/cw-context/src/types/msgs.rs#L204-L217) for each query type. This necessitates an [aggregate struct](https://github.com/cosmos/ibc-rs/blob/e36015c94122faddb2e6cacec464b6c6f9749c91/ibc-clients/cw-context/src/types/response.rs#L12-L23) with all fields optional. This approach is not ideal. Instead, the response types should be distinct and specific to each query type, similar to how they are defined in the [08-wasm module](https://github.com/cosmos/ibc-go/blob/0d06c9eacf74804c43397409d3889b57b93616fd/modules/light-clients/08-wasm/types/contract_api.go#L105-L120).

## Proposal

I have created a wrapper contract around the `ics07-tendermint` context that implements this improved approach. Here are the types I used in that contract: [response definitions](https://github.com/srdtrk/cw-ibc-lite/blob/main/packages/shared/src/types/clients/msg.rs#L65-L83) and [query definitions](https://github.com/srdtrk/cw-ibc-lite/blob/924b8e5356a805f92dda43dd6b2d6be411bb4e9c/packages/shared/src/types/clients/msg.rs#L33-L58). Note that I am using this contract for other purposes (not in 08-wasm). (This repository successfully implements all of IBC-lite in CosmWasm, allowing for end-to-end packet passing with an ibc-go counterparty without any dependency on ibc-go.)

By adopting distinct response types for each query, we can enhance clarity, maintainability, and adherence to best practices in the contract API.

srdtrk:
Closes: #1277

## Description

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-clients/cw-context/src/context/mod.rs
+++ b/ibc-clients/cw-context/src/context/mod.rs
@@ -11,14 +11,13 @@ use ibc_core::client::types::error::ClientError;
 use ibc_core::client::types::Height;
 use ibc_core::host::types::identifiers::ClientId;
 use ibc_core::host::types::path::{
-    iteration_key, ClientStatePath, ClientUpdateHeightPath, ClientUpdateTimePath,
-    ITERATE_CONSENSUS_STATE_PREFIX,
+    ClientStatePath, ClientUpdateHeightPath, ClientUpdateTimePath, ITERATE_CONSENSUS_STATE_PREFIX,
 };
 use ibc_core::primitives::proto::{Any, Protobuf};
 use prost::Message;
 
 use crate::api::ClientType;
-use crate::types::{ContractError, GenesisMetadata, HeightTravel, MigrationPrefix};
+use crate::types::{ContractError, HeightTravel, MigrationPrefix};
 use crate::utils::AnyCodec;
 
 /// - [`Height`] cannot be used directly as keys in the map,
@@ -217,58 +216,6 @@ where
         );
 
         client_update_height_path.leaf().into_bytes()
-    }
-
-    /// Returns the genesis metadata by iterating over the stored consensus
-    /// state metadata.
-    pub fn get_metadata(&self) -> Result<Option<Vec<GenesisMetadata>>, ContractError> {
-        let mut metadata = Vec::<GenesisMetadata>::new();
-
-        let iterator = CONSENSUS_STATE_HEIGHT_MAP
-            .keys(self.storage_ref(), None, None, Order::Ascending)
-            .map(|deserialized_result| {
-                let (rev_number, rev_height) =
-                    deserialized_result.map_err(|e| ClientError::Other {
-                        description: e.to_string(),
-                    })?;
-                Height::new(rev_number, rev_height)
-            });
-
-        for height_result in iterator {
-            let height = height_result?;
-
-            let processed_height_key = self.client_update_height_key(&height);
-            metadata.push(GenesisMetadata {
-                key: processed_height_key.clone().into(),
-                value: self.retrieve(&processed_height_key)?.into(),
-            });
-            let processed_time_key = self.client_update_time_key(&height);
-            metadata.push(GenesisMetadata {
-                key: processed_time_key.clone().into(),
-                value: self.retrieve(&processed_time_key)?.into(),
-            });
-        }
-
-        let iterator = CONSENSUS_STATE_HEIGHT_MAP
-            .keys(self.storage_ref(), None, None, Order::Ascending)
-            .map(|deserialized_result| {
-                let (rev_number, rev_height) =
-                    deserialized_result.map_err(|e| ClientError::Other {
-                        description: e.to_string(),
-                    })?;
-                Height::new(rev_number, rev_height)
-            });
-
-        for height_result in iterator {
-            let height = height_result?;
-
-            metadata.push(GenesisMetadata {
-                key: iteration_key(height.revision_number(), height.revision_height()).into(),
-                value: height.encode_vec().into(),
-            });
-        }
-
-        Ok(Some(metadata))
     }
 
     /// Returns the checksum of the current contract.
--- a/ibc-clients/cw-context/src/handlers.rs
+++ b/ibc-clients/cw-context/src/handlers.rs
@@ -9,10 +9,10 @@ use crate::api::ClientType;
 use crate::context::Context;
 use crate::types::{
     CheckForMisbehaviourMsg, CheckForMisbehaviourResponse, ContractError, ContractResult,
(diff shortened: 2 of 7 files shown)


ibc-clients/cw-context/src/context/mod.rs
```diff
@@ -219,58 +218,6 @@ where
         client_update_height_path.leaf().into_bytes()
     }
 
-    /// Returns the genesis metadata by iterating over the stored consensus
-    /// state metadata.
-    pub fn get_metadata(&self) -> Result<Option<Vec<GenesisMetadata>>, ContractError> {
```
srdtrk, line 224: I've removed metadata query. Read [my comment](https://github.com/cosmos/ibc-rs/issues/1277#issuecomment-2229161396) on this.

ibc-clients/cw-context/src/handlers.rs
```diff
@@ -178,26 +182,28 @@ where
...
-                let result =
+                let found_misbehaviour =
                     client_state.check_for_misbehaviour(self, &client_id, msg.client_message)?;
 
-                QueryResponse::success().misbehaviour(result)
+                to_json_binary(&CheckForMisbehaviourResponse { found_misbehaviour })
             }
-        };
-
-        Ok(to_json_binary(&resp)?)
```
srdtrk, line 201: I've moved `to_json_binary` inside the match statement because otherwise each match arm returns a different struct.

ibc-clients/cw-context/src/types/response.rs
```diff
@@ -1,57 +1,34 @@
...
 use ibc_core::client::types::Height;
 
+/// The response to [`super::msgs::QueryMsg::Status`]
 #[cw_serde]
-pub struct GenesisMetadata {
-    pub key: Binary,
-    pub value: Binary,
+pub struct StatusResponse {
+    /// The status of the client
+    // TODO: Turn this into an enum
```
srdtrk: Perhaps another issue should be created for this? Since I wasn't sure that if I replaced this with the `Status` enum, it would serialize correctly.

srdtrk (commented):
I've left some comments explaining some of my decisions

ibc-clients/cw-context/src/types/response.rs
Farhad-Shabani: I see. The thing is, we need as well to return errors propagate from the storage call as a `StatusResponse`, which does not truly indicates the status of a client. So, can't simply use the `Status` enum. What about converting the `StatusResponse` to an enum as follows?

```rust
pub enum StatusResponse {
    Ok(ibc_core::client::types::Status),
    Err(String),
}
```

If works, let's tackle it as part of this PR.

ibc-clients/cw-context/src/context/mod.rs
Farhad-Shabani: Thanks for the context. Makes sense.

tests-integration/tests/cosmwasm/fixture.rs
```diff
@@ -138,56 +138,44 @@ impl Fixture {
...
 
-    pub fn get_metadata(&self, deps: Deps<'_>) -> Option<Vec<GenesisMetadata>> {
-        self.query(deps, ExportMetadataMsg {}.into())
-            .map(|resp| resp.genesis_metadata)
-            .unwrap()
+        assert_eq!(resp.status, expected.to_string());
     }
 
-    pub fn query(&self, deps: Deps<'_>, msg: QueryMsg) -> StdResult<QueryResponse> {
+    pub fn query(&self, deps: Deps<'_>, msg: impl Into<QueryMsg>) -> StdResult<Binary> {
```
Farhad-Shabani, line 174: Can we use a generic type as follow, and remove the `.and_then(from_json)` call on the query result?

```rust
    pub fn query<T: DeserializeOwned>(
        &self,
        deps: Deps<'_>,
        msg: impl Into<QueryMsg>,
    ) -> StdResult<T> {...}
```

Farhad-Shabani (commented):
I left a couple of suggestions. Please add a changelog; otherwise, it looks neat! Thank you, @srdtrk.

tests-integration/tests/cosmwasm/fixture.rs
srdtrk: I did this initially, however, the tests didn't pass with no default features because `serde` is an optional dependency (so `serde::DeserializeOwned` can't be imported) whereas `cosmwasm_std` is not.

ibc-clients/cw-context/src/types/response.rs
srdtrk: Actually, why are we returning the error string in the status field? Doesn't it suffice for the query itself to return an error

Farhad-Shabani: Good question! 
In case of errors, the issue is that we anyway end up transmitting [a `Status` with the `Unknown` string](https://github.com/cosmos/ibc-go/blob/1156da43113d0275b3daa9d56ad1c9e7fdbe1b75/modules/light-clients/08-wasm/types/client_state.go#L68-L70) in ibc-go, which leaves us clueless about the error's cause.
By treating the error as a valid status here in `ibc-client-cw`, we avoid returning the `Unknown` and instead pass more details about the status failure calls on [`wasmQuery`](https://github.com/cosmos/ibc-go/blob/1156da43113d0275b3daa9d56ad1c9e7fdbe1b75/modules/light-clients/08-wasm/types/client_state.go#L67). Assuming we improve the error propagation in ibc-go, yes we can simply proceed with the `ibc_core::client::types::Status` type here.

tests-integration/tests/cosmwasm/fixture.rs
Farhad-Shabani: Hmm. In fact, the entire cosmwasm tests should be put under the `std` feature since they only work with `std`. Let's leave it as is for now, nothing to worry about. Thanks.

ibc-clients/cw-context/src/types/response.rs
srdtrk: added!

Farhad-Shabani (approved):
Thanks @srdtrk for the improvements!

PR #1294, opened 2024-08-02
fix: prevent expired client creation

### LINKED ISSUE #1088: bug(core): client creation only checks if the client is frozen
## Bug Summary

only non-frozen status is checked at client creation. we should just check if the client is active.

## Details

https://github.com/cosmos/ibc-rs/blob/3b260e209ec6111c1dc2520d14f65fd605fdc834/ibc-core/ics02-client/src/handler/create_client.rs#L35

~should be refactored to `if !status.is_active() { ... }`.~

The `status()` is the status of the stored client state and its corresponding stored consensus state. So this doesn't include the provided `ConsensusState`. So its status needs to be checked locally.

IBC-go also [checks for active status](https://github.com/cosmos/ibc-go/blob/c3ba0d03a99965471cd2a3c0ff0bd16875eeca78/modules/core/02-client/keeper/client.go#L41), instead of just non-frozen status.

## Version

`v0.50.0`

Farhad-Shabani:
Closes: #1088

## Highlight
This PR revamps the client creation tests, cutting out repetition and ensuring we handle all kinds of valid and invalid scenarios.
______

### PR author checklist:
- [X] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


ibc-clients/ics07-tendermint/src/client_state/common.rs
```diff
@@ -120,9 +134,49 @@ pub fn verify_consensus_state(consensus_state: Any) -> Result<(), ClientError> {
...
+    {
+        return Err(ClientError::ClientNotActive {
+            status: Status::Expired,
+        });
+    }
+
     Ok(())
 }
 
+pub fn consensus_state_status(
```
seanchen1991:
```suggestion
/// Determines the `Status`, whether it is `Active` or `Expired`, of a consensus
/// state, given its timestamp, the host's timestamp, and the trusting period.
pub fn consensus_state_status(
```

ibc-clients/ics07-tendermint/src/client_state/common.rs
```diff
@@ -120,9 +134,49 @@ pub fn verify_consensus_state(consensus_state: Any) -> Result<(), ClientError> {
...
+    consensus_timestamp: &Timestamp,
+    host_timestamp: &Timestamp,
+    trusting_period: Duration,
```
seanchen1991: There's nothing in this function's signature that actually enforces whether the `Status` it returns is actually associated with a consensus state. Not sure that's a valid criticism that warrants any changes to this implementation, but thought I would mention it.

Farhad-Shabani: Makes sense. See if [2c6b8e8](https://github.com/cosmos/ibc-rs/pull/1294/commits/2c6b8e8e818c2002b6f4cc63099d9660b670fd7f) works?

seanchen1991 (approved):

PR #1307, opened 2024-08-08
imp(ibc)!: refactor `Timestamp` type and distinguish timeout timestamp

### LINKED ISSUE #180: Timestamp define: not be strongly coupled with the time in tendermint
## Problem Definition
https://github.com/cosmos/ibc-rs/blob/da6653e0716350311d15e6acca08cad625f2d9ff/crates/ibc/src/timestamp.rs#L25
Can the definition of `timestamp` here not be strongly coupled with the time in `tendermint`.

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue

### LINKED ISSUE #1296: host and consensus state timestamp should not contain optional time field
## Background

The [`Timestamp`](https://github.com/cosmos/ibc-rs/blob/5779e868d0330999792d4148452b51f40310182d/ibc-primitives/src/types/timestamp.rs#L28-L32) definition was primarily designed to handle the packet's timestamp needs. Since a packet’s timestamp can be zero, which means it's absent. Therefore, the struct ended up using a `time` field with the type of `Option<Time>`.

## Problem Statement
This approach works fine for checking packet expiration. Though, when it comes to [consensus states](https://github.com/cosmos/ibc-rs/blob/dc116daa5e9df73b2f39b063463adc551020314e/ibc-core/ics02-client/context/src/consensus_state.rs#L19) or [asking a host for its timestamp](https://github.com/cosmos/ibc-rs/blob/5d1492dbaea05d0342859e89075d2aaf2ee9684d/ibc-core/ics24-host/src/context.rs#L40), we don't expect them to return `None`, and the timestamp should always be available. However, this definition allows light client developers and `ibc-rs` integrators to pass a `Timestamp { time: None }` to the system.

An an evidence of that optional non-needed: 
When examining how the timestamp of Tendermint consensus state is obtained, we never encounter a `None` timestamp. The conversion path is [proto Timestamp → Time → domain Timestamp](https://github.com/cosmos/ibc-rs/blob/2378cd4ba45094b8ed856ff7dba5f1d0882f59ae/ibc-clients/ics07-tendermint/types/src/consensus_state.rs#L60-L71), so it always results in a `Timestamp { time: Some(time) }`.

## Proposal

To address this, we should at least ensure timestamp availability wherever we call `host_timestamp()`.

A better design would be to create a generic type and trait for the timestamp, so that hosts and light client developers can introduce their own time types. Not being forced to use `Timestamp`, which can be challenging to conform to. Additionally, this introduces unnecessary `tendermint` dependency to downstream projects, which could cause resolving conflicts. 

## Relevant issues

   - https://github.com/cosmos/ibc-rs/issues/180
   - https://github.com/cosmos/ibc-rs/issues/459

### LINKED ISSUE #1306: `Timestamp::nanoseconds` panics when created from `tendermint::Time` with negative values
## Summary

The [`Timestamp::nanoseconds()`](https://github.com/cosmos/ibc-rs/blob/5779e868d0330999792d4148452b51f40310182d/ibc-primitives/src/types/timestamp.rs#L171-L180) method returns a `u64` value, assuming that a `Timestamp` with a negative UNIX timestamp is never constructed. While `Timestamp`'s own methods prevent this, a `Timestamp` can still be created [from `tendermint::Time`](https://github.com/cosmos/ibc-rs/blob/5779e868d0330999792d4148452b51f40310182d/ibc-primitives/src/types/timestamp.rs#L285-L291). 
The `tendermint::Time` struct can be constructed using the [`from_unix_timestamp`](https://github.com/informalsystems/tendermint-rs/blob/9331d5c3f102c51fc84a45c67b27bf1f067e0583/tendermint/src/time.rs#L87) method, which accepts an `i64` for seconds. This allows negative values to be passed. As a result, `Timestamp::nanoseconds()` can panic when created from a `tendermint::Time` with negative values.

## Version

<= v0.53.0

Farhad-Shabani:
Closes: #180
Closes: #1296
Closes: #1306

## Description
This PR introduces following improvements:
- Separates the `Timestamp` definitions used by the host from those used for timeout purposes. The host timestamp is mandatory, while a `None` value for the timeout timestamp indicates the absence of a timeout.
- Implements a direct, more efficient conversion to/from protobuf timestamps, eliminating the previous redirection through `tendermint::Time`, which was also intended for `tendermint_proto::google::Timestamp` instead of `ibc_proto::google::Timestamp`.
- Decouples timestamp definition from `tendermint::Time`, laying the groundwork for fully decoupling `ibc_primitive` from the `tendermint` dependency, making it chain-agnostic.
- Fixes an issue where `tendermint::Time` could produce negative UNIX timestamps, causing a panic when calling `Timestamp::nanoseconds()`.
- Unifies all timestamp-related errors under a single `TimestampError` type 
- Removes the error variant for `Timestamp::from_nanoseconds`, as it never fails.
______

### PR author checklist:
- [X] Added tests.
- [X] Linked to GitHub issue.

### Reviewer checklist:


ibc-core/ics04-channel/types/src/timeout/timestamp.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Returns the timestamp in nanoseconds, where 0 indicates that the absence
+    /// of a timeout.
+    pub fn nanoseconds(&self) -> u64 {
+        match self {
+            Self::At(timestamp) => timestamp.nanoseconds(),
+            Self::Never => 0,
+        }
+    }
+
+    /// Returns if the timeout timestamp is set.
```
seanchen1991:
```suggestion
    /// Returns `true` if the timeout timestamp is set.
```

ibc-primitives/src/types/timestamp.rs
```diff
@@ -2,153 +2,80 @@
...
-/// A new type wrapper over `Option<Time>` to keep track of
-/// IBC packet timeout.
-///
-/// We use an explicit `Option` type to distinguish this when converting between
-/// a `u64` value and a raw timestamp. In protocol buffer, the timestamp is
-/// represented as a `u64` Unix timestamp in nanoseconds, with 0 representing the absence
-/// of timestamp.
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+/// A new type wrapper over `PrimitiveDateTime` with extended capabilities to
+/// keep track of host timestamps.
```
seanchen1991: This doc comment makes it seem like this type is only used by hosts. I think something more along the lines of the following would be clearer:
```
/// A newtype wrapper over `PrimitiveDateTime` which serves as the foundational
/// basis for capturing timestamps. It is used directly to keep track of host
/// timestamps. It is also encoded as part of the [`ibc::channel::types::timeout::TimeoutTimestamp`]
/// type for expressly keeping track of timeout timestamps.

File: ibc-clients/ics07-tendermint/src/consensus_state.rs
Comment (seanchen1991) [inline ibc-clients/ics07-tendermint/src/consensus_state.rs:1]: I noticed the `ConsensusState` newtype wrapper defined in this file, which wraps a `ConsensusStateType`, defines its own `fn timestamp(&self) -> Time` method. This method's return type is inconsistent with the new `Timestamp` type this PR introduces; it returns `tendermint-rs`'s `Time` type, which is also a newtype wrapper around a `PrimitiveDateTime`. 

Does this method need to exist alongside the `ConsensusStateTrait::timestamp` method? More generally, there seems to be some redundancy with regards to there being multiple types for encoding timestamps used throughout ibc-rs. Are these intentional/necessary?

Comment (seanchen1991) [reply ibc-clients/ics07-tendermint/src/consensus_state.rs:1]: You mention in the PR description about starting to decouple ibc-rs from `tendermint-rs`, which the introduction of the new `Timestamp` type contributes towards. This makes me think this redundancy is not intentional and that we should perhaps remove as many uses of `tendermint::Time` as possible.

File: ibc-primitives/src/types/timestamp.rs (hunk 2 above)
Comment (Farhad-Shabani) [reply ibc-primitives/src/types/timestamp.rs]: Ahh, thanks. This is way more informative. 
Updated!

File: ibc-clients/ics07-tendermint/src/consensus_state.rs
Comment (Farhad-Shabani) [reply ibc-clients/ics07-tendermint/src/consensus_state.rs:1]: Very good point! As for the decoupling, yep. That's intentional. This has been brought up at different points by non-Cosmos users. (e.g. #180)
As for moving away from the `tendermint::Time`, I actually have some local changes (Will surface that in a separate PR later). I am aiming to introduce a generic time type for the timestamps returned by the `ValidationContext::host_timestamp` and `ConsensusStateTrait::timestamp` methods. This way, throughout the ICS-07 implementation, we still can continue using the `tendermint::Time` type. At the same time, other light clients or hosts integrating the entire `ibc-rs` will have the flexibility to introduce their own native timestamp type, whether it’s `tendermint::Time` or `ibc_primitive::Timestamp`.

File: ibc-core/ics04-channel/types/src/timeout/timestamp.rs [hunk 3]
```diff
@@ -0,0 +1,145 @@
...
+impl Add<Duration> for TimeoutTimestamp {
+    type Output = Result<Self, PacketError>;
+
+    fn add(self, rhs: Duration) -> Self::Output {
+        match self {
+            TimeoutTimestamp::At(timestamp) => {
+                let new_timestamp = timestamp.add(rhs)?;
+                Ok(TimeoutTimestamp::At(new_timestamp))
+            }
+            TimeoutTimestamp::Never => Err(PacketError::MissingTimeout),
```
seanchen1991, line 112: There's a bit of a logical inconsistency here, since a timestamp of 0 gets encoded to the `Never` variant, which can't be added or subtracted. But then that means that we can't add or subtract a timestamp of 0. Just wondering if this might cause issues.

ibc-testkit/src/testapp/ibc/core/types.rs
```diff
@@ -57,7 +57,8 @@ where
     /// A typed-store for next channel counter sequence
     pub channel_counter: JsonStore<SharedStore<S>, NextChannelSequencePath, u64>,
     /// Tracks the processed time for client updates
-    pub client_processed_times: JsonStore<SharedStore<S>, ClientUpdateTimePath, Timestamp>,
+    pub client_processed_times:
+        ProtobufStore<SharedStore<S>, ClientUpdateTimePath, Timestamp, RawTimestamp>,
```
seanchen1991: What prompted this change? Switching from `ibc::core::primitives::Timestamp` to `ibc_proto::google::protobuf::Timestamp` doesn't seem like a trivial change.

ibc-testkit/src/utils/mod.rs
```diff
@@ -1,21 +1,18 @@
...
-/// are initialized using [`Time::now()`], the second one will have a greater timestamp than the first one.
+/// are initialized using [`Timestamp::now()`], the second one will have a greater timestamp than the first one.
 /// So, the latest header of the second context cannot be submitted to first one.
 /// We can still set a custom timestamp via [`TestContextConfig`](crate::fixtures::core::context::TestContextConfig).
 pub fn year_2023() -> Timestamp {
     // Sun Jan 01 2023 00:00:00 GMT+0000
-    Time::from_unix_timestamp(1_672_531_200, 0)
-        .expect("should be a valid time")
-        .into()
+    Timestamp::from_unix_timestamp(1_672_531_200, 0).expect("should be a valid time")
```
seanchen1991, line 15: I don't suppose there's a way to make this function `const` is there? Marking the `year_2023` function `const` doesn't work since `Timestamp::from_unix_timestamp` itself isn't a `const` function. So maybe it's just not possible with this?

ibc-core/ics04-channel/types/src/timeout/timestamp.rs
Farhad-Shabani: There are few subtle details here:

(1) `Timestamp::from_nanoseconds(0)` gives us a date of 1970-01-01T00:00:00Z, on which it's reasonable to be able to perform addition or subtraction. In fact we can have `TimeoutTimestamp::At(Timestamp::from_nanoseconds(0))` which is the same way we used to with the previous `Option<Time>` type as well.

(2) Before, [any addition or subtraction involving `None` always resulted in `None`](https://github.com/cosmos/ibc-rs/blob/5779e868d0330999792d4148452b51f40310182d/ibc-primitives/src/types/timestamp.rs#L239-L240), which didn't seem right to me. Since `None` represents non-existence, there's nothing to add or subtract. so it just should return an error instead.

[9d547d3](https://github.com/cosmos/ibc-rs/pull/1307/commits/9d547d3a6dc2dccbb026359565251c5a26f24f49) I’ve added tests for different possible combinations to make things clearer. Hope those help.

ibc-testkit/src/testapp/ibc/core/types.rs
Farhad-Shabani: Good catch! Initially, I had implemented serde in the wrong way, so I was getting a stack overflow error because JsonStore couldn't encode the newly defined Timestamp. Did so as a way around. Reverted: [4b59321](https://github.com/cosmos/ibc-rs/pull/1307/commits/4b59321e437d64b7a4055929b5e1348d85cbc47b)

ibc-testkit/src/utils/mod.rs
Farhad-Shabani: Looks a no-go!
It sounds like as constants must be known at compile time, we cannot include any operations, like the function call here.

ibc-core/ics04-channel/types/src/timeout/timestamp.rs
seanchen1991: Sounds good! The tests are a great addition 🙂

seanchen1991 (approved):
Looks good to me. Awesome job putting this PR together 🙌

Issue #1319, opened 2024-08-22
Generalize and consolidate duplicated error variants


seanchen1991:
## Feature Summary

Duplicated error variants scattered across the ibc-rs codebase should be consolidated as much as possible. These should then live in a more centralized location in the repository, depending on how general they are and on how many ibc-rs crates make use of them. 

## Proposal

One concrete example of this is mentioned in [this](https://github.com/cosmos/ibc-rs/pull/1317#discussion_r1723689924) comment. All error variants that have to do with surfacing decoding errors can be consolidated under one `DecodingError` type that would live in the `ibc-primitives` crate:
```rust
pub enum DecodingError {
    /// invalid identifier: `{0}`
    InvalidIdentifier(IdentifierError),
    /// invalid field: `{0}`
    InvalidFeild(String)
    /// missing field: `{0}`
    MissingFeild(String)
    /// mismatched type URLs: expected `{expected}`, actual `{actual}`
    MismatchedTypeUrls { expected: String, actual: String },
     /// failed to decode with error: `{description}`
    FailedtoDecode { description: String },

    // ... any other relevant variants.
}
```
```[tasklist]
### Tasks
- [ ] https://github.com/cosmos/ibc-rs/pull/1325
```

seanchen1991:
@Farhad-Shabani I just realized that if this `DecodingError` type is going to belong in `ibc-primitives`, the `IdentiferError` variant can't contain an `IdentifierError` type, since that would then introduce a dependency on `ics24-host`. Would it make sense to move the `IdentifierError` type into `ibc-primitives`?

Farhad-Shabani:
> I just realized that if this `DecodingError` type is going to belong in `ibc-primitives`, the `IdentiferError` variant can't contain an `IdentifierError` type, since that would then introduce a dependency on `ics24-host`. Would it make sense to move the `IdentifierError` type into `ibc-primitives`?

Since `DecodingError` and `IdentifierError` mostly occur during Protobuf deserialization, it makes sense to keep them together. Though, I'm more inclined to place both under the `ics24-host`. After we recently refactored the `Timestamp` and differentiated the `TimeoutTimestamp` placing it under ics-04, the codebase under `ibc-primitives` now appears would be more suitable for merging into ics-24 in the future.

seanchen1991:
> the codebase under `ibc-primitives` now appears would be more suitable for merging into ics-24 in the future

[...]

Issue #950, opened 2023-11-07
Improve error handling when decoding `Any` to `MsgEnvelope`


Farhad-Shabani:
## Problem Statement

Currently, when converting an incoming IBC message of `Any` type to a `MsgEnvelope`, we are mapping decoding errors to a variant of `RouterError` (e.g. [here](https://github.com/cosmos/ibc-rs/blob/eff17e48da37dfad29e2da9f8993368510badc88/crates/ibc/src/core/msgs.rs#L60-L62)), leading to confusion. Thus, this issue should aim to:

- Fix incorrect mapping
- Utilize more efficient manner to capture the source error

We came across this issue in https://github.com/cosmos/ibc-rs/pull/949 and it is currently blocked on #270. We can address this after refactoring our error handling system.

PR #1325, opened 2024-08-28
Consolidate decoding-related errors into new `DecodingError` type

seanchen1991:
Closes: #1319 
Closes: #950 
Closes: #270 

## Description

______

### PR author checklist:
- [ ] Added tests.
- [ ] Linked to GitHub issue.

### Reviewer checklist:


ibc-apps/ics20-transfer/types/src/msgs/transfer.rs
```diff
@@ -104,12 +105,14 @@ impl TryFrom<Any> for MsgTransfer {
...
-            _ => Err(TokenTransferError::UnknownMsgType(raw.type_url)),
+            TYPE_URL => MsgTransfer::decode_vec(&raw.value)
+                .map_err(|e| TokenTransferError::Decoding(DecodingError::Protobuf(e))),
+            _ => Err(TokenTransferError::Decoding(
+                DecodingError::MismatchedTypeUrls {
+                    expected: TYPE_URL.to_string(),
+                    actual: raw.type_url,
+                },
+            )),
         }
```
Farhad-Shabani: Let's derive `derive_more::From` on `TokenTransferError` so that we can directly propagate `DecodingError` without needing to wrap with the `TokenTransferError` anymore.

Would you apply this on other locations we use `DecodingError`?

ibc-apps/ics721-nft-transfer/types/src/error.rs
```diff
@@ -69,19 +65,25 @@ impl From<Infallible> for NftTransferError {
...
+    fn from(e: IdentifierError) -> Self {
+        Self::Identifier(e)
+    }
+}
+
+impl From<DecodingError> for NftTransferError {
+    fn from(e: DecodingError) -> Self {
+        Self::Decoding(e)
     }
 }
```
Farhad-Shabani: Let's replace these lines with `derive_more::From`.

This actually can be applied all over the codebase.

ibc-core/ics02-client/types/src/status.rs
```diff
@@ -33,6 +32,15 @@ pub enum Status {
...
 }
 
+/// Encapsulates Status-related errors
+#[derive(Debug, Display)]
+pub enum StatusError {
+    /// invalid client status: `{0}`
+    InvalidStatus(String),
+    /// unexpected status found: `{0}`
+    UnexpectedStatus(Status),
+}
```
Farhad-Shabani: Honestly, I’d rather keep all the module errors in a single-level enum. I'm a bit hesitant to add another nested one, especially since we might be looking to reduce nested enums in the next step (if possible). I suggest we keep these two under `ClientError` for now. WDYT?

ibc-core/ics26-routing/types/src/error.rs
```diff
@@ -1,12 +1,13 @@
...
 /// Error type for the router module.
 #[derive(Debug, Display)]
 pub enum RouterError {
-    /// malformed message that could not be decoded: `{description}`
-    MalformedMessageBytes { description: String },
+    /// decoding error: `{0}`
+    Decoding(DecodingError),
     /// missing module
     MissingModule,
     /// unknown message type URL `{0}`
```
Farhad-Shabani: Should this `UnknownMessageTypeUrl` be removed since the `Decoding` variant can/should be used instead?

ibc-primitives/Cargo.toml
```diff
@@ -32,7 +32,8 @@ time        = { version = ">=0.3.0, <0.3.37", default-features = false }
 ibc-proto = { workspace = true }
 
 # cosmos dependencies
-tendermint = { workspace = true }
+tendermint       = { workspace = true }
+tendermint-proto = { workspace = true }
```
Farhad-Shabani: We can remove `tendermint-proto` since it's not being used anywhere. Even if we plan to use it later, it's better to avoid adding this dependency given our goal of moving away from `tendermint` altogether.

ibc-core/ics02-client/types/src/error.rs
```diff
@@ -4,31 +4,34 @@ use core::convert::Infallible;
...
     InvalidClientStateType(String),
     /// invalid client consensus state type: `{0}`
     InvalidConsensusStateType(String),
-    /// invalid header type: `{0}`
-    InvalidHeaderType(String),
     /// invalid update client message
     InvalidUpdateClientMessage,
+    // TODO(seanchen1991): Should this be removed in favor of DecodingError::InvalidIdentifier?
     /// invalid client identifier: `{0}`
     InvalidClientIdentifier(IdentifierError),
```
Farhad-Shabani: As chatted, variants like `InvalidClientIdentifier` can be merged into the `DecodingError` category. There are several other variants throughout our enum errors that occur during the decoding process, and they can also be consolidated. For example, variants like `InvalidRawHeader`, `MissingRawMisbehaviour`, and `MissingRawClientState` can all fall under `DecodingError`.Something like below if it makes sense. This would further reduce the number of variants in the error enums.


```rust
pub enum DecodingError {
    ... other variants
    
    /// missing raw data: `{description}`
    MissingRawData { description: String }
    /// invalid raw data: `{description}`
    InvalidRawData { description: String }
}

File: ibc-core/ics24-host/types/Cargo.toml [hunk 7]
```diff
@@ -33,6 +35,9 @@ ibc-primitives = { workspace = true }
 parity-scale-codec = { workspace = true, optional = true }
 scale-info         = { workspace = true, optional = true }
 
+# cosmos dependencies
+tendermint-proto = { workspace = true }
```
Farhad-Shabani: Let's shift away from `tendermint-proto`. I'm not sure if `ibc-proto-rs` exposes the `tendermint_proto::Error` we need. If it doesn't, I suggest we re-export it there, allowing us to remain dependent on `ibc-primitives` (since it exposes the necessary proto types).

Farhad-Shabani (changes requested):

ibc-core/ics26-routing/types/src/error.rs
seanchen1991: [77b42b1](https://github.com/cosmos/ibc-rs/pull/1325/commits/77b42b1e55254e0f9432c3dc941b6a15ab9ee3e1)

ibc-apps/ics20-transfer/types/src/msgs/transfer.rs
seanchen1991: [8024622](https://github.com/cosmos/ibc-rs/pull/1325/commits/802462271d1a34a28f7c247b222774975797053d)

ibc-apps/ics721-nft-transfer/types/src/error.rs
seanchen1991: As we discussed, it turns out an error type can only include at most one variant that holds a `String`. This quite heavily limits which error types we can use `derive_more::From` on. It's being used on `TokenTransferError`, `NtfTransferError`, and `RouterError`; [bfbc79e](https://github.com/cosmos/ibc-rs/pull/1325/commits/bfbc79ef2590f364f5e36c0ed78ce0c968e6d0a2)

ibc-primitives/Cargo.toml
seanchen1991: [531587c](https://github.com/cosmos/ibc-rs/pull/1325/commits/531587c4394909c3f0be2721b577d8f34a578d75)

ibc-core/ics02-client/types/src/status.rs
seanchen1991: [dc5906e](https://github.com/cosmos/ibc-rs/pull/1325/commits/dc5906edf30fa2b756b3aa34271889078bc840f3)

ibc-core/ics02-client/types/src/error.rs
seanchen1991: Cleaned up and consolidated more variants here: [f414ff7](https://github.com/cosmos/ibc-rs/pull/1325/commits/f414ff734fab0c169022079bc7940ceaed5fb810)

More probably exist that I missed, but those can be cleaned up in the future as we find them.

ibc-core/ics24-host/types/Cargo.toml
seanchen1991: It looks like `ibc-proto-rs` currently does not export the error type. Does this merely involve adding 
```rust
pub use tendermint_proto::Error;
```
to the root `lib.rs` file of `ibc-proto-rs`?

Farhad-Shabani: Exactly. Maybe we go with `String` for now and add a TODO to revisit this once it's added to `ibc-proto-rs` and the new version is integrated.

ibc-apps/ics20-transfer/types/src/coin.rs
```diff
@@ -76,7 +76,10 @@ where
                     .chars()
                     .all(|x| x.is_alphanumeric() || VALID_DENOM_CHARACTERS.contains(x))
             })
-            .ok_or_else(|| TokenTransferError::InvalidCoin(coin_str.to_string()))?;
+            .ok_or_else(|| TokenTransferError::FailedToParseType {
```
Farhad-Shabani: Following our chats, have you decided to stick with `FailedToParseType`, or should it be changed back to `InvalidCoin`?
If we’re going with `FailedToParseType`, the `expected` field is misleading IMO, as it's not representing the actual expected value.

ibc-apps/ics20-transfer/types/src/denom.rs
```diff
@@ -219,7 +219,10 @@ impl FromStr for TracePath {
         remaining_parts
             .is_none()
             .then_some(trace_path)
-            .ok_or_else(|| TokenTransferError::InvalidTrace(s.to_string()))
+            .ok_or_else(|| TokenTransferError::FailedToParseType {
+                expected: "TracePath".to_string(),
```
Farhad-Shabani: The same comment that I left for `expected: "Coin".to_string()` applies here too.

ibc-core/ics03-connection/src/handler/mod.rs
```diff
@@ -36,9 +38,7 @@ where
 
         let any_client_state = <Any as Message>::decode(wasm_client_state.data.as_slice())
             .map_err(|e| {
-                ContextError::ConnectionError(ConnectionError::InvalidClientState {
-                    description: e.to_string(),
-                })
+                ContextError::ConnectionError(ConnectionError::Decoding(DecodingError::Prost(e)))
             })?;
```
Farhad-Shabani:
```suggestion
            .map_err(|e| {
                ConnectionError::Decoding(DecodingError::Prost(e))
            })?;
```

ibc-core/ics02-client/types/src/error.rs
Farhad-Shabani: Yes actually. There are bunch of other variants across the entire repo that can be consolidated in the same way.

ibc-core/ics03-connection/src/handler/mod.rs
seanchen1991: [aeb80fc](https://github.com/cosmos/ibc-rs/pull/1325/commits/aeb80fc9ac413e1f641286bce142f7be3df6a9de)

ibc-apps/ics20-transfer/types/src/coin.rs
seanchen1991: What do you think would make more sense semantically for this variant?

ibc-core/ics24-host/types/Cargo.toml
seanchen1991: [b10ee71](https://github.com/cosmos/ibc-rs/pull/1325/commits/b10ee71322fef19bc456cc0f14aaa9d69bf9e3b5)

ibc-apps/ics20-transfer/types/src/coin.rs
Farhad-Shabani: I'd rather having them back to `InvalidCoin `, `InvalidTrace` as it was before.

I am suggesting as based on the pattern we’ve used for `expected` and `actual` fields in other cases, they represent **values** rather than **types**. Here the `expected` field is showing a type (like "Coin" or "TracePath"), but the `actual` field contains a value. They are not comparable and inconsistent with how we’ve defined similar variants elsewhere.

ibc-apps/ics20-transfer/types/src/denom.rs
seanchen1991: [2952291](https://github.com/cosmos/ibc-rs/pull/1325/commits/29522914999b02c31f3e933d7cfc273f2c24a916)

ibc-apps/ics20-transfer/types/src/coin.rs
seanchen1991: [2952291](https://github.com/cosmos/ibc-rs/pull/1325/commits/29522914999b02c31f3e933d7cfc273f2c24a916)

ibc-core/ics03-connection/types/src/error.rs
```diff
@@ -12,22 +12,18 @@ use crate::version::Version;
 
 #[derive(Debug, Display)]
 pub enum ConnectionError {
-    /// invalid identifier: `{0}`
-    InvalidIdentifier(IdentifierError),
+    /// decoding error: `{0}`
+    Decoding(DecodingError),
+    /// identifier error: `{0}`
+    Identifier(IdentifierError),
```
Farhad-Shabani: Since `IdentifierError` is now part of `DecodingError` and is used in `TryFrom` conversions, we can remove `IdentifierError` and use `DecodingError` instead.

ibc-core/ics04-channel/types/src/error.rs
```diff
@@ -16,10 +16,10 @@ use crate::Version;
 
 #[derive(Debug, Display)]
 pub enum ChannelError {
-    /// application module error: `{description}`
-    AppModule { description: String },
+    /// decoding error: `{0}`
+    Decoding(DecodingError),
     /// identifier error: `{0}`
-    InvalidIdentifier(IdentifierError),
+    Identifier(IdentifierError),
```
Farhad-Shabani: We can drop this `Identifier` variant too. Though the `From` conversion should get updated as follows:

```rust
impl From<IdentifierError> for ChannelError {
    fn from(e: IdentifierError) -> Self {
        Self::Decoding(DecodingError::Identifier(e))
    }
}
```

ibc-core/ics04-channel/types/src/error.rs
```diff
@@ -129,23 +115,43 @@ pub enum PacketError {
     },
     /// implementation-specific error
     ImplementationSpecific,
```
Farhad-Shabani: Reminder to resolve this `ImplementationSpecific` into the `HostError` later.

ibc-core/ics04-channel/types/src/error.rs
```diff
@@ -94,28 +92,16 @@ pub enum PacketError {
...
     /// invalid timeout height: `{0}`
     InvalidTimeoutHeight(ClientError),
     /// invalid timeout timestamp: `{0}`
     InvalidTimeoutTimestamp(TimestampError),
-    /// invalid identifier: `{0}`
-    InvalidIdentifier(IdentifierError),
-    /// empty acknowledgment not allowed
-    EmptyAcknowledgment,
+    /// identifier error: `{0}`
+    Identifier(IdentifierError),
```
Farhad-Shabani: The same comment I made for `ChannelError::Identifier` applies here as well.

ibc-apps/ics20-transfer/src/module.rs
```diff
@@ -203,11 +208,16 @@ pub fn on_acknowledgement_packet_validate<Ctx>(
 where
     Ctx: TokenTransferValidationContext,
 {
-    let data = serde_json::from_slice::<PacketData>(&packet.data)
-        .map_err(|_| TokenTransferError::FailedToDeserializePacketData)?;
+    let data = serde_json::from_slice::<PacketData>(&packet.data).map_err(|e| {
+        DecodingError::InvalidJson {
+            description: format!("failed to deserialize packet data: {e}"),
+        }
+    })?;
```
Farhad-Shabani: I recalled that acknowledgement status failures are written into the host's state, and changing its message would be a state machine-breaking change. This means previously stored acknowledgements wouldn't be retrievable after the update. Additionally, the exact content of this message should be controlled and specified solely by the IBC application. However, the `InvalidJson` variant prepends 'invalid JSON data' by default, which might not be desirable for some apps, like ICA. Let's revert the error variant for places dealing with `AcknowledgementStatus` to the previous one. Sorry didn't bring this up earlier.

ibc-apps/ics20-transfer/types/src/error.rs
```diff
@@ -14,14 +14,16 @@ use uint::FromDecStrErr;
...
     /// context error: `{0}`
     ContextError(ContextError),
-    /// invalid identifier: `{0}`
-    InvalidIdentifier(IdentifierError),
-    /// invalid trace: `{0}`
-    InvalidTrace(String),
+    /// decoding error: `{0}`
+    Decoding(DecodingError),
+    /// identifier error: `{0}`
+    Identifier(IdentifierError),
```
Farhad-Shabani: The same comment I made for `ChannelError::Identifier` applies here too. 
Basically wherever we have `Decoding` variant, if the `Identifier` not used other places than `TryFrom`, we can remove it.

ibc-clients/ics07-tendermint/types/src/error.rs
```diff
@@ -5,51 +5,33 @@ use core::time::Duration;
...
     /// invalid clock drift; must be greater than 0
     InvalidMaxClockDrift,
     /// invalid client proof specs: `{0}`
     InvalidProofSpec(CommitmentError),
-    /// invalid raw client state: `{description}`
-    InvalidRawClientState { description: String },
-    /// invalid raw header error: `{0}`
-    InvalidRawHeader(TendermintError),
     /// invalid raw misbehaviour: `{description}`
     InvalidRawMisbehaviour { description: String },
```
Farhad-Shabani: This variant can be removed. It seems to be incorrectly used in the places where it appears. We should either choose another variant or define a new one for those cases.

ibc-core/ics02-client/types/src/msgs/recover_client.rs
```diff
@@ -45,11 +46,11 @@ impl TryFrom<RawMsgRecoverClient> for MsgRecoverClient {
             subject_client_id: raw
                 .subject_client_id
                 .parse()
-                .map_err(ClientError::InvalidClientIdentifier)?,
+                .map_err(DecodingError::Identifier)?,
             substitute_client_id: raw
                 .substitute_client_id
                 .parse()
-                .map_err(ClientError::InvalidClientIdentifier)?,
+                .map_err(DecodingError::Identifier)?,
```
Farhad-Shabani: It would be good to add `impl From<IdentifierError> for ClientError`, so we won’t need to use `map_err` in places like this.

ibc-core/ics23-commitment/types/src/commitment.rs
```diff
@@ -94,7 +95,9 @@ impl TryFrom<Vec<u8>> for CommitmentProofBytes {
 
     fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
         if bytes.is_empty() {
-            Err(Self::Error::EmptyMerkleProof)
+            Err(DecodingError::InvalidRawData {
```
Farhad-Shabani:
```suggestion
            Err(DecodingError::MissingRawData {
```

ibc-core/ics02-client/types/src/msgs/upgrade_client.rs
```diff
@@ -59,26 +60,34 @@ impl TryFrom<RawMsgUpgradeClient> for MsgUpgradeClient {
...
+                UpgradeClientError::InvalidUpgradeClientStateProof(
+                    CommitmentError::InvalidMerkleProof,
+                )
             })?;
+
         let cs_bytes = CommitmentProofBytes::try_from(proto_msg.proof_upgrade_consensus_state)
             .map_err(|_| {
                 UpgradeClientError::InvalidUpgradeConsensusStateProof(
-                    CommitmentError::EmptyMerkleProof,
+                    CommitmentError::InvalidMerkleProof,
```
Farhad-Shabani: Since the only reason a `CommitmentProofBytes` might fail is if it's empty, the previous `EmptyMerkleProof` has been fine.

ibc-clients/ics07-tendermint/types/src/consensus_state.rs
```diff
@@ -52,27 +53,26 @@ impl TryFrom<RawConsensusState> for ConsensusState {
     fn try_from(raw: RawConsensusState) -> Result<Self, Self::Error> {
```
Farhad-Shabani, line 52: I experimented with using `DecodingError` as the type Error instead of `TendermintClientError`, while still keeping `DecodingError` as a variant within `TendermintClientError`. Everything works without any additional changes. If this holds true for other cases, I suggest we stick with `DecodingError` for all `TryFrom` conversions.

ibc-core/ics24-host/types/Cargo.toml
Farhad-Shabani: Sounds like the `tendermint-proto` somehow is re-introduced here.

ibc-core/ics26-routing/types/src/error.rs
```diff
@@ -1,16 +1,17 @@
...
 use ibc_primitives::prelude::*;
 
 /// Error type for the router module.
-#[derive(Debug, Display)]
+#[derive(Debug, Display, derive_more::From)]
 pub enum RouterError {
-    /// malformed message that could not be decoded: `{description}`
-    MalformedMessageBytes { description: String },
+    /// decoding error: `{0}`
+    Decoding(DecodingError),
```
Farhad-Shabani: This `Decoding` can be removed now.

Farhad-Shabani (changes requested):
Thank you, Sean, for addressing the previous comments. I've left a few more after a second review.

ibc-core/ics02-client/types/src/msgs/upgrade_client.rs
Farhad-Shabani: On a second thought, the `InvalidMerkleProof` is alright too.

ibc-core/ics04-channel/types/src/acknowledgement.rs
```diff
@@ -139,3 +142,9 @@ impl From<AcknowledgementStatus> for Acknowledgement {
             .expect("token transfer internal error: ack is never supposed to be empty")
     }
 }
+
+impl From<DecodingError> for StatusValue {
+    fn from(e: DecodingError) -> Self {
+        StatusValue::new(e.to_string()).expect("error message must not be empty")
+    }
+}
```
Farhad-Shabani: Based on the context [here](https://github.com/cosmos/ibc-rs/pull/1325#discussion_r1746098772), this conversion doesn't make sense.

ibc-apps/ics20-transfer/src/module.rs
seanchen1991: So any of the `InvalidJson` variants with the description "failed to deserialize acknowledgment status" need to be reverted, but do the "failed to deserialize packet data" errors in the same `on_acknowledgment_*` functions also need to be reverted?

seanchen1991: Also, I just realized that the error message for the `FailedToDeserializeAck` variant has a miss-spelling. However, from what you said about how acknowledgments are stored in hosts' state, we can't actually fix this miss-spelling by changing the error message, right?

Farhad-Shabani: Correct. More precisely both `on_recv_packet_execute` and `on_acknowledgement_packet_execute` should be reverted. In the former, the callback [returns an acknowledgement](https://github.com/cosmos/ibc-rs/blob/3ffdf2f1e5428a496d59e41117f1e75beea5c4c9/ibc-apps/ics20-transfer/src/module.rs#L173), and in the latter, it includes the acknowledgement [in the emitted event](https://github.com/cosmos/ibc-rs/blob/3ffdf2f1e5428a496d59e41117f1e75beea5c4c9/ibc-apps/ics20-transfer/src/module.rs#L273). So, any changes to either will cause breaking issues. 
In `on_acknowledgement_packet_validate`, it's not necessary, but for the sake of consistency better that be reverted too.

Farhad-Shabani: > we can't actually fix this miss-spelling by changing the error message, right?

The `acknowledgement` you mean? I feel both w/ and w/o "e" should be correct. Though not sure. Anyway, let's keep it as before.

seanchen1991: The same functions need to be reverted in the same way in the analogous `NftTransferError` application as well, right?

Farhad-Shabani: That's right.

seanchen1991: Made the changes in [283a606](https://github.com/cosmos/ibc-rs/pull/1325/commits/283a60620c5b0730be72e511a06631278d9b3d05), [3dc2ff5](https://github.com/cosmos/ibc-rs/pull/1325/commits/3dc2ff54458983d937c8e0dcca4b369649b51ec0), and [94b12af](https://github.com/cosmos/ibc-rs/pull/1325/commits/94b12affcc80762e965e4f79ff96580198f73f97).

ibc-clients/ics07-tendermint/types/src/error.rs
seanchen1991: [a53a04e](https://github.com/cosmos/ibc-rs/pull/1325/commits/a53a04eaa8e8b6b3c98a6d2c6d75b0044d93d04e)

ibc-apps/ics20-transfer/types/src/error.rs
seanchen1991: [191083d](https://github.com/cosmos/ibc-rs/pull/1325/commits/191083d9aee3475afc486d2ad2cc51c9278c0ae0)

ibc-core/ics02-client/types/src/msgs/recover_client.rs
seanchen1991: [e204dbc](https://github.com/cosmos/ibc-rs/pull/1325/commits/e204dbc3af6fed93fb57d3625290f5b525d9cb0b)

ibc-clients/ics07-tendermint/types/src/consensus_state.rs
seanchen1991: > I suggest we stick with DecodingError for all TryFrom conversions

I think this will require quite a bit of work to find all of the relevant `TryFrom` conversions, especially if we want to move towards all of the conversions actually returning `DecodingError`s: that will require a lot of refactoring. So I think this should be scoped into its own separate issue.

seanchen1991: Opened https://github.com/cosmos/ibc-rs/issues/1330 to capture this work.

ibc-core/ics26-routing/types/src/error.rs
seanchen1991: [7ed66da](https://github.com/cosmos/ibc-rs/pull/1325/commits/7ed66da9fba895d54dd1b0a3ae3c40e0d0e3e998)

ibc-core/ics04-channel/types/src/acknowledgement.rs
seanchen1991: [19bd0ce](https://github.com/cosmos/ibc-rs/pull/1325/commits/19bd0cecedbccbab5a863d270319aadcd796f602)

ibc-clients/ics07-tendermint/types/src/consensus_state.rs
Farhad-Shabani: Makes sense. Sure. Let's tackle it as a separate PR.

Farhad-Shabani (approved):

PR #1337, opened 2024-09-12
fix: remove faulty receipt check during `recv_packet_validate`

### LINKED ISSUE #1336: Faulty receipt check during receive packet validation
## Background

During the [validation of a received packet](https://github.com/cosmos/ibc-rs/blob/3055776e895bba32551ebef11b1cbdac31f0f620/ibc-core/ics04-channel/src/handler/recv_packet.rs#L139), in case of unordered channels, there is a [check](https://github.com/cosmos/ibc-rs/blob/966d0deff8e5122e92054cb3b793a190a1153862/ibc-core/ics04-channel/src/handler/recv_packet.rs#L257-L262) where the handler look for the receipt packet by calling `get_packet_receipt`. If the receipt exist or not it will do nothing and treats as a no-op. But if the storage call result in any other failure than `PacketReceiptNotFound` the handler throw out error and `recv_packet_validate` fail.

## Bug Summary

This is faulty and redundant check as it’s very probable that a host for non-existence case, be wrongly emitting another error variant, therefore `get_packet_receipt` call end up with failure instead of no-op. Besides, this storage call is redundant as in any case it's supposed to be no-op.

## Version

 <= 0.54.0

Farhad-Shabani:
Closes: #1336

______

### PR author checklist:
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-core/ics04-channel/src/handler/recv_packet.rs
+++ b/ibc-core/ics04-channel/src/handler/recv_packet.rs
@@ -50,7 +50,7 @@ where
                 let packet = &msg.packet;
                 let receipt_path_on_b =
                     ReceiptPath::new(&packet.port_id_on_b, &packet.chan_id_on_b, packet.seq_on_a);
-                ctx_b.get_packet_receipt(&receipt_path_on_b).is_ok()
+                ctx_b.get_packet_receipt(&receipt_path_on_b)?.is_ok()
             }
             Order::Ordered => {
                 let seq_recv_path_on_b =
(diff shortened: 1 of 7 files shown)


ibc-query/src/core/channel/query.rs
```diff
@@ -281,7 +281,7 @@ where
 
     // Receipt only has one enum
```
seanchen1991: This line should be removed as we added an additional variant to the `Receipt` enum.

ibc-core/ics04-channel/src/handler/recv_packet.rs
```diff
@@ -248,18 +248,11 @@ where
...
-            match packet_rec {
-                Ok(_receipt) => {}
-                Err(ContextError::PacketError(PacketError::PacketReceiptNotFound { sequence }))
-                    if sequence == msg.packet.seq_on_a => {}
-                Err(e) => return Err(e),
-            }
+            // Note: We don't check for the packet receipt here because another
+            // relayer may have already relayed the packet. If that's the case,
+            // we want to avoid failing the transaction and consuming
+            // unnecessary fees. Therefore, IBC core treats this as a no-op.
```
seanchen1991:
```suggestion
            // unnecessary fees.
```
I think this statement about this case being a no-op should be removed. With it present, it becomes confusing because the comment below mentions it being not a no-op.

seanchen1991 (approved):

PR #1353, opened 2024-09-24
imp!: define helper trait for `Timestamp` conversions + update `ConsensusState::timestamp()` to return `Result`

### LINKED ISSUE #1323: Improve compatibility between tendermint-rs's `Time` and ibc-rs's `Timestamp` types
## Feature Summary

Following the work done in #1307, ibc-rs's `Timestamp` [type](https://github.com/cosmos/ibc-rs/blob/ac1ec7327d37bdcb9189308207c179380a5f82ab/ibc-primitives/src/types/timestamp.rs#L33-L37) no longer relies upon tendermint-rs's `Time` type. Both of these types now co-exist, each one wrapping a private `PrimitiveDateTime`. For better interoperability between these co-existing types, at least one of them should expose the underlying `PrimitiveDateTime` type.

## Proposal

There are multiple avenues for how compatibility between these two types can be improved:
- [ ] Add conversions in the form of `TryFrom<PrimitiveDateTime>` and `Into<PrimitiveDateTime>` for the `Timestamp` type
- [ ] Allow host and consensus states to introduce their own timestamp types

### LINKED ISSUE #1352: `ConsensusState::timestamp()` can't properly handle possible conversion failures
## Summary

The ICS-02 `ConsensusState` trait forces any new light client implementation to [return a `Timestamp` for `timestamp()`](https://github.com/cosmos/ibc-rs/blob/dc116daa5e9df73b2f39b063463adc551020314e/ibc-core/ics02-client/context/src/consensus_state.rs#L19) implementation, without room for failure or error handling. This becomes problematic because each light client can define its own `ConsensusState` type, where the timestamp might not always match `ibc_primitives::Timestamp`. A good example is the tendermint `ConsensusState`, which uses `tendermint::Time`. Since the conversion to `ibc_primitives::Timestamp` isn’t guaranteed, there’s a risk of runtime errors.

## Proposal

To avoid these issues, it makes more sense for the `timestamp()` to return a `Result<Timestamp, ClientError>`. This would give implementations a safe way to handle conversion failures and other potential issues when dealing with different timestamp formats.

## Side Note

One way to solve this would be to redesign the trait to use an associated `type Timestamp` or a generic type, letting developers define their own timestamp type right within the trait. However, if we go that route, for consistency, we’d need to apply the same flexibility to other methods in validation and execution contexts. This would change how `ibc-rs` is integrated, since each light client would need to be integrated separately. Then we won't be needing to use the `AnyClient/ConsensusState` enums, and `ibc-derive` would no longer be needed. While this could be a more flexible and robust solution, it’s a much bigger architectural shift. Something that should be addressed through an ADR and at the right time.

Farhad-Shabani:
Closes: #1352
Closes: #1323
______

### PR author checklist:
- [X] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ibc-core/ics02-client/context/src/consensus_state.rs
+++ b/ibc-core/ics02-client/context/src/consensus_state.rs
@@ -1,5 +1,6 @@
 //! Defines the trait to be implemented by all concrete consensus state types
 
+use ibc_core_client_types::error::ClientError;
 use ibc_core_commitment_types::commitment::CommitmentRoot;
 use ibc_primitives::prelude::*;
 use ibc_primitives::proto::Any;
@@ -16,5 +17,5 @@ pub trait ConsensusState: Send + Sync + Convertible<Any> {
     fn root(&self) -> &CommitmentRoot;
 
     /// The timestamp of the consensus state
-    fn timestamp(&self) -> Timestamp;
+    fn timestamp(&self) -> Result<Timestamp, ClientError>;
 }
--- a/ibc-primitives/src/types/timestamp.rs
+++ b/ibc-primitives/src/types/timestamp.rs
@@ -36,8 +36,8 @@ pub struct Timestamp {
 
 impl Timestamp {
     pub fn from_nanoseconds(nanoseconds: u64) -> Self {
-        // As the `u64` representation can only represent times up to about year
-        // 2554, there is no risk of overflowing `Time` or `OffsetDateTime`.
+        // As the `u64` can only represent times up to about year 2554, there is
+        // no risk of overflowing `Time` or `OffsetDateTime`.
         let odt = OffsetDateTime::from_unix_timestamp_nanos(nanoseconds.into())
             .expect("nanoseconds as u64 is in the range");
         Self::from_utc(odt).expect("nanoseconds as u64 is in the range")
@@ -285,12 +285,12 @@ impl scale_info::TypeInfo for Timestamp {
 
 #[derive(Debug, Display, derive_more::From)]
 pub enum TimestampError {
-    /// parse int error: {0}
-    ParseInt(ParseIntError),
-    /// try from int error: {0}
-    TryFromInt(TryFromIntError),
-    /// failed to convert timestamp: {0}
-    Conversion(time::error::ComponentRange),
+    /// failed to parse integer error: {0}
+    FailedToParseInt(ParseIntError),
+    /// failed try from integer error: {0}
+    FailedTryFromInt(TryFromIntError),
+    /// failed to convert offset date: {0}
+    FailedToConvert(time::error::ComponentRange),
     /// invalid date: out of range
     InvalidDate,
     /// overflowed timestamp
@@ -301,9 +301,8 @@ pub enum TimestampError {
 impl std::error::Error for TimestampError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match &self {
-            Self::ParseInt(e) => Some(e),
-            Self::TryFromInt(e) => Some(e),
-            Self::Conversion(e) => Some(e),
+            Self::FailedToParseInt(e) => Some(e),
+            Self::FailedTryFromInt(e) => Some(e),
             _ => None,
         }
     }
--- a/ibc-primitives/src/types/timestamp.rs
+++ b/ibc-primitives/src/types/timestamp.rs
@@ -12,7 +12,6 @@ use ibc_proto::google::protobuf::Timestamp as RawTimestamp;
 use ibc_proto::Protobuf;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
-use tendermint::Time;
 use time::macros::offset;
 use time::{OffsetDateTime, PrimitiveDateTime};
 
@@ -104,11 +103,6 @@ impl Timestamp {
         s.try_into()
             .expect("Fails UNIX timestamp is negative, but we don't allow that to be constructed")
     }
-
-    pub fn into_tm_time(self) -> Time {
-        Time::try_from(self.time.assume_offset(offset!(UTC)))
-            .expect("Timestamp is in the range of 0..=9999 years")
-    }
 }
 
 impl Protobuf<RawTimestamp> for Timestamp {}
@@ -189,12 +183,35 @@ impl Sub<Duration> for Timestamp {
     }
 }
 
-impl TryFrom<Time> for Timestamp {
-    type Error = TimestampError;
+/// Utility trait for converting a `Timestamp` into a host-specific time format.
+pub trait IntoHostTime<T: Sized> {
+    /// Converts a `Timestamp` into another time representation of type `T`.
+    ///
+    /// This method adapts the `Timestamp` to a domain-specific format, which
+    /// could represent a custom timestamp used by a light client, or any
+    /// hosting environment that requires its own time format.
+    fn into_host_time(self) -> Result<T, TimestampError>;
+}
 
-    fn try_from(tm_time: Time) -> Result<Self, Self::Error> {
-        let odt: OffsetDateTime = tm_time.into();
-        odt.try_into()
+/// Utility trait for converting an arbitrary host-specific time format into a
+/// `Timestamp`.
+pub trait IntoTimestamp {
+    /// Converts a time representation of type `T` back into a `Timestamp`.
+    ///
+    /// This can be used to convert from custom light client or host time
+    /// formats back into the standard `Timestamp` format.
+    fn into_timestamp(self) -> Result<Timestamp, TimestampError>;
+}
+
+impl<T: TryFrom<OffsetDateTime>> IntoHostTime<T> for Timestamp {
+    fn into_host_time(self) -> Result<T, TimestampError> {
+        T::try_from(self.into()).map_err(|_| TimestampError::InvalidDate)
+    }
+}
+
+impl<T: Into<OffsetDateTime>> IntoTimestamp for T {
+    fn into_timestamp(self) -> Result<Timestamp, TimestampError> {
+        Timestamp::try_from(self.into())
     }
 }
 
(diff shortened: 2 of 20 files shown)


ibc-primitives/src/types/timestamp.rs
```diff
@@ -268,12 +285,12 @@ impl scale_info::TypeInfo for Timestamp {
 
 #[derive(Debug, Display, derive_more::From)]
 pub enum TimestampError {
-    /// parse int error: {0}
-    ParseInt(ParseIntError),
-    /// try from int error: {0}
-    TryFromInt(TryFromIntError),
-    /// failed to convert timestamp: {0}
-    Conversion(time::error::ComponentRange),
+    /// failed to parse integer error: {0}
```
Farhad-Shabani: Additionally along with this PR, I've addressed a few inconsistency in `TimestampError` with our new error handling convention.

ibc-core/ics02-client/context/src/consensus_state.rs
```diff
@@ -16,5 +17,5 @@ pub trait ConsensusState: Send + Sync + Convertible<Any> {
     fn root(&self) -> &CommitmentRoot;
```
rnbguy, line 17: Can we also change `root()` to return to just `CommitmentRoot`? Actually, this should be `Cow<'_, CommitmentRoot>` but `Cow` is only available on `std`.

ibc-primitives/src/types/timestamp.rs
rnbguy: A few nits:

```
failed to parse integer: {0}
failed try_from on integer: {0}
failed to convert offset date
```

ibc-primitives/src/types/timestamp.rs
```diff
@@ -284,9 +301,8 @@ pub enum TimestampError {
 impl std::error::Error for TimestampError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match &self {
-            Self::ParseInt(e) => Some(e),
-            Self::TryFromInt(e) => Some(e),
-            Self::Conversion(e) => Some(e),
```
rnbguy, line 289: why is `FailedToConvert` case removed ?

ibc-testkit/src/testapp/ibc/core/core_ctx.rs
```diff
@@ -48,7 +48,12 @@ where
     fn host_timestamp(&self) -> Result<Timestamp, HostError> {
         let host_height = self.host_height()?;
         let host_cons_state = self.host_consensus_state(&host_height)?;
-        Ok(host_cons_state.timestamp())
+        let timestamp = host_cons_state
+            .timestamp()
+            .map_err(|e| HostError::InvalidState {
```
rnbguy: we should implement the associated methods for `HostError` such as

```rs
impl HostError {
  fn invalid_state(description: impl ToString) -> Self {
    Self::InvalidSate { description: description.to_string() }
  }
}
```

and then, use it here as

```rs
.map_err(HostError::invalid_state)
```

ibc-primitives/src/types/timestamp.rs
```diff
@@ -189,12 +183,35 @@ impl Sub<Duration> for Timestamp {
...
+}
 
-    fn try_from(tm_time: Time) -> Result<Self, Self::Error> {
-        let odt: OffsetDateTime = tm_time.into();
-        odt.try_into()
+impl<T: Into<OffsetDateTime>> IntoTimestamp for T {
+    fn into_timestamp(self) -> Result<Timestamp, TimestampError> {
+        Timestamp::try_from(self.into())
     }
 }
```
rnbguy, line 216: Am I correct to understand, these helper traits with auto-impls only reduce extra function calls and imports ?

ibc-core/ics02-client/context/src/consensus_state.rs
Farhad-Shabani: I wonder if you've come across any limitations because of this in any of the integration projects we've worked on? I'm trying to get a better sense of whether it's really necessary.

ibc-primitives/src/types/timestamp.rs
Farhad-Shabani: That’s one benefit. But more importantly it removes `tendermint` dependency in `ibc_primitives`. This fits with the plan to completely decouple the ibc-rs core stack from the `tendermint`.

It also makes things easier for light client developers or integrators. By implementing these traits, they can smoother convert their own time type to `Timestamp` wherever ibc-rs asks for, or reconstruct their types from ibc-rs’s `Timestamp` if needed.

ibc-testkit/src/testapp/ibc/core/core_ctx.rs
Farhad-Shabani: Oh, slipped my mind we've added these methods. Ty. [4f11258](https://github.com/cosmos/ibc-rs/pull/1353/commits/4f112581c4155194e8830db65f85e14355e6db03)

ibc-primitives/src/types/timestamp.rs
Farhad-Shabani: [4f11258](https://github.com/cosmos/ibc-rs/pull/1353/commits/4f112581c4155194e8830db65f85e14355e6db03)

ibc-primitives/src/types/timestamp.rs
Farhad-Shabani: [4f11258](https://github.com/cosmos/ibc-rs/pull/1353/commits/4f112581c4155194e8830db65f85e14355e6db03)

ibc-core/ics02-client/context/src/consensus_state.rs
rnbguy: Currently, we don't have any limitations - except in the `MockConsensusState` - as our Tendermint Light Client's consensus state just stores the root as `CommitmentRoot` type.

But the issue is, we can't assume how the implementor of `ConsensusStateTrait` is storing the commitment root. `fn root(&self)` can only return a reference of `CommitmentRoot` if the commitment root is stored as `CommitmentRoot` in `Self`.

For example, the following `DummyConsensusState` can't implement the trait.

```rs
struct DummyConsensusState{ root: Vec<u8> };
```

ibc-primitives/src/types/timestamp.rs
rnbguy: I meant whether these traits are just _syntactic sugars_. Because, the developers could also just use `T::try_from(OffsetDateTime::from(ibc_timestamp))` instead of `ibc_timestamp.into_host_time()` and `Timestamp::try_from(OffsetDateTime::from(host_timestamp))` instead of `host_timestamp.into_timestamp()`.

Farhad-Shabani: Ah, I get what you mean now. It’s not quite that way.
We can't really assume the structure of a time type across different hosts or what conversions they use. They might not even be using `OffsetDateTime`. So, I added default implementations for any time type that follows this conversion path, just for convenience. As it's the case with `tendermint::Time`. But this won’t necessarily apply to other timestamp types in different `ConsensusState` structs.

ibc-core/ics02-client/context/src/consensus_state.rs
Farhad-Shabani: Got it, I see now what you're saying. Right, it implicitly forces developers to use `CommitmentRoot` as the `root` type. However, one concern I see with this update is that everywhere we call `.root()` we'll end up with unnecessary cloning and increased overhead. Alternatively, we could make `fn root(self)` take ownership of self, but that would still require cloning the consensus state in multiple places since we need `timestamp()` too from the consensus state in those instances. I'm actually prioritizing efficiency over flexibility here.
I suggest we hold off on this change until we get a clear request from our clients. Further improvements here looks to me be pushing us towards using associated types or generics, which would need more thoughts. So, it’s probably better to handle this in a separate PR since it’s outside the scope of this one. What do you think?

rnbguy: > we'll end up with unnecessary cloning

Yea [`Cow`](https://doc.rust-lang.org/stable/std/borrow/enum.Cow.html) is the solution for that :smile: But it requires `std`.

Alright, let's fix this in future.

ibc-primitives/src/types/timestamp.rs
rnbguy: Hmm. I am trying to understand, why would user need these two helper traits?

It seems to me, for a struct `HostTime`, they can just implement `impl TryFrom<Timestamp> for HostTime` and `impl TryFrom<HostTime> for Timestamp`.

We can just use the trait bounds as `TryFrom<Timestamp>` and `TryFrom<HostTime>` appropriately in ibc-rs.

rnbguy: Basically, `IntoHostTime<T>` is equivalent to `TryInto<T>` and `IntoTimestamp` is equivalent to `TryInto<Timestamp>`.

rnbguy (approved):
LGTM :sparkles: I will let you decide on the helper traits.

ibc-primitives/src/types/timestamp.rs
Farhad-Shabani: > Basically, `IntoHostTime<T>` is equivalent to `TryInto<T>` and `IntoTimestamp` is equivalent to `TryInto<Timestamp>`.

Totally, this works too. But if a host has multiple `TryFrom<HostTime>` implementations, which usually can be the case, the consumer would need to specify the exact type and use the full path. Can’t simply call `try_from` or `try_into` on the time object. Plus, as you mentioned, this would definitely benefit from being syntactic sugar, making it more readable and easier to work with.

rnbguy: Yea. I am ok with this. Thanks, Farhad, for explaining.

PR #1397, opened 2025-01-27
fix: serde support for `Height` without `revision_number`

### LINKED ISSUE #1262: wasm client fails to query `TimestampAtHeight`
## Bug Summary

With a wasm light client, compiled from ibc-rs, ibc-go fails to query `TimestampAtHeight`. The error it raises:

> Error: Status { code: Unknown, message: "failed to execute message; message index: 0: height (0-32): Error parsing into type ibc_client_cw::types::msgs::QueryMsg: missing field `revision_number`: wasm contract call failed [cosmos/ibc-go/modules/light-clients/08-wasm/types/vm.go:212] With gas wanted: '18446744073709551615' and gas used: '106008' ", metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "56"} }, source: None }

## Details

This happens as ibc-go [omits `revision_number`](https://github.com/cosmos/ibc-go/blob/99f3814015044d4f39857820ed9c3e81682b664e/modules/core/02-client/types/client.pb.go#L207-L212) field if it is set to zero [while mashaling to JSON](https://github.com/cosmos/ibc-go/blob/99f3814015044d4f39857820ed9c3e81682b664e/modules/light-clients/08-wasm/keeper/contract_keeper.go#L140). From [`json` go package](https://pkg.go.dev/encoding/json#Marshal):

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

```go
type Height struct {
	RevisionNumber int `json:"revision_number,omitempty"`
	RevisionHeight int `json:"revision_height,omitempty"`
}
h := Height{RevisionNumber: 0, RevisionHeight: 12}
hJSON, _ := json.Marshal(h)
fmt.Printf("%s", hJSON) // {"revision_height":12}
```

To make it work, we need to annotate our `revision_number` field with  [`#[serde(default)]`](https://serde.rs/attr-default.html) with `#[cfg_attr(feature = "serde", ...)]`.
https://github.com/cosmos/ibc-rs/blob/8424f6903045b203614f12a5839c330a3559cece/ibc-core/ics02-client/types/src/height.rs#L32-L34

## Version

v0.53.0

rnbguy:
Closes: #1262

## Description

Makes `revision_number` field optional for `Height` struct when deserializing via serde -- specifically for JSON. 

______

### PR author checklist:
- [x] Added tests.
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/bug-fixes/1262-deserialize-height-without-revision-number.md
+++ b/.changelog/unreleased/bug-fixes/1262-deserialize-height-without-revision-number.md
@@ -1,0 +1,2 @@
+- Serde support for `Height` without `revision_number`
+  ([#1262](https://github.com/informalsystems/ibc-rs/issues/1262)).
--- a/.changelog/unreleased/bug-fixes/1262-deserialize-height-without-revision-number.md
+++ b/.changelog/unreleased/bug-fixes/1262-deserialize-height-without-revision-number.md
@@ -1,2 +1,2 @@
-- Serde support for `Height` without `revision_number`
-  ([#1262](https://github.com/informalsystems/ibc-rs/issues/1262)).
+- [ibc-core-client-types] Serde support for `Height` without `revision_number`
+  ([#1262](https://github.com/cosmos/ibc-rs/issues/1262)).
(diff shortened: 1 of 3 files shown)


Farhad-Shabani (approved):
Thanks for following up on this and resolving it!

